PR #1, opened 2025-01-26
Sync Committee: Task Errors & Rescheduling

zadykian:
#### This PR provides several changes of task error handling and propagation

* Instead of plain error text, `TaskExecError` is returned from the executor (prover / proof_provider) as a part of `TaskResult`. Thus, it is now possible to distinguish critical errors from non-critical ones and handle them differently.

* `TaskStorage.ProcessTaskResult(...)` method is revorked to support new error system. In case of a non-critical error, task is reset with an increased `RetryCount`.

* ProofProvider now doesn't propagate non-critical error to the parent task (check [TaskStateChangeHandler](https://github.com/NilFoundation/nil/pull/1/files#diff-61a5e6162de21591605faf16f77307c62b251d9101a7283290fef073d7e970a7));

* Prover's `TaskHandler`:
    - Now tracks kernel signals and return codes of the child process (`proof-producer`);
    - Maps error returned from `handleImpl(...)` to `TaskExecErr` type;

#### Other changes:

* `TaskScheduler` now validates received result against task entry before `OnTaskTerminated` call;

* Metrics in `TaskStorage` are being published only after successful tx commit in `ProcessTaskResult` and `RescheduleHangingTasks` methods;

* Declared `commonStorage` type reused by `TaskStorage`, `BlockStorage` and `TaskResultStorage`;

* `client.Client` interface is replaced with `api.TracerRpcClient` which has a narrowed method set (only those that are actually used by tracer);

### DIFF
--- a/nil/services/synccommittee/internal/types/errors.go
+++ b/nil/services/synccommittee/internal/types/errors.go
@@ -18,8 +18,6 @@ var (
 	ErrTaskWrongExecutor = errors.New("task belongs to another executor")
 )
 
-var ErrRpcCallFailed = errors.New("rpc call failed")
-
 type TaskErrType int8
 
 const (
@@ -28,7 +26,10 @@ const (
 	// TaskErrTimeout indicates that a task failed as it exceeded the specified timeout duration.
 	TaskErrTimeout
 
-	// TaskErrIO indicates an error related to input/output operations (RPC calls / Disk storage access / Network).
+	// TaskErrRpc indicates an error related to remote procedure calls (RPC) during task execution.
+	TaskErrRpc
+
+	// TaskErrIO indicates an error related to input/output operations (Disk storage access / Network).
 	TaskErrIO
 
 	// TaskErrInvalidTask indicates an error caused by an invalid or improperly structured task.
@@ -51,6 +52,8 @@ const (
 )
 
 var RetryableErrors = map[TaskErrType]bool{
+	TaskErrTimeout:    true,
+	TaskErrRpc:        true,
 	TaskErrIO:         true,
 	TaskErrTerminated: true,
 	TaskErrUnknown:    true,
--- a/nil/services/synccommittee/core/task_state_change_handler.go
+++ b/nil/services/synccommittee/core/task_state_change_handler.go
@@ -35,6 +35,8 @@ func (h taskStateChangeHandler) OnTaskTerminated(ctx context.Context, task *type
 	}
 
 	if !result.IsSuccess() {
+		// TODO: handle critical errors here
+
 		log.NewTaskResultEvent(h.logger, zerolog.WarnLevel, result).
 			Msg("block proof task has failed, data won't be sent to the L1")
 		return nil
--- a/nil/services/synccommittee/prover/commands/partial_proof.go
+++ b/nil/services/synccommittee/prover/commands/partial_proof.go
@@ -7,17 +7,16 @@ import (
 	"path/filepath"
 	"slices"
 
-	"github.com/NilFoundation/nil/nil/client"
-	"github.com/NilFoundation/nil/nil/common/logging"
 	"github.com/NilFoundation/nil/nil/services/rpc/transport"
 	"github.com/NilFoundation/nil/nil/services/synccommittee/internal/rpc"
 	"github.com/NilFoundation/nil/nil/services/synccommittee/internal/types"
 	"github.com/NilFoundation/nil/nil/services/synccommittee/prover/tracer"
+	"github.com/NilFoundation/nil/nil/services/synccommittee/prover/tracer/api"
 	"github.com/rs/zerolog"
 )
 
 type partialProofCmd struct {
-	client         client.Client
+	client         api.RpcClient
 	logger         zerolog.Logger
 	nilRpcEndpoint string
 	cmdCommon
@@ -25,7 +24,7 @@ type partialProofCmd struct {
 
 func NewPartialProofCmd(config CommandConfig, logger zerolog.Logger) Command {
 	return &partialProofCmd{
-		client:         rpc.NewRetryClient(config.NilRpcEndpoint, logging.NewLogger("client")),
+		client:         rpc.NewRetryClient(config.NilRpcEndpoint, logger),
 		nilRpcEndpoint: config.NilRpcEndpoint,
 		cmdCommon:      makeCmdCommon(config),
 		logger:         logger,
--- a/nil/services/synccommittee/prover/task_handler.go
+++ b/nil/services/synccommittee/prover/task_handler.go
@@ -10,9 +10,8 @@ import (
 	"strings"
 	"syscall"
 
-	"github.com/NilFoundation/nil/nil/client"
+	"github.com/NilFoundation/nil/nil/client/rpc"
 	"github.com/NilFoundation/nil/nil/common"
-	"github.com/NilFoundation/nil/nil/common/logging"
 	"github.com/NilFoundation/nil/nil/services/synccommittee/internal/api"
 	"github.com/NilFoundation/nil/nil/services/synccommittee/internal/log"
 	"github.com/NilFoundation/nil/nil/services/synccommittee/internal/storage"
@@ -27,7 +26,6 @@ type taskHandler struct {
 	timer         common.Timer
 	logger        zerolog.Logger
 	config        taskHandlerConfig
-	client        client.Client
 }
 
 func newTaskHandler(
@@ -41,7 +39,6 @@ func newTaskHandler(
 		timer:         timer,
 		logger:        logger,
 		config:        config,
-		client:        NewRPCClient(config.NilRpcEndpoint, logging.NewLogger("client")),
 	}
 }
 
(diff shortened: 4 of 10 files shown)


nil/services/synccommittee/internal/metrics/task_storage_metrics.go
```diff
@@ -86,20 +91,24 @@ func (h *taskStorageMetricsHandler) RecordTaskTerminated(ctx context.Context, ta
 	}
 }
 
-func (h *taskStorageMetricsHandler) RecordTaskRescheduled(ctx context.Context, taskEntry *types.TaskEntry) {
-	taskAttributes := getTaskAttributes(taskEntry)
+func (h *taskStorageMetricsHandler) RecordTaskRescheduled(ctx context.Context, taskType types.TaskType, previousExecutor types.TaskExecutorId) {
```
akokoshn, line 94: if keep `*types.TaskEntry` it will be easy change, when need add more info to the record (taskId, ...)

nil/services/synccommittee/internal/storage/task_storage.go
```diff
@@ -18,9 +18,11 @@ import (
 )
 
 // TaskEntriesTable BadgerDB tables, TaskId is used as a key
-const (
-	taskEntriesTable db.TableName = "task_entries"
-)
+const taskEntriesTable db.TableName = "task_entries"
```
akokoshn, line 21:
```
const (
    taskEntriesTable db.TableName = "task_entries"

    // rescheduledTasksPerTxLimit defines the maximum number of tasks that can be rescheduled
    // in a single transaction of TaskStorage.RescheduleHangingTasks.
    rescheduledTasksPerTxLimit = 100
)
```
?

nil/services/synccommittee/internal/storage/task_storage.go
```diff
@@ -342,13 +341,43 @@ func (st *taskStorage) processTaskResultImpl(ctx context.Context, res *types.Tas
 		return err
 	}
 
+	if err := res.ValidateAgainst(entry); err != nil {
+		return err
+	}
+
+	if !res.IsSuccess() && res.Error.CanBeRetried() {
```
akokoshn, line 348: What about adding `panic` in case of `!res.IsSuccess() && !res.Error.CanBeRetried()`?

nil/services/synccommittee/prover/internal/constants/proof_producer_codes.go
```diff
@@ -0,0 +1,23 @@
...
+	ProofProducerIOError      ProofProducerResultCode = 10
+	ProofProducerInvalidInput ProofProducerResultCode = 20
+	ProofProducerProverError  ProofProducerResultCode = 30
+	ProofProducerUnknownError ProofProducerResultCode = 0xFF
+)
+
+var ProofProducerErrors = map[ProofProducerResultCode]types.TaskErrType{
+	ProofProducerIOError:      types.TaskErrIO,
+	ProofProducerInvalidInput: types.TaskErrInvalidInputData,
+	ProofProducerProverError:  types.TaskErrProofFailed,
```
akokoshn, line 21: `TaskErrProofGenerationFailed`

nil/services/synccommittee/internal/storage/task_storage.go
oclaw: It will add an opportunity to collapse the service work with some malformed database entry. Maybe better add a mechanism to skip processing of such tasks? Some alerting could be of use too

zadykian: Yes, panicing here is not a good option
Critical errors will be handled by `TaskStateChangeHandler` in the next PR

Here we just need to update the task and its dependencies

nil/services/synccommittee/internal/metrics/task_storage_metrics.go
zadykian: The problem here is that value of `TaskEntry.Owner` field cannot be used as a metric attribute, because it's already reset to zero at the moment of `RecordTaskRescheduled()` call

So anyway we need to pass `previousExecutor` separately. In that case passing the whole task makes even less sense, given that we actually need only one field from it - `TaskType`

nil/services/synccommittee/prover/internal/constants/proof_producer_codes.go
zadykian: renamed

nil/services/synccommittee/internal/metrics/task_storage_metrics.go
akokoshn: got it, thanks!

akokoshn (approved):

nil/services/synccommittee/internal/types/errors.go
```diff
@@ -11,12 +14,92 @@ var (
...
 var (
-	ErrTaskInvalidStatus    = errors.New("task has invalid status")
-	ErrTaskWrongExecutor    = errors.New("task belongs to another executor")
-	ErrUnexpectedTaskType   = errors.New("unexpected task type")
-	ErrBlockProofTaskFailed = errors.New("block proof task failed")
+	ErrTaskInvalidStatus = errors.New("task has invalid status")
+	ErrTaskWrongExecutor = errors.New("task belongs to another executor")
+)
+
+type TaskErrType int8
```
oclaw, line 21: For such usage I suggest to use some kind of a registry for such cases
```
type TaskErr int8

var taskErrStr = make(map[TaskErr]string)

func RegisterTaskError(errType TaskErr, desc string) {
    if dupDesc, exists := taskErrStr[errType]; exists {
        panic("duplicate!")
    }
    taskErrStr[errType] = desc
}

var (
TaskErrTimeout = ResigterTaskError(1, "request timed out")
...
)

```
and use it for implementing `String()` and other stuff

nil/services/synccommittee/internal/types/task_result.go
```diff
@@ -34,13 +36,37 @@ type TaskResultData []byte
...
 	Data            TaskResultData      `json:"binaryData,omitempty"`
 }
 
+func (r *TaskResult) IsSuccess() bool {
+	return r.Error == nil
+}
+
+// ValidateAgainst checks the correctness of the TaskResult
+// against the given TaskEntry and returns an error if invalid.
+func (r *TaskResult) ValidateAgainst(entry *TaskEntry) error {
```
oclaw, line 51: Sounds more like `ValidateForTask`?

nil/services/synccommittee/internal/types/task_result.go
```diff
@@ -34,13 +36,37 @@ type TaskResultData []byte
...
+
+	if r.Sender == UnknownExecutorId || r.Sender != entry.Owner {
+		return fmt.Errorf(
+			"%w: taskId=%v, taskStatus=%v, taskOwner=%v, requestSenderId=%v",
+			ErrTaskWrongExecutor, entry.Task.Id, entry.Status, entry.Owner, r.Sender,
+		)
+	}
+
+	if entry.Status != Running {
+		return errTaskInvalidStatus(entry, "Validate")
```
oclaw, line 64: Let's put the actual status into the error?

nil/services/synccommittee/proofprovider/task_state_change_handler.go
```diff
@@ -43,18 +40,29 @@ func (h taskStateChangeHandler) OnTaskTerminated(ctx context.Context, task *type
 		return nil
 	}
 
-	if !result.IsSuccess {
-		log.NewTaskEvent(h.logger, zerolog.WarnLevel, task).Msgf("Prover task has failed")
+	if !result.IsSuccess() && result.Error.CanBeRetried() {
```
oclaw, line 43: Maybe add separate`HasRetryableError()` method to the result?

nil/services/synccommittee/prover/internal/constants/proof_producer_codes.go
```diff
@@ -0,0 +1,23 @@
+package constants
+
+import (
+	"github.com/NilFoundation/nil/nil/services/synccommittee/internal/types"
+)
+
+// ProofProducerResultCode represents the result codes for proof-producer binary.
```
oclaw, line 7: Let's refer to the placeholder sources here?

makxenov (approved):

nil/services/synccommittee/internal/types/errors.go
zadykian: `String()` method is already auto generated by the `stringer` tool [here](https://github.com/NilFoundation/nil/pull/1/files#diff-1ded1aa78c950d33cc1ab6e51de40c597683b2b4323998b6f7ac82dbf24b8ba9R7) + `RetryableErrors` map is declared

I'd like to stick with the current approach for now, it's simpler and compile-time safe

nil/services/synccommittee/internal/types/task_result.go
zadykian: Actual status is already included in the message [here](https://github.com/NilFoundation/nil/blob/4e597ffbf4c6d9e30236e0046b832f5f0ee17a11/nil/services/synccommittee/internal/types/prover_tasks.go#L241)

oclaw (approved):

PR #43, opened 2025-01-29
[loadgen] [WIP] Nil load generator update

Zerg1996:
Make possible to call swap externally from wallet on shard 0


olegrok (approved):

nil/services/nil_load_generator/jsonrpc.go
```diff
@@ -1,12 +1,50 @@
...
+	"github.com/NilFoundation/nil/nil/services/cliservice"
+	uniswap "github.com/NilFoundation/nil/nil/services/nil_load_generator/contracts"
 )
 
 type NilLoadGeneratorAPI interface {
 	HealthCheck() bool
 	SmartAccountsAddr() []types.Address
+	CallSwap(tokenName1, tokenName2 string, amountSwap, expectedAmount uint64) (common.Hash, error)
+	CallQuote(tokenName1, tokenName2 string, swapAmount uint64) (*big.Int, error)
+	CallInfo(hash common.Hash) (UniswapTransactionInfo, error)
```
KlonD90, line 20: It doesn't work now with in flight for progress. Transaction not found happens. It's only will work when all hapens.

nil/services/nil_load_generator/jsonrpc.go
```diff
@@ -18,13 +56,126 @@ func NewNilLoadGeneratorAPI() *NilLoadGeneratorAPIImpl {
...
+
+func getRandomSmartAccount() (uniswap.SmartAccount, *cliservice.Service, error) {
+	index, err := rand.Int(rand.Reader, big.NewInt(int64(len(uniswapSmartAccounts))))
+	if err != nil {
+		return uniswap.SmartAccount{}, nil, err
+	}
+	return uniswapSmartAccounts[index.Int64()], uniswapServices[index.Int64()], nil
+}
+
+func (c NilLoadGeneratorAPIImpl) CallSwap(tokenName1, tokenName2 string, amountSwap, expectedAmount uint64) (common.Hash, error) {
```
KlonD90, line 81: amountSwap should be check on limits (limit should be configured)

PR #69, opened 2025-01-31
Declared `rpc.CallError` type

zadykian:
Introduced a dedicated `rpc.CallError` type for RPC errors to detect them using `errors.As(...)` call

### DIFF
--- a/nil/client/rpc/client.go
+++ b/nil/client/rpc/client.go
@@ -27,13 +27,27 @@ import (
 	"github.com/rs/zerolog"
 )
 
+// CallError represents an error that occurs during a remote procedure call,
+// including the data marshalling and unmarshalling parts
+type CallError struct {
+	message string
+}
+
+func (e CallError) Error() string {
+	return e.message
+}
+
+func newCallErr(message string) error {
+	return &CallError{message}
+}
+
 var (
-	ErrFailedToMarshalRequest    = errors.New("failed to marshal request")
-	ErrFailedToSendRequest       = errors.New("failed to send request")
-	ErrUnexpectedStatusCode      = errors.New("unexpected status code")
-	ErrFailedToReadResponse      = errors.New("failed to read response")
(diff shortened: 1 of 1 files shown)


nil/client/rpc/client.go
```diff
@@ -27,21 +27,35 @@ import (
...
+// including the data marshalling and unmarshalling parts
+type CallError struct {
+	message string
+}
+
+func (e CallError) Error() string {
+	return "RPC failure: " + e.message
+}
+
+func newRpcErr(message string) error {
```
oclaw, line 40: Why do we use callErr vs RpcErr for constructing the same objects?

zadykian: Do you mean function name? Sure, `newCallErr` is more suitable

nil/client/rpc/client.go
```diff
@@ -27,21 +27,35 @@ import (
...
-	ErrUnexpectedStatusCode      = errors.New("unexpected status code")
-	ErrFailedToReadResponse      = errors.New("failed to read response")
-	ErrFailedToUnmarshalResponse = errors.New("failed to unmarshal response")
-	ErrRPCError                  = errors.New("rpc error")
+	ErrFailedToMarshalRequest    = newRpcErr("failed to marshal request")
+	ErrFailedToSendRequest       = newRpcErr("failed to send request")
+	ErrUnexpectedStatusCode      = newRpcErr("unexpected status code")
+	ErrFailedToReadResponse      = newRpcErr("failed to read response")
+	ErrFailedToUnmarshalResponse = newRpcErr("failed to unmarshal response")
+	ErrRPCError                  = newRpcErr("rpc error")
```
olegrok, line 50: Do you really need to add `RPC failure:` prefix.

We will have "RPC failure: rpc error" that's a bit weird.

zadykian: Okay, it can be omitted, `Error()` will just return the original message

olegrok (approved):

oclaw (approved):

PR #110, opened 2025-02-01
[nil_block_generator] Fix deploy contract

akokoshn:
- use `cliservice.DeployContractViaSmartAccount`
- extract correct block hash from receipt


olegrok (approved):

nil/cmd/nil_block_generator/internal/commands/client.go
```diff
@@ -130,7 +130,7 @@ func CreateNewSmartAccount(logger zerolog.Logger) (string, string, error) {
 	return smartAccount.Hex(), hexKey, nil
 }
 
-func DeployContract(path, hexKey string, logger zerolog.Logger) (string, error) {
+func DeployContract(samrtAccountAdr, path, hexKey string, logger zerolog.Logger) (string, error) {
```
makxenov, line 133: Typo: `smartAccountAddr`

nil/cmd/nil_block_generator/internal/commands/client.go
```diff
@@ -142,14 +142,24 @@ func DeployContract(path, hexKey string, logger zerolog.Logger) (string, error)
...
+	amount := types.Value0
 
 	srv, err := CreateCliService(hexKey, logger)
 	if err != nil {
 		return "", err
 	}
 
-	_, addr, err := srv.DeployContractExternal(types.BaseShardId, payload, fee)
+	var samrtAccountAddress types.Address
+	if err := samrtAccountAddress.Set(samrtAccountAdr); err != nil {
```
makxenov, line 153: Variables look the same. Maybe specify the difference in names?
```suggestion
	if err := smartAccountAddress.Set(smartAccountAddrStr); err != nil {
```

nil/cmd/nil_block_generator/main.go
```diff
@@ -88,7 +88,7 @@ func buildAddContractCmd(logger zerolog.Logger) (*cobra.Command, error) {
 			if err != nil {
 				return err
 			}
-			adr, err := commands.DeployContract(contractPath, cfg.PrivateKey, logger)
+			adr, err := commands.DeployContract(cfg.WalletAdr, contractPath, cfg.PrivateKey, logger)
```
makxenov, line 91: Let's get rid of `wallet` term in favor of `smartAccount`

akokoshn: good catch, thanks!

nil/cmd/nil_block_generator/internal/commands/client.go
akokoshn: done

makxenov (approved):

PR #111, opened 2025-02-02
Add l1 proof verifier scripts

AndreyMlashkin:
This PR adds l1 to nil and also adds it to the deb package.
It's a preliminary version


l1-proof-veriffier/Dockerfile
```diff
@@ -0,0 +1,46 @@
+# Stage 1: Build stage
+FROM node:20.3.0 AS build
```
akokoshn, line 2: Do we really need docker here?

AndreyMlashkin: Cleaned

defistar (approved):

PR #114, opened 2025-02-03
Sync Committee: Service Workers Refactoring

zadykian:
### Sync Committee: Service Workers Refactoring

* Introduced a unified service management mechanism, replacing individual `Run(ctx)` methods with a shared `srv.Service` type;
* Declared `srv.Worker` interface implemented by core components (`Aggregator`, `Proposer`, `TaskScheduler`, etc.);
* Workers are started sequentially in forward order and stopped in reverse order;

### Tests:
* Replaced tcp polling (`testaide.WaitForEndpoint` func) with channel-based signaling;

```go
type Worker interface {
	Name() string
	Run(ctx context.Context, started chan<- struct{}) error
}

type Service struct {
	workers []Worker
	logger  zerolog.Logger
}

func (s *Service) Run(ctx context.Context) error {
	// Here workers are started in forward order and stopped in reverse order
}

```


nil/services/synccommittee/internal/srv/service.go
```diff
@@ -0,0 +1,77 @@
...
+		workers: workers,
+		config:  config,
+		logger:  logger,
+	}
+}
+
+func (s *Service) Run(ctx context.Context) error {
+	defer telemetry.Shutdown(ctx)
+
+	if s.config.GracefulShutdown {
```
oclaw, line 41: Couple questions: 
- do we actually have cases where GracefulShutdown should not be used?
- It is a bit unobvious that app without graceful shutdown enabled does not even try to monitor OS signals. IMO monitoring them anyway + return from `Run` method without blocking on the errgroup makes more sense of this option

zadykian: Initially I just copied this section with `GracefulShutdown` from the old impl of `SyncCommittee.Run(...)`

In fact we don't really need it (at least now), so `SIGTERM` and `SIGINT` will be always monitored by default

nil/services/synccommittee/internal/srv/service.go
```diff
@@ -0,0 +1,77 @@
...
+		signalCtx, stop := signal.NotifyContext(ctx, syscall.SIGTERM, syscall.SIGINT)
+		defer stop()
+		ctx = signalCtx
+	}
+
+	group, ctx := errgroup.WithContext(ctx)
+
+	for _, worker := range s.workers {
+		group.Go(func() error {
+			return s.runWorker(worker, ctx)
```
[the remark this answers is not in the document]
zadykian: Now workers are started and stopped strictly according to the order

nil/services/synccommittee/internal/srv/service.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+}
+
+func (s *Service) runWorker(workerErrors chan<- error, worker Worker) (control *workerControl, shouldContinue bool) {
+	logger := WorkerLogger(s.logger, worker)
+	wCtx, cancel := context.WithCancel(context.Background())
+	control = &workerControl{cancel, make(chan struct{}), make(chan error)}
+
+	go func(wCtx context.Context, control *workerControl) {
+		logger.Info().Msg("starting worker")
```
oclaw, line 84: This might be a good place to recover from unexpected panics and log them in a proper way

nil/services/synccommittee/internal/srv/service.go
```diff
@@ -0,0 +1,139 @@
...
+	}
+}
+
+func (s *Service) runWorker(
+	ctx context.Context,
+	workerErrors chan<- error,
+	worker Worker,
+) (control *workerControl, shouldContinue bool) {
+	logger := WorkerLogger(s.logger, worker)
+	wCtx, cancel := context.WithCancel(context.Background())
```
oclaw, line 79: I suggest keep using top level context here (since it may contain some useful for logging/tracing/etc values for example) but do it in `context.WithoutCancel` way

nil/services/synccommittee/internal/srv/service.go
```diff
@@ -0,0 +1,139 @@
...
+
+		logger := WorkerLogger(s.logger, worker)
+		control.cancel()
+
+		select {
+		case <-time.After(workerStopTimeout):
+			logger.Warn().Msg("worker did not stop in time")
+
+		case err := <-control.stopped:
+			if nilOrCancelled(err) {
```
oclaw, line 128: you can use `readValue, ok := <-channel` syntax to ensure that error is nil or not

nil/services/synccommittee/internal/srv/service.go
zadykian: Added recovery with logging and termination of other workers

nil/services/synccommittee/internal/srv/service.go
zadykian: done

akokoshn (approved):

oclaw (approved):
LGTM in general
Please take a look at the last comments  :)

nil/services/synccommittee/core/aggregator.go
```diff
@@ -62,30 +63,25 @@ func NewAggregator(
...
+	agg.WorkerLoop = srv.NewWorkerLoop("aggregator", pollingDelay, agg.runOnce)
+	agg.logger = srv.WorkerLogger(logger, agg)
+	return agg
+}
 
-			if errors.Is(err, types.ErrBatchNotReady) {
-				agg.logger.Warn().Err(err).Msg("received unready block batch, skipping")
-				return
-			}
+func (agg *Aggregator) runOnce(ctx context.Context) {
```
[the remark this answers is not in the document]
zadykian: Renamed to `runIteration` across the project

nil/services/synccommittee/internal/srv/service_test.go
```diff
@@ -0,0 +1,338 @@
...
+		}
+
+		s.Run(testCase.name+"_Panics", func() {
+			s.runWorkersAndReturnErrAt(testCase, true)
+		})
+	}
+}
+
+// runWorkersAndReturnErrAt runs a specified number of workers and returns the error from a designated faulty worker.
+func (s *ServiceTestSuite) runWorkersAndReturnErrAt(testCase runAndCancelTestCase, workerPanics bool) {
```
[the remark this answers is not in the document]
zadykian: Not really, moved it to the test case field

PR #129, opened 2025-02-03
[docs] return of the switch

khannanov-nil:
This diff attempts to fix incorrect CSS loading in docs.


KlonD90 (approved):
best pr in my world

PR #134, opened 2025-02-03
[GH-#] Integrated faucet service used in testes

Un1oR:
For a considerable time now, the faucet service has been running inside a simple nild configuration. Now we will use it in tests for simplicity.


clijs/test/testEnv.ts
```diff
@@ -1,5 +1,5 @@
 const defaultRpcEndpoint = "http://127.0.0.1:8529";
-const defaultFaucetServiceEndpoint = "http://127.0.0.1:8527";
+const defaultFaucetServiceEndpoint = "http://127.0.0.1:8529";
 const defaultCometaServiceEndpoint = "http://127.0.0.1:8528";
```
olegrok, line 3: I guess you can do the same for cometa

khannanov-nil (approved):

PR #138, opened 2025-02-04
[docs] added explanation of the new fee model

khannanov-nil:
This diff adjusts the 'gas and value' tutorial to reflect the new fee model.

### DIFF
--- a/docs/nil/getting-started/gas-and-value.mdx
+++ b/docs/nil/getting-started/gas-and-value.mdx
@@ -6,8 +6,16 @@ In nil;, every transaction has two fields responsible for handling gas and value
 
 * `FeeCredit`
 * `Value`
+* `MaxPriorityFeePerGas`
+* `MaxFeePerGas`
 
-`FeeCredit` contains the amount of =nil; base tokens to be paid for transaction execution. This is different from regular Ethereum in that there is no base fee or the premium fee. Instead, `FeeCredit` is the gas limit (the maximum possible amount of gas the transaction sender is willing to pay) multiplied by the possible gas price within a specific shard.
+`FeeCredit` is the maximum possible amount of tokens the transaction sender is willing to pay for transaction execution. `Value` is the amount of =nil; base tokens to be debited to another smart account. 
+
+`MaxFeePerGas` is the limit set on the total price of gas that the sender is willing to pay for transaction processing. These two fields ensure that transaction processing on =nil; is handled similarly to Ethereum with only marginal differences.
+
+Each shard also has a `BaseFee` which is the gas price that determines whether a transaction is eligible to be included in a block. If the sender is not willing to pay at least the `BaseFee` when specifying the gas price, their transaction will be rejected. Similarly to Ethereum, `BaseFee` is history-dependent and fluctuates depending on shard congestion. 
+
+In turn, `MaxPriorityFeePerGas` refers to the maximum possible additional price per gas that should act as a reward for validators to prioritize processing a specific transaction. There can be cases where the actual priority fee is lower than the value of `MaxPriorityFeePerGas`. For example, if `BaseFee` plus `MaxPriorityFeePerGas` is higher than `MaxFeePerGas`, the validator reward will be reduced so that the total gas price matches `MaxFeePerGas`.  
 
 :::info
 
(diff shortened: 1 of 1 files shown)


docs/nil/getting-started/gas-and-value.mdx
```diff
@@ -6,8 +6,14 @@ In nil;, every transaction has two fields responsible for handling gas and value
 
 * `FeeCredit`
 * `Value`
+* `MaxPriorityFeePerGas`
+* `MaxFeePerGas`
 
-`FeeCredit` contains the amount of =nil; base tokens to be paid for transaction execution. This is different from regular Ethereum in that there is no base fee or the premium fee. Instead, `FeeCredit` is the gas limit (the maximum possible amount of gas the transaction sender is willing to pay) multiplied by the possible gas price within a specific shard.
+`FeeCredit` contains the maximum possibleamount of =nil; base tokens to be paid for transaction execution. `FeeCredit` is the gas limit (the maximum possible amount of gas the transaction sender is willing to pay) multiplied by the possible gas price within a specific shard. `Value` is the amount of =nil; base tokens to be sent to another smart account. Across async call chains, `Value` can be used to pay for transaction execution [**as detailed in this tutorial**](../smart-contracts/gas-forwarding.mdx).
```
shermike, line 12: missed space: possibleamount

shermike, line 12: > the maximum possible amount of gas the transaction sender is willing to pay

`FeeCredit` is not a gas, it is native token

docs/nil/getting-started/gas-and-value.mdx
```diff
@@ -6,8 +6,14 @@ In nil;, every transaction has two fields responsible for handling gas and value
...
+
```
shermike, line 13: > `Value` can be used to pay for transaction execution

Value is just credited into target account, it doesn't somehow relate to fee payment. It can indirectly relate to NO_FORWARDING type of transaction, because in this case feeCredit of outbound transaction is debited from the source account. But I don't think it is worth saying about this as it too unclear imho.

docs/nil/getting-started/gas-and-value.mdx
```diff
@@ -17,48 +23,123 @@ To estimate gas in a different way, use the `eth_estimateFee` JSON-RPC method or
...
 
+As shown above, `MaxFeePerGas` is always propagated across the entire async call chain. This means that the transaction sender can never be charged more than their absolute maximum gas price.
+
+:::note[Payment for sync calls]
+
+Payment for sync calls uses the same fields and the same mechanism. 
+
+Unless they trigger new async calls, sync calls do not spawn additional transactions. This makes it possible to easily revert failed sync calls. For example, if someone sends an external transaction which triggers a sync call, and the call reverts for any reason, the initial external transaction is also reverted.
+
+If a sync call triggers an async call, the resulting async call chain uses the leftover `FeeCredit` from the sync call to pay for its execution.
```
shermike, line 137: Actually, it depends on forwarding type of the transaction.

docs/nil/getting-started/gas-and-value.mdx
khannanov-nil: Sure, fixed

khannanov-nil: You are right, fixed

docs/nil/getting-started/gas-and-value.mdx
khannanov-nil: Agreed, removed

docs/nil/getting-started/gas-and-value.mdx
khannanov-nil: Fair enough, rephrased

docs/nil/getting-started/gas-and-value.mdx
```diff
@@ -17,48 +23,123 @@ To estimate gas in a different way, use the `eth_estimateFee` JSON-RPC method or
...
-    abi: CONTRACT_ABI,
-    functionName: "payable_func_name()",
-    args: [ARGS],
-  }),
-});
+If `MaxFeePerGas` is less than `BaseFee`, a transaction is bounced automatically.
+
+:::
+
+## Example
```
shermike, line 38: Probably, it is worth to show the case when `baseFee + priorityFee` > `maxFeePerGas`. In last shard they are equal, but I think it would be better to showcase when effective priorityFee is less than given one.

khannanov-nil: I don't think that the diagram needs to be modified to account for this case. I've added a note with the explanation below, however

docs/nil/getting-started/gas-and-value.mdx
```diff
@@ -17,48 +23,137 @@ To estimate gas in a different way, use the `eth_estimateFee` JSON-RPC method or
...
+    S1 --> S2
+    S2 --> S3
+    S3 --> S4
 ```
 
+As shown above, `MaxFeePerGas` is always propagated across the entire async call chain. This means that the transaction sender can never be charged more than their absolute maximum gas price.
+
+:::note[Effective priority fee]
+
+There may be cases when `BaseFee` plus the specified `MaxPriorityFeePerGas` is larger than the provided `MaxFeePerGas`. In this situation, the effective priority fee would be negative which is not allowed under the fee model rules.
```
shermike, line 133: There is also the case when effective priority fee would be just less than the given one(`MaxPriorityFeePerGas`), but not negative. E.g. `BaseFee` = 10, `MaxPriorityFeePerGas` = 3, `MaxFeePerGas` = 12, here 10 + 3 > 12 but effective priority fee will be 2, not negative.

docs/nil/getting-started/gas-and-value.mdx
```diff
@@ -17,48 +23,137 @@ To estimate gas in a different way, use the `eth_estimateFee` JSON-RPC method or
...
+
+The following formula is used to calculate the effective priority fee:
+
+$$
+max(min(MaxPriorityFeePerGas, MaxFeePerGas - BaseFee), 0)
```
shermike, line 138: I'd just left `min(MaxPriorityFeePerGas, MaxFeePerGas - BaseFee)` because if it is less than zero, it will be rejected rather than executed with zero priority fee.

docs/nil/getting-started/gas-and-value.mdx
```diff
@@ -6,8 +6,14 @@ In nil;, every transaction has two fields responsible for handling gas and value
 
 * `FeeCredit`
 * `Value`
+* `MaxPriorityFeePerGas`
+* `MaxFeePerGas`
 
-`FeeCredit` contains the amount of =nil; base tokens to be paid for transaction execution. This is different from regular Ethereum in that there is no base fee or the premium fee. Instead, `FeeCredit` is the gas limit (the maximum possible amount of gas the transaction sender is willing to pay) multiplied by the possible gas price within a specific shard.
+`FeeCredit` contains the maximum possible amount of =nil; base tokens to be paid for transaction execution. `FeeCredit` is the maximum possible amount of tokens the transaction sender is willing to pay multiplied by the possible gas price within a specific shard. `Value` is the amount of =nil; base tokens to be sent to another smart account.
```
alex-damjanovic, line 12: I think this section is unecessary: FeeCredit` is the maximum possible amount of tokens the transaction sender is willing to pay multiplied by the possible gas price within a specific shard. 

Because  FeeCredit provided  can be any value. How we estimate it in the background it is a different topic. The user can change it.

docs/nil/getting-started/gas-and-value.mdx
```diff
@@ -6,8 +6,14 @@ In nil;, every transaction has two fields responsible for handling gas and value
 
 * `FeeCredit`
 * `Value`
+* `MaxPriorityFeePerGas`
+* `MaxFeePerGas`
 
-`FeeCredit` contains the amount of =nil; base tokens to be paid for transaction execution. This is different from regular Ethereum in that there is no base fee or the premium fee. Instead, `FeeCredit` is the gas limit (the maximum possible amount of gas the transaction sender is willing to pay) multiplied by the possible gas price within a specific shard.
+`FeeCredit` contains the maximum possible amount of =nil; base tokens to be paid for transaction execution. `FeeCredit` is the maximum possible amount of tokens the transaction sender is willing to pay multiplied by the possible gas price within a specific shard. `Value` is the amount of =nil; base tokens to be sent to another smart account. 
+
+In turn, `MaxPriorityFeePerGas` refers to the additional price per gas that should act as a reward for validators to prioritize processing a specific transaction. `MaxFeePerGas` is the limit set on the total price of gas that the sender is willing to pay for transaction processing. These two fields ensure that transaction processing on =nil; is handled similarly to Ethereum with only marginal differences.
```
alex-damjanovic, line 14: * `MaxPriorityFeePerGas` refers to the **maximum** additional price per gas that can be given as as a reward for validators to prioritize processing a specific transaction. 

Comment: 
It is important to say that it is the maximum possible additional price because in a rare case where the basefee spikes the tip charged will not be the `MaxPriorityFeePerGas ` but an EffectivePriorityFeePerGas. 

Simple Example: 

MaxFeePerGas = 6 
MaxPriorityFeePerGas  = 2 
Base Fee = 5 

You cannot give more fee per gas than 6 (MaxFeePerGas) , thus the actual priorityFeePerGas charged per gas  becomes 1 instead of two.

I leave to you to explain in a way that it is most intuitive for you.

docs/nil/getting-started/gas-and-value.mdx
```diff
@@ -17,48 +23,137 @@ To estimate gas in a different way, use the `eth_estimateFee` JSON-RPC method or
...
-  data: encodeFunctionData({
-    abi: CONTRACT_ABI,
-    functionName: "func_name()",
-    args: [ARGS],
-  }),
-});
-```
+:::info
+
+If `MaxFeePerGas` is less than `BaseFee`, a transaction is bounced automatically.
```
alex-damjanovic, line 34: Comment: Maybe to mention that this is not the case for an initial external transaction. Initial external  should wait in the mempool until the validator includes it. For the transactions that come after (CSTs, async...) it bounces. 

cc @shermike

docs/nil/getting-started/gas-and-value.mdx
alex-damjanovic: Yes, this is important to showcase imo

docs/nil/getting-started/gas-and-value.mdx
alex-damjanovic: Makes sense, maybe frame it in a following way in the docs:

Assuming the maxFeePerGas > Base Fee,  the EffectivePriorityFeePerGas is calculated as follows:
min(MaxPriorityFeePerGas, MaxFeePerGas - BaseFee) . 

If MaxFeePerGas < BaseFee the transaction is rejected.

alex-damjanovic (changes requested):
I reviewed the sections, and in general, they look good. Provided a few small suggestions that might improve clarity.

docs/nil/getting-started/gas-and-value.mdx
shermike: I agree.

docs/nil/getting-started/gas-and-value.mdx
khannanov-nil: Thank you! Rephrased

docs/nil/getting-started/gas-and-value.mdx
khannanov-nil: Added, thank you!

docs/nil/getting-started/gas-and-value.mdx
khannanov-nil: Added, thanks!

docs/nil/getting-started/gas-and-value.mdx
khannanov-nil: fixed, thanks!

docs/nil/getting-started/gas-and-value.mdx
khannanov-nil: added!

docs/nil/getting-started/gas-and-value.mdx
```diff
@@ -6,8 +6,16 @@ In nil;, every transaction has two fields responsible for handling gas and value
...
+
+`MaxFeePerGas` is the limit set on the total price of gas that the sender is willing to pay for transaction processing. These two fields ensure that transaction processing on =nil; is handled similarly to Ethereum with only marginal differences.
+
+Each shard also has a `BaseFee` which is the gas price that determines whether a transaction is eligible to be included in a block. If the sender is not willing to pay at least the `BaseFee` when specifying the gas price, their transaction will be rejected. Similarly to Ethereum, `BaseFee` is history-dependent and fluctuates depending on shard congestion. 
+
+In turn, `MaxPriorityFeePerGas` refers to the maximum possible additional price per gas that should act as a reward for validators to prioritize processing a specific transaction. There can be cases where the actual priority fee is lower than the value of `MaxPriorityFeePerGas`. For example if `BaseFee` plus `MaxPriorityFeePerGas` is less than `MaxFeePerGas`, the validator reward will be reduced so that the total gas price matches `MaxFeePerGas`.
```
alex-damjanovic, line 18: For example if `BaseFee` plus `MaxPriorityFeePerGas` is **less** than `MaxFeePerGas`, 
 
 Not less but **higher**

docs/nil/getting-started/gas-and-value.mdx
```diff
@@ -17,48 +25,137 @@ To estimate gas in a different way, use the `eth_estimateFee` JSON-RPC method or
...
+      direction LR
+      SHFOUR --> OUTCOMEFOUR
+    end
+    ET --> S1
+    S1 --> S2
+    S2 --> S3
+    S3 --> S4
 ```
 
+As shown above, `MaxFeePerGas` is always propagated across the entire async call chain. This means that the transaction sender can never be charged more than their absolute maximum gas price.
```
alex-damjanovic, line 131: maybe instead just gas price add **gas price valuation**

docs/nil/getting-started/gas-and-value.mdx
```diff
@@ -17,48 +25,137 @@ To estimate gas in a different way, use the `eth_estimateFee` JSON-RPC method or
...
+
+There may be cases when `BaseFee` plus the specified `MaxPriorityFeePerGas` is larger than the provided `MaxFeePerGas`. In this situation, the effective priority fee would be different compared to `MaxPriorityFeePerGas`.
+
+The following formula is used to calculate the effective priority fee when `MaxFeePerGas` is larger than `BaseFee`:
+
+$$
+min(MaxPriorityFeePerGas, MaxFeePerGas - BaseFee)
+$$
+
+If `MaxFeePerGas` is less than `BaseFee`, the transaction is rejected.
```
alex-damjanovic, line 143: Worth mentioning that this rule does not stand for the first external transaction as we explained earlier . Not required but maybe good for precision.

alex-damjanovic (changes requested):
Small nitpicks

docs/nil/getting-started/gas-and-value.mdx
khannanov-nil: Got it, fixed!

docs/nil/getting-started/gas-and-value.mdx
khannanov-nil: This feels like it introduces some ambiguity in terms. I'd rather use gas price throughout.

alex-damjanovic: Kk, you are right.

alex-damjanovic (approved):
lgtm

shermike (approved):

Issue #125, opened 2025-02-03
Incorrect waiting for receipt in nil.js


Un1oR:
To reproduce the problem, some changes to the code are needed. I believe that a test for such a scenario should be added after the fix.
```diff
diff --git a/niljs/src/utils/receipt.ts b/niljs/src/utils/receipt.ts
index ee42a87..5638bdf 100644
--- a/niljs/src/utils/receipt.ts
+++ b/niljs/src/utils/receipt.ts
@@ -25,6 +25,7 @@ export async function waitTillCompleted(
   while (cur !== hashes.length) {
     const [hash] = hashes[cur];
     const receipt = await client.getTransactionReceiptByHash(hash);
+    console.log(receipt)
     if (!receipt) {
       await new Promise((resolve) => setTimeout(resolve, interval));
       continue;
diff --git a/niljs/test/integration/deploy.test.ts b/niljs/test/integration/deploy.test.ts
index ec82ab3..e544813 100644
--- a/niljs/test/integration/deploy.test.ts
+++ b/niljs/test/integration/deploy.test.ts
@@ -65,7 +65,7 @@ test("External deployment", async ({ expect }) => {
   const addr = bytesToHex(deploymentTransaction.to);
   expect(addr).toBeDefined();
 
-  await topUpTest(addr);
+  await topUpTest(addr, "NIL", 50_000_000);
 
   const hash = await deploymentTransaction.send(client);
 
diff --git a/niljs/test/integration/helpers.ts b/niljs/test/integration/helpers.ts
index e6f68b0..55579df 100644
--- a/niljs/test/integration/helpers.ts
+++ b/niljs/test/integration/helpers.ts
@@ -35,12 +35,13 @@ export function generatePublicKey() {
   return getPublicKey(generateRandomPrivateKey());
 }
 
-export async function topUpTest(address: Hex, token = "NIL") {
+export async function topUpTest(address: Hex, token = "NIL", amount = 1e18) {
   await topUp({
     address,
     rpcEndpoint: testEnv.endpoint,
     faucetEndpoint: testEnv.faucetServiceEndpoint,
     token,
+    amount: amount,
   });
 }

```

The point is, we're trying to deploy a contract with knowingly insufficient funds to do so.

<details>
<summary>Test log example</summary>

```
$ npm run test:integration

> @nilfoundation/niljs@0.23.0 test:integration
> CI=true vitest -c ./test/vitest.integration.config.ts --isolate false


[...]

PR #156, opened 2025-02-04
niljs: handle failed external failed

Rolaman:
Optimisation and speeding for contact deployment and sending transactions

- WaitTillMainShard default set to false
- Fix cases for failed external transactions

Fixed https://github.com/NilFoundation/nil/issues/125


knazarov:
Please squash commits, and add PR description so that people understand what's been done.

Also, if a relevant issue exists - please link it as well.

shermike (approved):

PR #158, opened 2025-02-04
[GH-65] Add clijs macOS CI build over nix

### LINKED ISSUE #65: Add a clijs building to CI under macOS
Probably the easiest way to do this would be to convert macOS builds to use nix in the meantime

Un1oR:
This PR adds `clijs` builds for macOS (x86_64 and ARM) via Nix.

The builds take a long time (about an hour) due to unconfigured caching in Nix. This will be addressed in a separate task. Builds are not marked required, so we consider this option acceptable for now. For the same reason, we are not adding other checks via Nix yet.

`install-nix-action` uses `max-jobs = auto` by default, which leads to test instability. Took a long time going through images and disabling tests until I got to the cause. The setting was changed to the standard `max-jobs = 1`.

Closes #65.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,11 +26,55 @@ jobs:
       - name: run check
         run: nix build .#checks.x86_64-linux.${{ matrix.check == 'nil' && 'others' || matrix.check }} -L
 
+  nix_check_macos:
+    name: macOS CI (${{ matrix.platform.os}}, ${{ matrix.platform.arch }}) - ${{ matrix.check }}
+    environment: prod
+    runs-on: ${{ matrix.platform.os }}
+    strategy:
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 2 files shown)


olegrok (approved):
`Successful in 68m`

Looks dangerous

.github/workflows/main.yml
```diff
@@ -26,11 +26,61 @@ jobs:
...
+      - name: Run check
+        run: nix build .#checks.${{ matrix.platform.nixArch }}.${{ matrix.check == 'nil' && 'others' || matrix.check }} -L
+
+      - name: Upload nil binary as artifact
+        if: github.event_name == 'workflow_dispatch'
+        uses: actions/upload-artifact@v4
+        with:
+          name: nil-darwin-${{ matrix.platform.arch }}
+          path: |
+            build/bin/nil
```
Un1oR, line 74: Damn, I guess the copypaste is incorrect here

PR #159, opened 2025-02-04
l1 rollup-contracts and bridge-contracts - forge and hardhat project.

defistar:
## Background

1. Nil L1 contracts contains the solidity contracts to 
   - handle commit-batch & update-state rollup operations
   - solidity contracts to handle bridge actions : L1 -> L2 & L2 -> L1 Solidity contracts

2. This solidity_forge project directory is added as a sub-directory in `nil` project

## Project Contents List
1. Solidity Contracts
2. Forge and hardhat dependencies
3. npm package file with dependencies, scripts to handle deploy and contract operations
4. formatting, linting and vulnerability and static analysis plugin and configuration files


l1-contracts/.gitmodules
```diff
@@ -0,0 +1,9 @@
+[submodule "lib/forge-std"]
```
akokoshn, line 1: we are using nix for handling external dependency

l1-contracts/README.md
```diff
@@ -0,0 +1,304 @@
+# Nil L1-Contracts
+
+This repository contains all the L1 contracts that are to be deployed on mainnent & sepolia testnet
```
akokoshn, line 3: could you please update readme. It' not a repository. See [example of internal readme style](https://github.com/NilFoundation/nil/blob/main/niljs/README.md)

defistar: updated readme and restructured the directory as `rollup-bridge-contracts`

l1-contracts/.gitmodules
defistar: removed the dependency on git modules.

rollup-bridge-contracts/contracts/bridge/libraries/Queue.sol
```diff
@@ -0,0 +1,92 @@
+// SPDX-License-Identifier: MIT
+pragma solidity 0.8.27;
```
akokoshn, line 2: Let's split this huge PR into:

1. NilRollUp only
2. Bridge implementation (on top of first one)

PR #160, opened 2025-02-04
tracer: keccak opcode tracing

oclaw:
Our proof system has independent circuit for proving KECCAK256 opcode execution. Inputs and results of these executions must be stored in traces that sync committee collects from the EVM while re-executing the blocks. 

This PR adds collecting and serializing these traces + fixes couple of bugs found while testing with the placeholder part

### DIFF
--- a/nil/services/synccommittee/prover/tracer/keccak_tracer_test.go
+++ b/nil/services/synccommittee/prover/tracer/keccak_tracer_test.go
@@ -1,0 +1,70 @@
+package tracer
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/NilFoundation/nil/nil/internal/vm"
+	"github.com/NilFoundation/nil/nil/services/synccommittee/prover/tracer/internal/testutils"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/holiman/uint256"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// traceExpOperation encapsulates the setup and invocation of tracing an KECCAK256 operation.
+func traceKeccakOperation(t *testing.T, tracer *KeccakTracer, opCode vm.OpCode, offset, size uint64, mem []byte) bool {
+	t.Helper()
+	stack := []uint256.Int{*uint256.NewInt(size), *uint256.NewInt(offset)}
+	context := &testutils.OpContextMock{
+		StackDataFunc:  func() []uint256.Int { return stack },
+		MemoryDataFunc: func() []byte { return mem },
+	}
+
+	traced, err := tracer.TraceOp(opCode, context)
+	require.NoError(t, err)
+
+	if opCode == vm.KECCAK256 {
+		stack = stack[:1]
+		hash := crypto.Keccak256(mem[offset : offset+size])
+		var u256val uint256.Int
+		u256val.SetBytes(hash)
+		stack[0] = u256val
+	}
+
+	tracer.FinishPrevOpcodeTracing()
+	return traced
+}
+
+func TestKeccakTracer_HandlesKeccakOperation(t *testing.T) {
+	t.Parallel()
+	tracer := NewKeccakTracer()
+
+	buf := bytes.Repeat([]byte{0xFF}, 8)
+	assert.True(t, traceKeccakOperation(t, tracer, vm.KECCAK256, 0, 4, buf))
+
+	require.Len(t, tracer.hashes, 1)
+	assert.Equal(t, buf[:4], tracer.hashes[0].buf)
+	expectedHash, err := uint256.FromHex("0x29045A592007D0C246EF02C2223570DA9522D0CF0F73282C79A1BC8F0BB2C238")
+	require.NoError(t, err)
+	assert.Equal(t, expectedHash.Bytes(), tracer.hashes[0].hash.Bytes())
+}
+
+func TestKeccakTracer_IgnoresOtherOperations(t *testing.T) {
+	t.Parallel()
+	tracer := NewKeccakTracer()
+
+	assert.False(t, traceKeccakOperation(t, tracer, vm.ADD, 0, 2, []byte{1, 2, 3, 4})) // Non-KECCAK256 opcode should result in no operation captured
+
+	assert.Empty(t, tracer.hashes)
+}
+
+func TestKeccakTracer_MaintainsCorrectStateAcrossCalls(t *testing.T) {
+	t.Parallel()
+	tracer := NewKeccakTracer()
+
+	assert.True(t, traceKeccakOperation(t, tracer, vm.KECCAK256, 1, 4, []byte{0xFF, 0xFF, 0xFF, 0xFF, 0xFF}))
+	assert.True(t, traceKeccakOperation(t, tracer, vm.KECCAK256, 0, 4, []byte{0xFF, 0xFF, 0xFF, 0xFF}))
+
+	require.Len(t, tracer.hashes, 2)
+}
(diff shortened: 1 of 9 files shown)


nil/services/synccommittee/prover/tracer/state_db.go
```diff
@@ -318,6 +329,7 @@ func (tsdb *TracerStateDB) initTransactionTraceContext(
 		stackTracer:   NewStackOpTracer(&tsdb.RwCounter, txnId),
 		memoryTracer:  NewMemoryOpTracer(&tsdb.RwCounter, txnId),
 		expTracer:     NewExpOpTracer(txnId),
+		keccakTracer:  NewKeccakTracer(),
```
akokoshn, line 332: why we are no using transaction index in keccak tracer?

oclaw: KECCAK256 has a specific tracing in terms of copy circuit and I believe all checks regarding the data consistency and operations order are placed there

But @AntoineCyr  and @ETatuzova definitely know better :)

AntoineCyr: That is correct. The execution, data consistency and order of operations is constrained in our proof system.

akokoshn: Oh, i see - bytecode also has no txIdx

akokoshn (approved):

nil/services/synccommittee/prover/tracer/keccak_tracer_test.go
```diff
@@ -0,0 +1,71 @@
...
+	"github.com/NilFoundation/nil/nil/internal/vm"
+	"github.com/NilFoundation/nil/nil/services/synccommittee/prover/tracer/internal/testutils"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/holiman/uint256"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// traceExpOperation encapsulates the setup and invocation of tracing an KECCAK256 operation.
+func traceKeccakOperation(t *testing.T, tracer *KeccakTracer, opCode vm.OpCode, offset, size uint64, mem []byte) bool {
```
makxenov, line 16: How about [mock client testing](https://github.com/NilFoundation/nil/blob/main/nil/services/synccommittee/prover/tracer/tracer_mock_client_test.go) instead? You won't need to reimplement tracer logic for testing, and also it will check that trace+finish are actually called.

nil/services/synccommittee/prover/tracer/keccak_tracer_test.go
```diff
@@ -0,0 +1,71 @@
...
+func TestKeccakTracer_HandlesKeccakOperation(t *testing.T) {
+	t.Parallel()
+	tracer := NewKeccakTracer()
+
+	buf := bytes.Repeat([]byte{0xFF}, 8)
+	assert.True(t, traceKeccakOperation(t, tracer, vm.KECCAK256, 0, 4, buf))
+
+	require.Len(t, tracer.hashes, 1)
+	assert.Equal(t, buf[:4], tracer.hashes[0].buf)
+	expectedHash, err := uint256.FromHex("0x29045A592007D0C246EF02C2223570DA9522D0CF0F73282C79A1BC8F0BB2C238")
```
makxenov, line 49: Is it possible to use explicit hash calculation here instead of the magic number?

nil/services/synccommittee/prover/tracer/keccak_tracer_test.go
oclaw: Having separate tracer for each possible tracer is not bad IMO
However I do not mind to extend the mock client tracer suite too, will add the test

nil/services/synccommittee/prover/tracer/keccak_tracer_test.go
oclaw: It is actually a very specific magic number used here https://www.evm.codes/#FD
This number is also going to be reused for the proof-producer testing

I can replace it with hash but maybe it would be useful to have exactly this string in our sources

nil/services/synccommittee/prover/tracer/keccak_tracer_test.go
oclaw: Done, extended copy event finalizer test

nil/services/synccommittee/prover/tracer/memory_op_tracer.go
```diff
@@ -43,8 +43,9 @@ type opRanges struct {
 var opsToMemoryRanges = map[vm.OpCode]func(stack *StackAccessor, memoryLen int) opRanges{
 	vm.KECCAK256: func(stack *StackAccessor, _ int) opRanges {
 		offset := stack.Pop()
+		lengthToRead := stack.Pop()
 		return opRanges{
-			before: memoryRange{offset.Uint64(), 32},
+			before: memoryRange{offset.Uint64(), lengthToRead.Uint64()},
```
oclaw, line 48: It was a bug

nil/services/synccommittee/prover/tracer/tracer_mock_client_test.go
```diff
@@ -338,10 +339,13 @@ func (s *TracerMockClientTestSuite) TestSwapOperations() {
 }
 
 // Check that copy event finalizer is called
-func (s *TracerMockClientTestSuite) TestCopyEventFinalizer() {
+func (s *TracerMockClientTestSuite) TestKeccakOpCodeTracing() {
 	// Check two keccak opcodes, since they are finalized in different place:
 	// we have finalizer call after the previous opcode and at the end of transaction
 	dataToCopy, err := hex.DecodeString("000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1f")
+	expectedHash := crypto.Keccak256(dataToCopy)
+	_ = expectedHash // TODO remove me
```
makxenov, line 347: Seems it's time :)

nil/services/synccommittee/prover/tracer/keccak_tracer_test.go
oclaw: Also I have used an approach with the explicit hash calc in the new mock tracer tests :)

makxenov: Just curious, how are you going to reuse this number? It sounds like testing of keccak hash function implementation, which is not a responsibility of these tests.

nil/services/synccommittee/prover/tracer/keccak_tracer_test.go
```diff
@@ -0,0 +1,71 @@
...
+	context := &testutils.OpContextMock{
+		StackDataFunc:  func() []uint256.Int { return stack },
+		MemoryDataFunc: func() []byte { return mem },
+	}
+
+	traced, err := tracer.TraceOp(opCode, context)
+	require.NoError(t, err)
+
+	if opCode == vm.KECCAK256 {
+		// NOTE(oclaw): nil EVM now operates different
```
makxenov, line 28: It's unclear to me what is different and why did it happen. Maybe add some details?

nil/services/synccommittee/prover/tracer/keccak_tracer_test.go
oclaw: It is not, but while debugging of the keccak opcode on different EVM or proof-producer stages this number might appear for quite a number of times and it also could be found here.

I do not have strong opinion here, but since this test completely mocks execution of the opcode it seems pretty reasonable to me to mock this call with some well-known real example of the keccak execution

nil/services/synccommittee/prover/tracer/keccak_tracer_test.go
oclaw: Previously we figured out that the KECCAK256 opcode was doing the Poseidon hash in reality :)
But it looks like the current main is fixed to do the real keccak now, I have also checked it with the proof-producer and everything matches. Removed the comment

oclaw:
placeholder changes: https://github.com/NilFoundation/placeholder/pull/315

makxenov (approved):

PR #170, opened 2025-02-04
Change of Assign tab to Import Contract tab

gitshreevatsa:
This PR introduces renaming of "Assign" Tab -> "Import Contract"  and a small hint under the address input to tell the user to insert the address of the already deployed smart contract(with its name).

image


ukorvl:
Rename tab from "Import contract" to "Import existing contract"

Issue #143, opened 2025-02-04
Main shard consensus


knazarov:
Main shard will have a slightly special type of consensus where it not only will process transactions as other shards do, but also it needs to set references to latest blocks on each shard.

Not only that, but also as compared to regular shards, the entire validator set votes for the main shard blocks, and so it needs a special flow in the validation code.

PR #176, opened 2025-02-05
[consensus] run consensus on main shard by all validators

olegrok:
This patch enables consensus for the main shard. Now, it is validated by all available validators.
The list of validators is simply a collection of validators from all shards without duplicates.
The main issue is the tests since the sharded suite now has to wait for the zero shard to start ticking.

Need for #143


isergeyam:
I suggest we wait until #219 merges and then rework this patch

isergeyam (approved):

PR #191, opened 2025-02-05
[consensus] Move validator config on the blockchain

isergeyam:
Here, we move the validators' configuration on the blockchain. This will allow us to modify the configuration using smart contracts in the future. Right now, this configuration is fixed and needs to be provided at the zero state.


nil/internal/execution/zerostate.go
```diff
@@ -157,6 +157,27 @@ func (es *ExecutionState) GenerateZeroStateYaml(configYaml string) error {
...
+	} else {
+		rightConfig = &ZeroStateConfig{}
+	}
+	if leftConfig == nil {
+		leftConfig = &ZeroStateConfig{}
+	}
+	return es.GenerateZeroState(
+		&ZeroStateConfig{
+			ConfigParams: leftConfig.ConfigParams,
+			Contracts:    rightConfig.Contracts,
```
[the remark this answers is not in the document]
isergeyam: Right now we used the yaml config for contracts, but the logic of generating ConfigParams is done separately. I would actually favour a solution when we remove ZeroStateYaml completely and serialize it inside the tests framework to use just ZeroState, I think we should rework it that way, this is just a workaround

nil/internal/consensus/ibft/ibft.go
```diff
@@ -25,11 +25,10 @@ type ConsensusParams struct {
 	Validator  validator
 	NetManager *network.Manager
 	PrivateKey *ecdsa.PrivateKey
-	Validators []config.ValidatorInfo
 }
 
 type validator interface {
-	BuildProposal(ctx context.Context) (*execution.Proposal, error)
```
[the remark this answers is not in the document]
olegrok: Probably the answer is no. Since we anyway have mainshard replicated and mainShardHash as a part of proposal.

nil/internal/consensus/ibft/ibft.go
```diff
@@ -58,7 +58,7 @@ func (i *backendIBFT) unmarshalProposal(raw []byte) (*execution.Proposal, error)
 }
 
 func (i *backendIBFT) BuildProposal(view *protoIBFT.View) []byte {
-	proposal, err := i.validator.BuildProposal(i.ctx)
+	proposal, err := i.validator.BuildProposal(i.ctx, i.tx)
```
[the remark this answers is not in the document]
isergeyam: right, fixed

nil/internal/collate/proposer.go
```diff
@@ -129,26 +122,25 @@ func (p *proposer) fetchPrevBlock() error {
...
 			shardId := types.ShardId(i)
-			lastBlockHash, err := db.ReadLastBlockHash(p.roTx, shardId)
+			lastBlockHash, err := db.ReadLastBlockHash(tx, shardId)
 			if err != nil && !errors.Is(err, db.ErrKeyNotFound) {
 				return err
 			}
 
 			p.proposal.ShardHashes[i-1] = lastBlockHash
 		}
-	} else {
```
olegrok, line 144: Seems you need to add `return` here to keep logic the same

nil/internal/collate/syncer.go
```diff
@@ -135,15 +136,16 @@ func (s *Syncer) Run(ctx context.Context, wgFetch *sync.WaitGroup) error {
...
+		return fmt.Errorf("Failed to generate zerostate for shard %s: %w", s.config.ShardId, err)
+	}
+
 	// Wait until snapshots for shards are fetched.
 	// It's impossible to load data and commit transactions at the same time.
+	// We also wait for zerostate generation, because otherwise consensus won't be able to access the config.
 	wgFetch.Done()
 	wgFetch.Wait()
 
-	if err := s.generateZerostate(ctx); err != nil {
```
olegrok, line 143: No. You need a separate sync primitive here for that because it's easy to get transaction conflict if you try to generate zerostate but some other syncer fetch data at the same time.

nil/internal/collate/syncer.go
```diff
@@ -157,6 +159,7 @@ func (s *Syncer) Run(ctx context.Context, wgFetch *sync.WaitGroup) error {
 	s.logger.Debug().
 		Stringer(logging.FieldBlockHash, hash).
 		Uint64(logging.FieldBlockNumber, uint64(block.Id)).
+		Uint32(logging.FieldShardId, uint32(s.config.ShardId)).
```
olegrok, line 162: It already contains this field

https://github.com/NilFoundation/nil/pull/191/files#diff-ab230848b41511019d4484b9165904556e6b86d41eef2bf72c8279933b73b1b8L63

nil/internal/collate/syncer.go
```diff
@@ -210,6 +213,7 @@ func (s *Syncer) processTopicTransaction(ctx context.Context, data []byte) (bool
 	s.logger.Debug().
 		Stringer(logging.FieldBlockNumber, block.Id).
 		Stringer(logging.FieldBlockHash, block.Hash(s.config.ShardId)).
+		Uint32(logging.FieldShardId, uint32(s.config.ShardId)).
```
olegrok, line 216: The same as above

nil/internal/consensus/ibft/ibft.go
```diff
@@ -74,6 +84,11 @@ func (i *backendIBFT) InsertProposal(proposal *protoIBFT.Proposal, committedSeal
 	if err != nil {
 		return
 	}
+	i.logger.Debug().
+		Uint64(logging.FieldBlockNumber, proposalBlock.PrevBlockId.Uint64()+1).
```
olegrok, line 88: It's better to use "Height" for consensus

nil/internal/consensus/ibft/ibft.go
```diff
@@ -74,6 +84,11 @@ func (i *backendIBFT) InsertProposal(proposal *protoIBFT.Proposal, committedSeal
...
+		Uint64(logging.FieldRound, proposal.Round).
+		Uint32(logging.FieldShardId, uint32(i.shardId)).
```
olegrok, line 90: This field is already defined here

https://github.com/NilFoundation/nil/blob/b3a123a44ede3cffe12b327f0d1ef762db9d6ca5/nil/internal/consensus/ibft/ibft.go#L99

nil/internal/consensus/ibft/ibft.go
```diff
@@ -102,32 +117,63 @@ func NewConsensus(cfg *ConsensusParams) *backendIBFT {
...
+		return nil
+	}
+	return i.setupTransport(ctx)
+}
+
 func (i *backendIBFT) GetVotingPowers(height uint64) (map[string]*big.Int, error) {
-	result := make(map[string]*big.Int, len(i.validators))
-	for _, v := range i.validators {
+	// Here we take the latest config, but we should take the config based ot proposer that isn't available at this point
+	// TODO(@isergeyam): I think we should rewrite the ibft/core part to get voting powers after the proposer is calculated
```
olegrok, line 142: I think we don't need to touch ibft/core.
An approach I guess should be following:
- We definitely know only the last committed block (that can be source of truth)
- This last block contains a link to main shard
- We can use validators from this point

That leads to the case that we have 2 blocks lag between current state and config. But it shouldn't be a problem since we generate new blocks each 5s. So 10 seconds is relatively quick period for that.
And also we don't need to wait for some sync with main hash. We definitely have (height-2) main block in our DB.

But there is a bit different issue. We need to generate zerostate in a bit different way. At the first time we need to generate main shard zerostate and only then we need to generate zerostates for other shards.

nil/internal/consensus/ibft/transport.go
```diff
@@ -63,6 +63,13 @@ func (i *backendIBFT) setupTransport(ctx context.Context) error {
 					continue
 				}
 
+				if data == nil {
+					i.logger.Debug().
+						Str(logging.FieldTopic, protocol).
+						Msg("Received nil data")
```
olegrok, line 69:
```suggestion
						Msg("Received empty message")
```

nil/internal/consensus/ibft/ibft.go
```diff
@@ -25,26 +27,26 @@ type ConsensusParams struct {
...
+	transportCtx context.Context
+	db           db.DB
+	consensus    *core.IBFT
+	shardId      types.ShardId
+	validator    validator
+	logger       zerolog.Logger
+	nm           *network.Manager
+	transport    transport
+	signer       *Signer
+	mainBlockMap sync.Map
```
olegrok, line 49: I guess you need to clean this main for each new sequence

nil/internal/consensus/ibft/ibft.go
isergeyam: > . We definitely have (height-2) main block in our DB.

I don't think that's true, but I agree with the idea of using shardChain[height-1].mainBlockHash, done

nil/internal/collate/syncer.go
isergeyam: My bad, done

nil/tests/sharded_suite.go
```diff
@@ -69,21 +69,34 @@ func (s *ShardedSuite) Cancel() {
 	}
 }
 
+func zeroState(validators []config.ListValidators) *execution.ZeroStateConfig {
```
olegrok, line 72: mb `newZeroState`?

nil/internal/collate/syncer.go
```diff
@@ -143,6 +144,7 @@ func (s *Syncer) Run(ctx context.Context, wgFetch *sync.WaitGroup) error {
 	if err := s.generateZerostate(ctx); err != nil {
 		return fmt.Errorf("Failed to generate zerostate for shard %s: %w", s.config.ShardId, err)
 	}
+	wgZeroState.Done()
```
olegrok, line 147: To be honest it looks quite strange now. Let's move zerostate generation out of syncer then. You anyway need to generate them in a synchronous way. At first you need to generate it for shard 0. And then specify its hash as a mainShardHash to other shards.

nil/internal/consensus/ibft/ibft.go
```diff
@@ -58,14 +58,22 @@ func (i *backendIBFT) unmarshalProposal(raw []byte) (*execution.Proposal, error)
...
+	tx, err := i.db.CreateRoTx(i.ctx)
 	if err != nil {
 		return nil
 	}
+	defer tx.Rollback()
+
+	proposal, err := i.validator.BuildProposal(i.ctx, tx)
+	if err != nil {
+		return nil
+	}
```
olegrok, line 70: Do you still need this change?

nil/internal/consensus/ibft/ibft.go
```diff
@@ -102,32 +117,77 @@ func NewConsensus(cfg *ConsensusParams) *backendIBFT {
...
+
+	var mainBlockHash *common.Hash
+
+	// TODO(@isergeyam): propagate mainchain zerostate hash to shardchains' zerostate, then remove this condition
+	if height > 1 {
+		block, err := db.ReadBlockByNumber(tx, i.shardId, types.BlockNumber(height-1))
+		if err != nil {
+			return nil, err
+		}
+		if i.shardId == types.MainShardId {
```
olegrok, line 172:
```suggestion
		if i.shardId.IsMainShard() {
```

nil/internal/consensus/ibft/transport.go
```diff
@@ -63,6 +63,13 @@ func (i *backendIBFT) setupTransport(ctx context.Context) error {
 					continue
 				}
 
+				if data == nil {
+					i.logger.Debug().
+						Str(logging.FieldTopic, protocol).
+						Msg("Received empty message")
+					continue
+				}
```
olegrok, line 71: It's better to move such check before `isActiveValidator`

olegrok: Also I suggest to change log level to trace

nil/internal/consensus/ibft/verifier.go
```diff
@@ -26,7 +26,18 @@ func (i *backendIBFT) IsValidValidator(msg *protoIBFT.IbftMessage) bool {
 		return false
 	}
 
-	index := slices.IndexFunc(i.validators, func(v config.ValidatorInfo) bool {
+	// Here we use transportCtx because this method could be called from the transport goroutine
```
olegrok, line 29: What's a problem to use simple `ctx`?

isergeyam: There will a data race between updating `i.ctx` and validating messages

olegrok: Could you please show me a trace? We shouldn't have it...
At least after this commit https://github.com/NilFoundation/nil/commit/7d966f6a39b124e3a54c737b6ab01a67e22c7c83

nil/internal/collate/syncer.go
isergeyam: Introduced `initSyncers`

nil/internal/signer/block_verifier.go
```diff
@@ -5,37 +5,65 @@ import (
...
+		return fmt.Errorf("%w: failed to create config accessor: %w", errBlockVerify, err)
+	}
+
+	validatorsList, err := config.GetParamValidators(accessor)
 	if err != nil {
 		return fmt.Errorf("%w: failed to get validators set: %w", errBlockVerify, err)
 	}
 
+	if b.shardId == 0 {
+		b.shardId = 0
```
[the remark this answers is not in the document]
isergeyam: Debug lefovers, removed

nil/internal/consensus/ibft/verifier.go
isergeyam: RunSequence runs concurrently with the transport goroutine, and the transport goroutine calls `IsValidValidator` inside `AddMessage`, that's where the race happens

isergeyam: It is actually right to use `transportCtx` in `IsValidValidator` whenever it is being called from `AddMessage`

olegrok: > RunSequence runs concurrently with the transport goroutine, and the transport goroutine calls IsValidValidator inside AddMessage, that's where the race happens

Well... Can we preload list of validators to use inside "IsValidValidator"?

nil/internal/consensus/ibft/ibft.go
```diff
@@ -75,6 +83,13 @@ func (i *backendIBFT) InsertProposal(proposal *protoIBFT.Proposal, committedSeal
 		return
 	}
 
+	logger := i.logger.With().
+		Uint64(logging.FieldHeight, proposalBlock.PrevBlockId.Uint64()+1).
+		Uint64(logging.FieldRound, proposal.Round).
+		Logger()
+
+	logger.Debug().Msg("Inserting proposal")
```
olegrok, line 91: Let's use trace here

nil/internal/consensus/ibft/ibft.go
```diff
@@ -102,32 +117,77 @@ func NewConsensus(cfg *ConsensusParams) *backendIBFT {
...
+	if err != nil {
+		return nil, err
+	}
+
+	validatorsList, err := config.GetParamValidators(configAccessor)
+	if err != nil {
+		return nil, err
+	}
+
+	return validatorsList.Validators[i.shardId].List, nil
```
olegrok, line 190: Let's preload them once at start and use cached value

nil/services/nilservice/service.go
```diff
@@ -461,34 +486,44 @@ func createShards(
...
-		}
-		for i := range cfg.NShards {
-			cfg.Validators[types.ShardId(i)] = []config.ValidatorInfo{
-				{PublicKey: config.Pubkey(pubkey)},
+	if cfg.ZeroState == nil {
+		cfg.ZeroState = &execution.ZeroStateConfig{}
+	}
+
+	if !cfg.SplitShards {
+		if len(cfg.ZeroState.ConfigParams.Validators.Validators) == 0 {
```
olegrok, line 506: Let's implement a getter for that

```suggestion
		if len(cfg.ZeroState.GetValidators()) == 0 {
```

nil/services/nilservice/service.go
```diff
@@ -193,6 +193,24 @@ func validateArchiveNodeConfig(cfg *Config, nm *network.Manager) error {
...
+	var wgFetch sync.WaitGroup
+	wgFetch.Add(len(syncers))
+	for _, syncer := range syncers {
+		go func() {
+			check.PanicIfErr(syncer.FetchSnapshot(ctx, &wgFetch))
+		}()
+	}
+	wgFetch.Wait() // Wait for snapshots to avoid data races in DB
+	for _, syncer := range syncers {
+		if err := syncer.GenerateZerostate(ctx); err != nil {
```
olegrok, line 206: You need to generate zerostate for main shard and then pass its hash to other shards as MainShardHash

isergeyam: https://github.com/NilFoundation/nil/blob/c3d231f5d226102369303412b9245eb3d09a6a52/nil/internal/execution/block_generator.go#L161

nil/internal/consensus/ibft/verifier.go
isergeyam: Introduced `validatorsMap`

nil/internal/consensus/ibft/validators.go
```diff
@@ -1,10 +1,109 @@
...
+
+	var mainBlockHash *common.Hash
+
+	block, err := db.ReadBlockByNumber(tx, v.shardId, types.BlockNumber(v.height-1))
+	if err != nil {
+		return nil, err
+	}
+	if v.shardId.IsMainShard() {
+		hash := block.Hash(v.shardId)
+		mainBlockHash = &hash
```
olegrok, line 87: Why not `prevBlock`?

nil/internal/signer/block_verifier.go
```diff
@@ -5,37 +5,61 @@ import (
...
+	var accessor config.ConfigAccessor
+	_, err = db.ReadBlock(tx, types.MainShardId, block.MainChainHash)
+	if err != nil {
+		// It is possible that the needed main chain block has not arrived yet, or that this one is some byzantine block.
+		// Because right now the config is actually constant, we can use whatever version we like in this case,
+		// so we use the latest accessible config.
+		// TODO(@isergeyam): create some subscription mechanism that will handle this correctly.
+		if errors.Is(err, db.ErrKeyNotFound) {
+			logger.Warn().
+				Str(logging.FieldBlockHash, block.MainChainHash.String()).
```
olegrok, line 45:
```suggestion
				Stringer(logging.FieldBlockHash, block.MainChainHash).
```

nil/internal/signer/block_verifier.go
```diff
@@ -5,37 +5,61 @@ import (
...
+	}
+
+	validatorsList, err := config.GetParamValidators(accessor)
 	if err != nil {
 		return fmt.Errorf("%w: failed to get validators set: %w", errBlockVerify, err)
 	}
 
 	// TODO: for now check that block is signed by one known validator
-	for _, v := range validators.List {
+	for _, v := range validatorsList.Validators[b.shardId].List {
```
olegrok, line 62: Could you please introduce a function `getValidatorsForShards` or something similar. It will help for https://github.com/NilFoundation/nil/pull/176. Because for main shard we will have different rules.

nil/services/nilservice/service.go
olegrok: Indeed!

nil/tests/archive_node/archive_node_test.go
```diff
@@ -41,11 +41,11 @@ func (s *SuiteArchiveNode) TestGetDebugBlock() {
 		b, err := debugBlock.DecodeSSZ()
 		s.Require().NoError(err)
 
-		s.Eventually(func() bool {
+		s.Require().Eventually(func() bool {
```
olegrok, line 44: Is it still relevant?

nil/internal/consensus/ibft/validators.go
isergeyam: Here I am actually not so sure, let's use PrevBlock but I think we should discuss this tomorrow

nil/internal/consensus/ibft/ibft.go
```diff
@@ -35,16 +33,17 @@ type validator interface {
...
-	consensus  *core.IBFT
-	shardId    types.ShardId
-	validator  validator
-	logger     zerolog.Logger
-	nm         *network.Manager
-	transport  transport
-	signer     *Signer
-	validators []config.ValidatorInfo
+	ctx             context.Context
+	transportCtx    context.Context
```
olegrok, line 37: Please, describe difference between ctx and transportCtx in a comment

olegrok (approved):

PR #193, opened 2025-02-05
Add niljs/ prefix to niljs Client-Version header

ukorvl:


olegrok (approved):

nil/services/rpc/internal/http/server.go
```diff
@@ -98,8 +98,9 @@ func (s *Server) validateRequest(r *http.Request) (int, error) {
 
 	// Niljs is supported by server
 	header := r.Header.Get(nilJsVersionHeader)
-	if header != "" {
-		version, err := semver.NewVersion(header)
+	parsedHeader := header[len(niljsClientVersionPrefix):]
```
olegrok, line 101: please check that header length is enough. Otherwise it's easy to crash server by invalid header

nil/tests/basic/basic_test.go
```diff
@@ -697,13 +697,13 @@ func (s *SuiteRpc) TestUnsupportedCliVersion() {
 func (s *SuiteRpc) TestUnsupportedNiljsVersion() {
 	logger := zerolog.New(os.Stderr)
 	s.Run("Unsupported version", func() {
-		client := rpc_client.NewClientWithDefaultHeaders(s.Endpoint, logger, map[string]string{"Client-Version": "0.0.1"})
+		client := rpc_client.NewClientWithDefaultHeaders(s.Endpoint, logger, map[string]string{"Client-Version": "niljs/0.0.1"})
 		_, err := client.ChainId(s.Context)
 		s.Require().ErrorContains(err, "unexpected status code: 426: specified niljs version 0.0.1, minimum supported is")
 	})
```
olegrok, line 703: You can add new test that checks that `"Client-Version": "abc"` is handled properly

PR #195, opened 2025-02-05
prover: implement token methods, gas price, fix copy tracer on deploy

x-mass:
To enhance the tracing capabilities for a broader set of contracts, several new methods need to be implemented: `AddToken`, `SubToken`, `SetTokenTransfer`, `AddRefund`, `SubRefund`, and `GetGasPrice`. This pull request enables us to gather traces from every block produced by the `nil_load_generator`. 
Additionally, a bug in the copy tracer has been fixed; previously, there was a discrepancy between the execution EVM and the tracer EVM during the deployment transaction.

### DIFF
--- a/nil/internal/execution/state_accessor.go
+++ b/nil/internal/execution/state_accessor.go
@@ -218,6 +218,7 @@ type rawBlockAccessorResult struct {
 	receipts        fieldAccessor[[][]byte]
 	childBlocks     fieldAccessor[[]common.Hash]
 	dbTimestamp     fieldAccessor[uint64]
+	config          fieldAccessor[map[string][]byte]
 }
 
 func (r rawBlockAccessorResult) Block() []byte {
@@ -244,6 +245,10 @@ func (r rawBlockAccessorResult) DbTimestamp() uint64 {
 	return r.dbTimestamp()
 }
 
+func (r rawBlockAccessorResult) Config() map[string][]byte {
+	return r.config()
+}
+
 type rawBlockAccessor struct {
 	rawShardAccessor    *rawShardAccessor
 	withInTransactions  bool
@@ -251,6 +256,7 @@ type rawBlockAccessor struct {
 	withReceipts        bool
 	withChildBlocks     bool
 	withDbTimestamp     bool
+	withConfig          bool
 }
 
 func (b rawBlockAccessor) WithChildBlocks() rawBlockAccessor {
@@ -275,6 +281,11 @@ func (b rawBlockAccessor) WithReceipts() rawBlockAccessor {
 
 func (b rawBlockAccessor) WithDbTimestamp() rawBlockAccessor {
 	b.withDbTimestamp = true
+	return b
+}
+
+func (b rawBlockAccessor) WithConfig() rawBlockAccessor {
+	b.withConfig = true
 	return b
 }
 
@@ -318,6 +329,7 @@ func (b rawBlockAccessor) ByHash(hash common.Hash) (rawBlockAccessorResult, erro
 		receipts:        notInitialized[[][]byte]("Receipts"),
 		childBlocks:     notInitialized[[]common.Hash]("ChildBlocks"),
 		dbTimestamp:     notInitialized[uint64]("DbTimestamp"),
+		config:          notInitialized[map[string][]byte]("Config"),
 	}
 
 	if b.withInTransactions {
@@ -370,6 +382,17 @@ func (b rawBlockAccessor) ByHash(hash common.Hash) (rawBlockAccessorResult, erro
 		res.dbTimestamp = initWith(ts)
 	}
 
+	// config is included only for main shard, empty for others
+	if b.withConfig {
+		root := mpt.NewDbReader(sa.tx, sa.shardId, db.ConfigTrieTable)
+		root.SetRootHash(block.ConfigRoot)
+		configMap := make(map[string][]byte)
+		for key, value := range root.Iterate() {
+			configMap[string(key)] = value
+		}
+		res.config = initWith(configMap)
+	}
+
 	return res, nil
 }
 
@@ -391,6 +414,7 @@ type blockAccessorResult struct {
 	receipts        fieldAccessor[[]*types.Receipt]
 	childBlocks     fieldAccessor[[]common.Hash]
 	dbTimestamp     fieldAccessor[uint64]
+	config          fieldAccessor[map[string][]byte]
 }
 
 func (r blockAccessorResult) Block() *types.Block {
@@ -417,6 +441,10 @@ func (r blockAccessorResult) DbTimestamp() uint64 {
 	return r.dbTimestamp()
 }
 
+func (r blockAccessorResult) Config() map[string][]byte {
+	return r.config()
+}
+
 type blockAccessor struct {
 	rawBlockAccessor
 }
@@ -441,6 +469,10 @@ func (b blockAccessor) WithDbTimestamp() blockAccessor {
 	return blockAccessor{b.rawBlockAccessor.WithDbTimestamp()}
 }
 
+func (b blockAccessor) WithConfig() blockAccessor {
+	return blockAccessor{b.rawBlockAccessor.WithConfig()}
+}
+
 func (b blockAccessor) ByHash(hash common.Hash) (blockAccessorResult, error) {
 	sa := b.rawBlockAccessor.rawShardAccessor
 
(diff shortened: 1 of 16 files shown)


knazarov:
If this PR is not a draft, it needs a description that explains "why" this change is made and goes into details of key changes.

nil/services/synccommittee/prover/tracer/utils.go
```diff
@@ -17,3 +17,16 @@ func getDataOverflowSafe(data []byte, start uint64, size uint64) []byte {
...
 }
+
+// getDataIfStartValid returns a slice from `data` starting at `start` with a length of `size` bytes.
+// It returns nil if `size` is zero or if `start` is out of bounds.
+// Note: It does not check if `start+size` exceeds `data` length.
+func getDataIfStartValid(data []byte, start uint64, size uint64) []byte {
+	if size == 0 {
+		return nil
+	}
+	if len(data) > int(start) {
```
akokoshn, line 28: `len(data) > int(start + size)`?

nil/services/synccommittee/prover/tracer/state_db.go
```diff
@@ -378,12 +386,29 @@ func (tsdb *TracerStateDB) GetTransactionFlags() types.TransactionFlags {
...
+	if acc == nil {
+		return nil
+	}
+
+	res := make(map[types.TokenId]types.Value)
+	for k, v := range acc.TokenTree.Iterate() {
+		var c types.TokenBalance
+		c.Token = types.TokenId(k)
+		if err := c.Balance.UnmarshalSSZ(v); err != nil {
+			tsdb.logger.Error().Err(err).Msg("failed to unmarshal token balance")
```
oclaw, line 401: If SSZ serialization is somehow broken this code will silently return empty map instead of explicit indicating the problem. Are you sure we want this behavior?

nil/services/synccommittee/prover/tracer/state_db.go
```diff
@@ -489,17 +563,20 @@ func (tsdb *TracerStateDB) SetCode(addr types.Address, code []byte) error {
...
-	panic("not implemented")
+func (tsdb *TracerStateDB) AddRefund(gas uint64) {
+	tsdb.refund += gas
 }
 
-func (tsdb *TracerStateDB) SubRefund(uint64) {
-	panic("not implemented")
+func (tsdb *TracerStateDB) SubRefund(gas uint64) {
+	if gas > tsdb.refund {
+		panic(fmt.Sprintf("Refund counter below zero (gas: %d > refund: %d)", gas, tsdb.refund))
```
oclaw, line 572: Let's keep using the `check` package

nil/internal/execution/transactions.go
```diff
@@ -35,10 +36,10 @@ func (m dummyPayer) String() string {
 
 type transactionPayer struct {
 	transaction *types.Transaction
-	es          *ExecutionState
+	es          vm.StateDB
```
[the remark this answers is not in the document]
x-mass: https://github.com/NilFoundation/nil/pull/195/files#diff-ef608f1d479b197325b2c9a185d11ff98ba73c29a27782792cd151630879f65eR115

nil/services/synccommittee/prover/tracer/state_db.go
x-mass: Just the same flow as in original [state.go](https://github.com/NilFoundation/nil/blob/9d091d81a06cf437a18318fa39308364650699b2/nil/internal/execution/state.go#L1642). If there will be such issue, we need to copy the behaviour.

nil/services/synccommittee/prover/tracer/utils.go
x-mass: In this case, the conditions appear to be interchangeable, as the logic underpinning the execution of the EVM remains consistent. If the start is valid, then the end of the range must also be valid; otherwise, the execution would have already failed.
https://github.com/NilFoundation/nil/blob/9d091d81a06cf437a18318fa39308364650699b2/nil/internal/vm/memory.go#L85

akokoshn: The [check](https://github.com/NilFoundation/nil/blob/9d091d81a06cf437a18318fa39308364650699b2/nil/internal/vm/memory.go#L85) also looks strange.
If assume that `size` was checked before, condition `len(data) > int(start + size)` more strict than required. At the same time not leads over calculation and looks better for understand.

nil/services/synccommittee/prover/tracer/state_db.go
```diff
@@ -489,17 +563,18 @@ func (tsdb *TracerStateDB) SetCode(addr types.Address, code []byte) error {
...
-	panic("not implemented")
+func (tsdb *TracerStateDB) SubRefund(gas uint64) {
+	check.PanicIff(gas > tsdb.refund, "Refund counter below zero (gas: %d > refund: %d)", gas, tsdb.refund)
+	tsdb.refund -= gas
 }
 
 // GetRefund returns the current value of the refund counter.
 func (tsdb *TracerStateDB) GetRefund() uint64 {
-	panic("not implemented")
+	panic("unreachable") // Not used in ExecutionState for now
```
akokoshn, line 577: if PR still in progress, let move to draft

nil/services/synccommittee/prover/tracer/state_db.go
```diff
@@ -698,3 +773,15 @@ func (tsdb *TracerStateDB) FinalizeTraces() error {
 	tsdb.Stats.AffectedContractsN = uint(len(mptTraces.ContractTrieTraces))
 	return nil
 }
+
+func (es *TracerStateDB) updateGasPrice(txn *types.Transaction) {
```
akokoshn, line 777: why not `tsdb`?

nil/services/synccommittee/prover/tracer/state_db.go
x-mass: I meant, it is not used in `ExecuitonState` neither

akokoshn (commented):
We also need implementation for `GetConfigAccessor`, panic fails on unit tests `TestTracerNildTestSuite` after enable feature which should be in main soon (https://github.com/NilFoundation/nil/pull/249)

nil/internal/execution/state_accessor.go
```diff
@@ -370,6 +383,29 @@ func (b rawBlockAccessor) ByHash(hash common.Hash) (rawBlockAccessorResult, erro
 		res.dbTimestamp = initWith(ts)
 	}
 
+	if b.withConfig {
+		mainChainHash := block.MainChainHash
+		if sa.shardId.IsMainShard() {
+			mainChainHash = block.Hash(sa.shardId)
```
olegrok, line 389: For consistency it's better to use `PrevBlockHash`

nil/services/rpc/jsonrpc/types.go
```diff
@@ -93,6 +93,7 @@ type DebugRPCBlock struct {
 	OutTransactions []hexutil.Bytes        `json:"outTransactions"`
 	Receipts        []hexutil.Bytes        `json:"receipts"`
 	Errors          map[common.Hash]string `json:"errors"`
+	Config          map[string][]byte      `json:"config"`
```
olegrok, line 96: You need to return at least `map[string]hexutil.Bytes`. Or ideally you need to decode to some human-readable format.

akokoshn (commented):
ok for me. Let's wait approve from @olegrok

nil/services/rpc/jsonrpc/types.go
```diff
@@ -93,6 +94,60 @@ type DebugRPCBlock struct {
 	OutTransactions []hexutil.Bytes        `json:"outTransactions"`
 	Receipts        []hexutil.Bytes        `json:"receipts"`
 	Errors          map[common.Hash]string `json:"errors"`
+	Config          ChainConfig            `json:"config"`
```
olegrok, line 97: Let's use pointer here. Because this field is empty for non-main shard

x-mass: It's not empty, we fetch config from corresponding `block.MainChainHash`

nil/internal/execution/state_accessor.go
```diff
@@ -370,6 +383,23 @@ func (b rawBlockAccessor) ByHash(hash common.Hash) (rawBlockAccessorResult, erro
...
+
+		configTrie, err := config.GetConfigTrie(sa.tx, &mainChainHash)
+		if err != nil {
+			return rawBlockAccessorResult{}, fmt.Errorf("failed to create config trie: %w", err)
+		}
+		configMap := make(map[string][]byte)
+		for k, v := range configTrie.Iterate() {
+			configMap[string(k)] = v
+		}
+		res.config = initWith(configMap)
```
olegrok, line 400:
```suggestion
                root := mpt.NewDbReader(sa.tx, sa.shardId, db.ConfigTrieTable)
		root.SetRootHash(block.ConfigRoot)

		configMap := make(map[string][]byte)
		for key, value := range root.Iterate() {
			configMap[string(key)] = value
		}

		res.config = initWith(configMap)
```

PR #199, opened 2025-02-06
[tests] move cli tests that don't require nil server

olegrok:
Such tests can be run in parallel without waiting for consensus, zerostate generation, etc. This patch moves them into separate file to speedup testing process.


Zerg1996 (approved):
Thanks for PR!

PR #203, opened 2025-02-06
[tests] speedup faucet test suites

olegrok:
Just change `SetupTest` for `SetupSuite` for independent test. Also here we remove empty gas_price test. It's better to add a new one on demand in future.


nil/tests/gas_price/gas_price_test.go
Zerg1996, line 1: We don't need this test anymore?

Zerg1996 (approved):
Thanks for a PR! Just interesting why we remove one test.

nil/tests/gas_price/gas_price_test.go
olegrok: This test is empty (after new fee model is implemented).

https://github.com/NilFoundation/nil/blob/a8416a8f450c5ae8bfee3825169d9f30e6a77f51/nil/tests/gas_price/gas_price_test.go#L33-L35

As I mentioned in commit message in future we can restore this test on demand.

PR #207, opened 2025-02-06
[tests] speedup economy test

olegrok:
All tests are independent. It allows us to run nil server only once for all tests. Also `CollatorTickPeriodMs` is decreased a bit.


shermike (approved):
Nice!

PR #208, opened 2025-02-06
Introduce L1 Block fetcher

shermike:
L1 Block fetcher is a service that fetches information about Ethereum blockchain and stores it in the on-chain config.
The workflow is as follows:
1. Poll Ethereum RPC nodes for new blocks at some interval.
2. Once new block is fetched, create new transaction to `L1BlockInfo` contract with L1blockInfo in calldata.
3. `L1BlockInfo` contract writes L1 block info to the config parameter `l1info`.
4. L1 block info is updated and stored in the on-chain config. Go to 1.

### DIFF
--- a/nil/internal/collate/proposer.go
+++ b/nil/internal/collate/proposer.go
@@ -10,10 +10,13 @@ import (
 	"github.com/NilFoundation/nil/nil/common/check"
 	"github.com/NilFoundation/nil/nil/common/logging"
 	"github.com/NilFoundation/nil/nil/internal/config"
+	"github.com/NilFoundation/nil/nil/internal/contracts"
 	"github.com/NilFoundation/nil/nil/internal/db"
 	"github.com/NilFoundation/nil/nil/internal/execution"
 	"github.com/NilFoundation/nil/nil/internal/types"
+	"github.com/NilFoundation/nil/nil/services/rollup"
 	"github.com/NilFoundation/nil/nil/services/txnpool"
+	l1types "github.com/ethereum/go-ethereum/core/types"
 	"github.com/rs/zerolog"
 )
 
@@ -37,6 +40,8 @@ type proposer struct {
 	executionState *execution.ExecutionState
 
 	ctx context.Context
+
+	l1BlockFetcher rollup.L1BlockFetcher
 }
 
 func newProposer(params Params, topology ShardTopology, pool TxnPool, logger zerolog.Logger) *proposer {
@@ -56,10 +61,11 @@ func newProposer(params Params, topology ShardTopology, pool TxnPool, logger zer
 		params.MaxForwardTransactionsInBlock = defaultMaxForwardTransactionsInBlock
 	}
 	return &proposer{
-		params:   params,
-		topology: topology,
-		pool:     pool,
-		logger:   logger,
+		params:         params,
+		topology:       topology,
+		pool:           pool,
+		logger:         logger,
+		l1BlockFetcher: params.L1Fetcher,
 	}
 }
 
@@ -97,6 +103,11 @@ func (p *proposer) GenerateProposal(ctx context.Context, txFabric db.DB) (*execu
 
 	if err := p.fetchLastBlockHashes(tx); err != nil {
 		return nil, fmt.Errorf("failed to fetch last block hashes: %w", err)
+	}
+
+	if err := p.handleL1Attributes(tx); err != nil {
+		// TODO: change to Error severity once Consensus/Proposer increase time intervals
+		p.logger.Trace().Err(err).Msg("Failed to handle L1 attributes")
 	}
 
 	if err := p.handleTransactionsFromNeighbors(tx); err != nil {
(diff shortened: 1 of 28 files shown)


nil/internal/config/params.go
```diff
@@ -68,13 +86,31 @@ type ParamGasPrice struct {
...
 
 func (p *ParamGasPrice) Accessor() *ParamAccessor {
 	return CreateAccessor[ParamGasPrice]()
 }
 
+type ParamL1BlockInfo struct {
+	Number      uint64        `json:"number" yaml:"number"`
+	Timestamp   uint64        `json:"timestamp" yaml:"timestamp"`
+	BaseFee     types.Uint256 `json:"baseFee" yaml:"baseFee"`
+	BlobBaseFee types.Uint256 `json:"blobBaseFee" yaml:"blobBaseFee"`
```
olegrok, line 100: Please check serialization of `Uint256`. Maybe you need to add `ssz-size:"32"` as in other places

nil/services/nilservice/service.go
```diff
@@ -313,6 +314,11 @@ func CreateNode(ctx context.Context, name string, cfg *Config, database db.DB, i
 		return nil, err
 	}
 
+	if cfg.L1Fetcher == nil && cfg.IsShardActive(types.MainShardId) &&
```
olegrok, line 317: In fact `cfg.IsShardActive(types.MainShardId)` is always true

nil/services/rollup/eip-4844.go
```diff
@@ -0,0 +1,44 @@
...
+		_, overflow = numeratorAccum.MulDivOverflow(numeratorAccum, numerator, divisor)
+		if overflow {
+			return nil, fmt.Errorf("FakeExponential: overflow in MulDivOverflow(numeratorAccum=%v, numerator=%v, divisor=%v)", numeratorAccum, numerator, divisor)
+		}
+	}
+	return output.Div(output, denom), nil
+}
+
+func GetBlobGasPrice(excessBlobGas uint64) (*uint256.Int, error) {
+	return FakeExponential(uint256.NewInt(MinBlobGasPrice), uint256.NewInt(BlobGasPriceUpdateFraction), excessBlobGas)
```
olegrok, line 43: Isn't it possible to wrap geth implementation?

https://github.com/ethereum/go-ethereum/blob/913fee4be9e7b825bf6c6ff87918dde93a1be7ac/consensus/misc/eip4844/eip4844.go#L83

nil/internal/config/params.go
shermike: No, it should work as big.Int. I patched ssz lib some time ago about this.
Anyway, I've added a test case for this.

nil/services/rollup/eip-4844.go
shermike: Sure. Switched to using `geth` directly.

nil/internal/collate/proposer.go
```diff
@@ -150,6 +161,80 @@ func (p *proposer) fetchLastBlockHashes(tx db.RoTx) error {
...
+			}
+		}
+	}
+
+	txn, err := CreateL1BlockUpdateTransaction(block)
+	if err != nil {
+		return fmt.Errorf("failed to create L1 block update transaction: %w", err)
+	}
+
+	p.logger.Trace().Msgf("Add L1 block update transaction")
```
olegrok, line 195: Maybe let's add message hash to log entry?
```suggestion
	p.logger.Trace().Msg("Add L1 block update transaction")
```

nil/services/rollup/l1_fetcher.go
```diff
@@ -0,0 +1,114 @@
...
+	"https://eth-mainnet.public.blastapi.io",
+	"https://rpc.ankr.com/eth",
+	"https://eth-mainnet.public.blastapi.io",
+	"https://rpc.flashbots.net",
+	"https://cloudflare-eth.com",
+}
+
+func NewL1BlockFetcherRpc(ctx context.Context) L1BlockFetcher {
+	p := &L1BlockFetcherRpc{
+		ctx:    ctx,
```
olegrok, line 43: I think it's better to pass context directly to each function. You don't have any complex callback systems to keep context as a struct field.

nil/internal/collate/proposer.go
```diff
@@ -150,6 +161,80 @@ func (p *proposer) fetchLastBlockHashes(tx db.RoTx) error {
...
+
+	p.proposal.InternalTxns = append(p.proposal.InternalTxns, txn)
```
olegrok, line 197: Will you plan to add some logic to `VerifyProposal` or somewhere else to make sure that proposed L1 block number is correct?

shermike: Yes, there are thoughts about verifying the content of the proposed L1 transaction across validators. But no concrete plans. For now, we can stick this centralized version and make further improvements in the future.

olegrok (approved):

PR #211, opened 2025-02-06
[execution] use specific main shard hash from block generation

olegrok:
It's not correct to rely on the latest block.
Main block can be committed earlier than current one that can lead to the case when we read data from a newer source.
We fixed almost the same issue before for `collectGasPrices` function because it affected consensus. This part can also affect it in some ways.

### DIFF
--- a/nil/internal/collate/syncer.go
+++ b/nil/internal/collate/syncer.go
@@ -403,7 +403,7 @@ func (s *Syncer) generateZerostate(ctx context.Context) error {
 
 	s.logger.Info().Msg("Generating zero-state...")
 
-	gen, err := execution.NewBlockGenerator(ctx, s.config.BlockGeneratorParams, s.db)
+	gen, err := execution.NewBlockGenerator(ctx, s.config.BlockGeneratorParams, s.db, &common.EmptyHash)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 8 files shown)


nil/internal/collate/syncer.go
```diff
@@ -403,7 +403,7 @@ func (s *Syncer) generateZerostate(ctx context.Context) error {
 
 	s.logger.Info().Msg("Generating zero-state...")
 
-	gen, err := execution.NewBlockGenerator(ctx, s.config.BlockGeneratorParams, s.db)
+	gen, err := execution.NewBlockGenerator(ctx, s.config.BlockGeneratorParams, s.db, &common.EmptyHash)
```
shermike, line 406: Why not just `nil`?

shermike: Maybe I'm missing something, but as I see, `mainShardHash` in `NewBlockGenerator` is used only for creation of ConfigAccessor. But there is no difference between nil and EmptyHash in ConfigAccessor initialization: https://github.com/NilFoundation/nil/blob/refs/heads/main/nil/internal/config/config.go#L239

shermike (approved):

nil/internal/collate/syncer.go
olegrok: Indeed. Thanks for pointing on this. It's not expected behavior. Fixed config accessor logic a bit.

PR #218, opened 2025-02-07
[types] exclude bloom from receipt/block hashes

### LINKED ISSUE #52: Exclude bloom filters from the block / block hash
We can already verify the correctness of events / logs by checking merkle proof of inclusion to the block.

But the bloom filter data structure is difficult to prove in context of ZK, so it should be safe to exclude it.

olegrok:
That's purely internal change that affects only hashes. From user POV nothing was changed.

Closes #52


knazarov:
Please add motivation to the description of this PR. It's always a good idea to explain why a certain change has been introduced. Otherwise it's hard to understand when looking back at these things some time later.

nil/internal/types/block.go
```diff
@@ -46,14 +46,14 @@ type BlockData struct {
...
 	ConfigRoot          common.Hash      `json:"configRoot" ch:"config_root"`
-	LogsBloom           Bloom            `json:"logsBloom" ch:"logs_bloom"`
 	Timestamp           uint64           `json:"timestamp" ch:"timestamp"`
 	BaseFee             Value            `json:"gasPrice" ch:"gas_price"`
 	GasUsed             Gas              `json:"gasUsed" ch:"gas_used"`
 }
 
 type Block struct {
 	BlockData
+	LogsBloom Bloom     `json:"logsBloom" ch:"logs_bloom"`
```
knazarov, line 56: This is still in the block, and influences the block hash. This means that it still has to be proven by the ZK proof system. What we should do instead is to store it in a separate table "aside", with key equal to the block hash.

shermike: Block hash is calculated from `BlockData`: `common.MustPoseidonSSZ(&b.BlockData)`. LogsBloom has been moved out from it, so it seems that it doesn't influence the block hash.

shermike (approved):

olegrok:
> Please add motivation to the description of this PR. It's always a good idea to explain why a certain change has been introduced. Otherwise it's hard to understand when looking back at these things some time later.

I expected that a reference to the issue should be enough. But ok, I updated description.

PR #221, opened 2025-02-07
synccomittee(proposer): update to new L1 rollup contract, add blob tx

x-mass:
- Update L1 rollup contract abi, adjusted proposer code to use updated version (including batch tx creation). 
- Add offchain checks for `UpdateState` input data to avoid `execution reverted` errors. 

There are two kind of errors that are ignored while submitting blob and state updating. If the batch is already commited, we skip `commitBatch` call, if it is already finalized, we skip `updateState` as well.

### DIFF
--- a/nil/services/synccommittee/core/proposer.go
+++ b/nil/services/synccommittee/core/proposer.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"math/big"
 	"time"
 
 	"github.com/NilFoundation/nil/nil/common"
@@ -15,6 +14,7 @@ import (
 	"github.com/NilFoundation/nil/nil/services/synccommittee/internal/storage"
 	scTypes "github.com/NilFoundation/nil/nil/services/synccommittee/internal/types"
 	ethtypes "github.com/ethereum/go-ethereum/core/types"
+	"github.com/ethereum/go-ethereum/crypto/kzg4844"
 	"github.com/rs/zerolog"
 )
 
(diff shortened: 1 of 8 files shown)


nil/services/synccommittee/core/proposer.go
```diff
@@ -199,37 +194,104 @@ func (p *Proposer) getLatestProvedStateRoot(ctx context.Context) (common.Hash, e
 }
 
 func (p *Proposer) sendProof(ctx context.Context, data *scTypes.ProposalData) error {
```
akokoshn, line 196: let's split into: `CommitBatch` and `UpdateState`

akokoshn (approved):

nil/services/synccommittee/core/proposer.go
```diff
@@ -199,37 +194,104 @@ func (p *Proposer) getLatestProvedStateRoot(ctx context.Context) (common.Hash, e
...
 	}
 
+	blobTxSidecar := tx.BlobTxSidecar()
+	dataProofs, err := rollupcontract.ComputeDataProofs(blobTxSidecar)
+	if err != nil {
+		return err
+	}
+
+	// to make sure proofs are correct before submission, not necessary
+	if err := p.rollupContract.VerifyDataProofs(ctx, blobTxSidecar.BlobHashes(), dataProofs); err != nil {
```
akokoshn, line 248: Data proof verification should be part of `UpdateState`

akokoshn (approved):

PR #231, opened 2025-02-07
Add lavamoat to prevent install scripts

ukorvl:
This patch adds `@lavamoat/allow-scripts` package to prevent npm lifecycle scripts running without direct intention of a developer.
This is one of the steps of improving our npm security. Restricting npm lifecycle scripts decreases the possibility of supply chain attacks using malicious npm packages.

How it works? By adding `ignore-scripts=true` to `.npmrc` we are restricting all lifecycle scripts by default, ensuring that we don't run occasionally a malicious one. Lavamoat package runs all allowlisted scripts separately.

`@lavamoat/preinstall-always-fail` always fails on installation and is set to `false` in lavamoat config to ensure that no script runs unintentionally.


KlonD90 (approved):

ukorvl:
pnpm does the same thing by default. see `pnpm.ignoredBuiltDependencies` `package.json` filed.

PR #232, opened 2025-02-07
Remove barrel files from niljs

ukorvl:
All barrel files are removed - they a). produce circular dependencies b). increase build time. All circular dependencies are gone now.
```json
"typescript.preferences.importModuleSpecifier": "relative",
"typescript.preferences.autoImportFileExcludePatterns": ["**/index.ts"]
```
are set to help devs in vscode not to import from barrel file. `biome` linting will now fail if there is an import from a barrel file.

Lots of unnecessary staff is removed from niljs, which was unused - `IDeployData` duplicate type, unused block utilities, unused assert helpers, some other unused types like `ISignedMessage`.

`"typescript.tsdk": "node_modules/typescript/lib"` added to vscode setting to make dev process better, cause developer can have outdated typescript version in vscode, and we started to use native `NoInfer` in contract-factory (available since Typescript 5.4).

Smart-account is moved from `./contracts` dir and the dir was deleted. Actually we no longer store smart-contracts bytecode in niljs, so the `contracts` name is incorrect. Better to store smart-account logic in `smart-accounts`.


KlonD90 (approved):
great feature

PR #233, opened 2025-02-08
[syncer] Fix syncer

shermike:
The following is fixed:
- Don't replay block if its MainShardBlock was not porcessed in this node yet. Otherwise, on-chain config will be incorect as it is bound to the main block.
- Commit replayed block(write to database) only after validation succeed. To be able to do that, some refactoring in block building was made.

Also, Syncer's logger name was changed - not it contains synced shard and active shards. It allows to better understand which syncer the log belongs to.

### DIFF
--- a/nil/internal/collate/syncer.go
+++ b/nil/internal/collate/syncer.go
@@ -216,7 +216,7 @@ func (s *Syncer) processTopicTransaction(ctx context.Context, data []byte) (bool
 	if block.Id != lastBlock.Id+1 {
 		s.logger.Debug().
 			Stringer(logging.FieldBlockNumber, block.Id).
-			Msgf("Received block is out of order with the last block %d", lastBlock.Id)
+			Msgf("Received block %d is out of order with the last block %d", block.Id, lastBlock.Id)
 
 		// todo: queue the block for later processing
 		return false, nil
@@ -376,7 +376,12 @@ func (s *Syncer) saveDirectly(ctx context.Context, block *types.BlockWithExtract
 			block.Block.OutTransactionsRoot, txnRoot)
 	}
 
-	_, err = execution.PostprocessBlock(tx, s.config.ShardId, block.Block.BaseFee, blockHash)
+	blockRes := &execution.BlockGenerationResult{
+		Block:     block.Block,
+		BlockHash: blockHash,
+	}
+
+	err = execution.PostprocessBlock(tx, s.config.ShardId, blockRes)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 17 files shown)


nil/internal/collate/syncer.go
```diff
@@ -245,13 +261,12 @@ func (s *Syncer) processTopicTransaction(ctx context.Context, data []byte) (bool
 func (s *Syncer) fetchBlocks(ctx context.Context) {
 	// todo: fetch blocks until the queue (see todo above) is empty
 	for {
-		s.logger.Debug().Msg("Fetching next block")
-
```
shermike, line 249: These logs were removed because they look like trace logs and I don't see why they are needed.

nil/internal/collate/syncer.go
```diff
@@ -54,12 +56,26 @@ func NewSyncer(cfg SyncerConfig, db db.DB, networkManager *network.Manager) (*Sy
...
+	}
+
 	return &Syncer{
 		config:         cfg,
 		topic:          topicShardBlocks(cfg.ShardId),
 		db:             db,
 		networkManager: networkManager,
-		logger: logging.NewLogger("sync").With().
+		logger: logging.NewLogger(loggerName).With().
 			Stringer(logging.FieldShardId, cfg.ShardId).
```
olegrok, line 79: Seems you don't need to add "ShardId" to name because you have the same tag for each log entry

nil/internal/collate/syncer.go
olegrok: Let's just change its level to trace

nil/internal/collate/syncer.go
shermike: It just seems for me more convenient to have a component name with embedded shard id, because it is easier to filter it from other logs in the output. You don't need to find `shardId=[2]` string in the log message (which can be very long btw) of all sync's logs.
Maybe I'm mistaken and there are some pitfalls with embedded shard id?

olegrok: Yes. You are right. I agree that it's more useful but it adds some kind of duplication in log entries

nil/internal/collate/syncer.go
```diff
@@ -22,6 +22,8 @@ import (
 )
 
 type SyncerConfig struct {
+	IsArchive     bool
```
olegrok, line 25: This field seems unused

nil/internal/collate/syncer.go
```diff
@@ -22,6 +22,8 @@ import (
...
+	MyShards      []uint
```
olegrok, line 26: This is also unused as well

nil/internal/execution/postprocess.go
```diff
@@ -6,28 +6,22 @@ import (
 	"github.com/NilFoundation/nil/nil/internal/types"
 )
 
-func PostprocessBlock(tx db.RwTx, shardId types.ShardId, defaultGasPrice types.Value, blockHash common.Hash) (*types.Block, error) {
-	postprocessor, err := newBlockPostprocessor(tx, shardId, defaultGasPrice, blockHash)
-	if err != nil {
-		return nil, err
+func PostprocessBlock(tx db.RwTx, shardId types.ShardId, blockResult *BlockGenerationResult) error {
+	postprocessor := blockPostprocessor{tx, shardId, blockResult}
+	if blockResult.Block == nil {
```
olegrok, line 11: Is it required for tests?

PR #234, opened 2025-02-08
[mpt] Refactoring to introduce SSZ for sparse tries

dmtrskv:
We need proofs for internal transactions in the proposal.
Since they come in batches in consecutive order, storing their proofs is possible via sparse mpts (containing only the desired paths).
An mpt can be stored as a key-value map. Since serialization requires order which is not provided by the standard map, a wrapper is introduced.

The rest of the PR is a light refactoring of mpt and proposal.

1. Map wrapper in `ssz` package.
2. Simplified mpt ctors.
3. Renamed ssz package to sszx, because it conflicts with fastssz in generated code (in my future code, not here).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,6 +31,7 @@ test: generated
 
 $(COMMANDS): %: generated %.cmd
 
+include nil/common/sszx/Makefile.inc
 include nil/internal/db/Makefile.inc
 include nil/internal/mpt/Makefile.inc
 include nil/internal/types/Makefile.inc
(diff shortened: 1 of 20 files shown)


nil/internal/execution/proposal.go
```diff
@@ -8,15 +8,15 @@ import (
 )
 
 type Proposal struct {
```
olegrok, line 10: We need json tags for a proposal.

https://github.com/NilFoundation/nil/blob/b3a123a44ede3cffe12b327f0d1ef762db9d6ca5/nil/internal/execution/block_generator.go#L180

olegrok (approved):

Un1oR:
Oh, I [tried FlatMap](https://github.com/NilFoundation/nil-old/pull/857/commits/87a44c702ef7da7397275e6c0b30d134bb29afb2) when I wanted to stuff dictionaries into SSZ. But in the end `fastssz` couldn't chew so many generics 😒, and luckily we gave up SSZ as a container for network transmission in favor of Protobuf.

PR #239, opened 2025-02-10
[syncer] use stream instead of request-response

olegrok:
Such approach should be more network-friendly than request blocks by batches. Here we send blocks one by one in a following way: [header (message size)] [message].

### DIFF
--- a/nil/internal/collate/block_listener.go
+++ b/nil/internal/collate/block_listener.go
@@ -2,16 +2,22 @@ package collate
 
 import (
 	"context"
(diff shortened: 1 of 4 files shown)


olegrok:
> all blocks after this one

You read blocks from the stream. You can close this stream once you understand you downloaded enough blocks and change a peer.

olegrok:
> 1. I'd like to see it in a comment near the actual encoding/decoding.

Done

avm (approved):

Issue #62, opened 2025-01-30
Add block headers and merkle proofs to message batches produced by leader


knazarov:
When block validation happens, the leader may have information that the rest of the validators of that shard don't have yet. For example, it can be slightly ahead and receive cross-shard transactions from other shards that nobody else has. In this case, the leader should be able to persuade the other validators that these transactions are legit.

The only way to do so is to send along with the transactions the signed block header where such transactions originate, and also merkle proofs of inclusion of those messages into the block. 

Right now we don't do that, and we need data structures and logic that:

- can compose such batches with additional supporting information
- a function that can validate those batches at the start block execution

If message authenticity cannot be validated, the production of block should be aborted (validators shouldn't put their signature on it)

PR #242, opened 2025-02-10
[proposal] Add parent block headers and txn proofs

dmtrskv:
https://github.com/NilFoundation/nil/issues/62
Fulfilling the requirement of internal txn proofs in the proposal.

### DIFF
--- a/nil/internal/collate/proposer.go
+++ b/nil/internal/collate/proposer.go
@@ -13,6 +13,7 @@ import (
 	"github.com/NilFoundation/nil/nil/internal/contracts"
 	"github.com/NilFoundation/nil/nil/internal/db"
 	"github.com/NilFoundation/nil/nil/internal/execution"
+	"github.com/NilFoundation/nil/nil/internal/mpt"
 	"github.com/NilFoundation/nil/nil/internal/types"
 	"github.com/NilFoundation/nil/nil/services/rollup"
 	"github.com/NilFoundation/nil/nil/services/txnpool"
@@ -36,7 +37,7 @@ type proposer struct {
 
 	logger zerolog.Logger
 
-	proposal       *execution.Proposal
+	proposal       *execution.ProposalSSZ
 	executionState *execution.ExecutionState
 
 	ctx context.Context
@@ -69,8 +70,8 @@ func newProposer(params *Params, topology ShardTopology, pool TxnPool, logger ze
 	}
 }
 
-func (p *proposer) GenerateProposal(ctx context.Context, txFabric db.DB) (*execution.Proposal, error) {
-	p.proposal = execution.NewEmptyProposal()
+func (p *proposer) GenerateProposal(ctx context.Context, txFabric db.DB) (*execution.ProposalSSZ, error) {
+	p.proposal = &execution.ProposalSSZ{}
 
 	tx, err := txFabric.CreateRoTx(ctx)
 	if err != nil {
@@ -119,8 +120,12 @@ func (p *proposer) GenerateProposal(ctx context.Context, txFabric db.DB) (*execu
 		return nil, fmt.Errorf("failed to handle transactions from pool: %w", err)
 	}
 
-	p.logger.Debug().Msgf("Collected %d internal, %d external (%d gas) and %d forward transactions",
-		len(p.proposal.InternalTxns), len(p.proposal.ExternalTxns), p.executionState.GasUsed, len(p.proposal.ForwardTxns))
+	if len(p.proposal.InternalTxnRefs) == 0 && len(p.proposal.ExternalTxns) == 0 && len(p.proposal.ForwardTxnRefs) == 0 {
+		p.logger.Trace().Msg("No transactions collected")
+	} else {
+		p.logger.Debug().Msgf("Collected %d internal, %d external (%d gas) and %d forward transactions",
+			len(p.proposal.InternalTxnRefs), len(p.proposal.ExternalTxns), p.executionState.GasUsed, len(p.proposal.ForwardTxnRefs))
+	}
 
 	return p.proposal, nil
 }
@@ -195,7 +200,7 @@ func (p *proposer) handleL1Attributes(tx db.RoTx) error {
 		Stringer("base_fee", block.BaseFee).
 		Msg("Add L1 block update transaction")
 
-	p.proposal.InternalTxns = append(p.proposal.InternalTxns, txn)
+	p.proposal.SpecialTxns = append(p.proposal.SpecialTxns, txn)
 
 	return nil
 }
@@ -349,9 +354,11 @@ func (p *proposer) handleTransactionsFromNeighbors(tx db.RoTx) error {
 
 	checkLimits := func() bool {
 		return p.executionState.GasUsed < p.params.MaxInternalGasInBlock &&
-			len(p.proposal.InternalTxns) < p.params.MaxInternalTransactionsInBlock &&
-			len(p.proposal.ForwardTxns) < p.params.MaxForwardTransactionsInBlock
+			len(p.proposal.InternalTxnRefs) < p.params.MaxInternalTransactionsInBlock &&
+			len(p.proposal.ForwardTxnRefs) < p.params.MaxForwardTransactionsInBlock
 	}
+
+	var parents []*execution.ParentBlock
 
 	for _, neighborId := range p.topology.GetNeighbors(p.params.ShardId, p.params.NShards, true) {
 		position, ok := neighborIndexes[neighborId]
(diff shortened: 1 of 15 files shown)


nil/internal/collate/proposer.go
```diff
@@ -269,6 +270,8 @@ func (p *proposer) handleTransactionsFromNeighbors() error {
 			len(p.proposal.ForwardTxns) < p.params.MaxForwardTransactionsInBlock
 	}
 
+	var parents []*execution.ParentBlock
```
olegrok, line 273: For https://github.com/NilFoundation/nil/pull/242 (at least for main shard) we need block headers of all parents (or at least their gas prices).

To be able to use value from proposal instead of read DB 
https://github.com/NilFoundation/nil/blob/728124d4e4d6b95fe763feaaca67328f5fb3533a/nil/internal/execution/block_generator.go#L129

nil/internal/execution/proposal.go
```diff
@@ -4,21 +4,72 @@ import (
...
+		Block:        b.Block,
+		TxnMPT:       mpt.NewMPTFromMap(holder),
+		txnMPTHolder: holder,
+	}
+}
+
+func (pb *ParentBlock) ToSerializable() *ParentBlockSSZ {
+	return &ParentBlockSSZ{
+		Block:  pb.Block,
+		TxnMPT: ssz.NewMapHolder(pb.txnMPTHolder),
```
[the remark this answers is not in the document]
dmtrskv: The MPT is a map internally. And here I put a sparse MPT proving all txns from this block.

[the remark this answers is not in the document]
dmtrskv: The pairs in the map are not the keys and values that were put into the mpt, they are nodes; this way we store the actual mpt with full paths (but only for certain txns, so it's sparse).

nil/internal/collate/proposer.go
dmtrskv: I suggest adding the required blocks to this slice later, it doesn't look like a problem to me.
But I wouldn't like to add this code here, because the PR is already complicated.

nil/internal/execution/state.go
```diff
@@ -1219,7 +1219,10 @@ func decodeRevertTransaction(data []byte) string {
 
 func (es *ExecutionState) handleRefundTransaction(_ context.Context, transaction *types.Transaction) error {
 	err := es.AddBalance(transaction.To, transaction.Value, tracing.BalanceIncreaseRefund)
-	logger.Debug().Err(err).Msgf("Refunded %s to %v", transaction.Value, transaction.To)
+	logger.Debug().Err(err).
+		Stringer(logging.FieldTransactionHash, transaction.Hash()).
+		Stringer(logging.FieldTransactionTo, transaction.To).
+		Msgf("Refunded %s", transaction.Value)
```
dmtrskv, line 1225: updating this log helped me to find a bug

nil/tests/economy/economy_test.go
```diff
@@ -245,368 +245,397 @@ func (s *SuiteEconomy) TestGasForwarding() { //nolint
...
+		s.Require().True(info.ContainsOnly(s.smartAccountAddress, s.testAddress1))
+		s.checkBalance(info, initialBalance)
 	})
-	data = s.AbiPack(s.abiTest, "testForwarding", args)
-	receipt = s.SendTransactionViaSmartAccountNoCheck(s.smartAccountAddress, s.testAddress1, execution.MainPrivateKey, data,
-		feePack, types.NewValueFromUint64(0), nil)
-	info = s.AnalyzeReceipt(receipt, s.namesMap)
-	s.Require().False(info.AllSuccess())
-	s.Require().True(info.ContainsOnly(s.smartAccountAddress, s.testAddress1))
-	s.checkBalance(info, initialBalance)
```
dmtrskv, line 609: this test failed
having it split into runs helps debugging a lot

nil/internal/collate/proposer.go
```diff
@@ -198,7 +203,7 @@ func (p *proposer) handleL1Attributes(tx db.RoTx) error {
 		Stringer("base_fee", block.BaseFee).
 		Msg("Add L1 block update transaction")
 
-	p.proposal.InternalTxns = append(p.proposal.InternalTxns, txn)
+	p.proposal.SpecialTxns = append(p.proposal.SpecialTxns, txn)
```
dmtrskv, line 206: @shermike please take a look
Other internal transactions now have proofs, because they originate in certain blocks of certain shards.
This one doesn't have such a proof. Tho it might have a different kind of proof added later. I put it in a separate slice for now.

nil/internal/collate/proposer.go
```diff
@@ -390,6 +397,29 @@ func (p *proposer) handleTransactionsFromNeighbors(tx db.RoTx) error {
...
 			outTxnTrie.SetRootHash(block.OutTransactionsRoot)
+
+			saveProof := func() (*execution.InternalTxnReference, error) {
+				if len(parents) == 0 ||
+					parents[len(parents)-1].ShardId != neighborId ||
+					parents[len(parents)-1].Block.Id != block.Id {
+					parents = append(parents, execution.NewParentBlock(neighborId, block))
+				}
+
+				blockIndex := uint32(len(parents) - 1)
```
olegrok, line 408: if `len(parents) == 0` you have an overflow

nil/internal/execution/proposal.go
```diff
@@ -53,3 +119,51 @@ func SplitOutTransactions(transactions []*types.Transaction, shardId types.Shard
...
+	if err != nil {
+		return nil, fmt.Errorf("invalid internal transactions: %w", err)
+	}
+	forwardTxns, err := ConvertTxnRefs(proposal.ForwardTxnRefs, parentBlocks)
+	if err != nil {
+		return nil, fmt.Errorf("invalid forward transactions: %w", err)
+	}
+
+	// todo: special txns should be validated
+	internalTxns = append(internalTxns, proposal.SpecialTxns...)
```
shermike, line 156: I think `SpecialTxns` should be placed before internal transactions. At least the only one special transaction at the moment - L1BlockInfo - should be before other transactions, hence they can see updated L1 attributes during execution.
It is not how it will be 100%, especially if we will use config with delay of N-blocks. But still, as an initial point I think it is better to place them before internal transactions.

nil/internal/collate/proposer.go
shermike: Sure, they will be validated somehow in the future. Maybe through bridge machinery, or something else, I don't know yet.
But yes, moving them to separate slice is a good idea.

nil/internal/execution/state.go
```diff
@@ -1219,7 +1219,10 @@ func decodeRevertTransaction(data []byte) string {
 
 func (es *ExecutionState) handleRefundTransaction(_ context.Context, transaction *types.Transaction) error {
 	err := es.AddBalance(transaction.To, transaction.Value, tracing.BalanceIncreaseRefund)
-	logger.Debug().Err(err).Msgf("Refunded %s to %v", transaction.Value, transaction.To)
+	logger.Debug().Err(err).
+		Stringer(logging.FieldTransactionHash, transaction.Hash()).
```
olegrok, line 1223: Hash here is calculated even if we don't log a message itself

nil/internal/mpt/proof.go
```diff
@@ -166,6 +167,16 @@ func PopulateMptWithProof(mpt *MerklePatriciaTrie, p *Proof) error {
 	return nil
 }
 
+func ValidateHolder(holder InMemHolder) error {
```
dmtrskv, line 170: It appeared that a separate validation is required.
Proofs work by one, because the sparse MPT is constructed out of values and their hashes (thus manipulated values will have incorrect keys).

Added tests on that.

nil/internal/execution/state.go
dmtrskv: Reverted this change altogether

nil/internal/collate/proposer.go
```diff
@@ -388,6 +395,29 @@ func (p *proposer) handleTransactionsFromNeighbors(tx db.RoTx) error {
...
+				proof, err := mpt.BuildProof(outTxnTrie.Reader, neighbor.TransactionIndex.Bytes(), mpt.ReadMPTOperation)
+				if err != nil {
+					return nil, err
+				}
+				if err := mpt.PopulateMptWithProof(parents[blockIndex].TxnTrie.MPT(), &proof); err != nil {
+					return nil, err
+				}
```
olegrok, line 413: Do you really need it. Seems it doesn't modify proof itself. Maybe it's reasonable to move it under `assert.Enabled`?

dmtrskv: I store mpts, not proofs. I need proofs only as paths that need to be added to the tree.

nil/internal/mpt/proof_test.go
```diff
@@ -112,6 +120,121 @@ func TestReadProof(t *testing.T) {
...
+			require.NoError(t, err)
+			holder[k] = modified
+			modifiedKey = k
+			break
+		}
+
+		// The holder is not valid anymore, because the hash of the value is not matching the key
+		err := ValidateHolder(holder)
+		require.Error(t, err)
+		require.ErrorContains(t, err, fmt.Sprintf("%x", modifiedKey))
```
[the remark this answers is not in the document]
dmtrskv: No, f in assert functions only affects the error:
`ErrorContainsf(t, err,  expectedErrorSubString, "error message %s", "formatted")`

olegrok (approved):

Issue #54, opened 2025-01-30
Replicate data of all shards to other shards


knazarov:
For quite a while our thinking went like this:

- Every shard will have its own data and compute, which should be enough to process its own blocks
- Shards exchange messages with each other to relay cross-shard transactions (with delivery guarantees)
- The only common part that each shard should have is the "main shard"

This is fine overall, but creates problem with validator rotation and potentially allowing contracts to migrate between shards. There really isn't a simple way to guarantee that one shard's validators will provide data to another shard's validators (there's no way to punish them for not doing so). 

As such, we decided that all shards will have data of all other shards by replaying it in the same way as archive nodes do. And voting for inclusion to the main shard will happen based on the fact that other shards have successfully replicated the data from a particular shard. If that shard refuses to share its data, it won't be voted for.

This seemingly creates a problem of throughput (every shard now has to replicate data from the rest of the network), but this problem can be solved by applying the principle of disaggregated storage if we'll face this problem in the future. But untill we face this problem, we can probably reach at least thousands RPS before we saturate typical network interfaces of a single machine.

So, let's do the right thing architecturally at this point:

- First let's make sure we replicate data of all shards
- Implement voting for the main shard block inclusion based on how far everybody has replicated a particular shard

PR #245, opened 2025-02-10
[syncer] Replicate data from all shards.

Zerg1996:
In this PR, we aim to replicate data across all shards for the validator, even if the validator is not actively validating a specific shard (#54). To ensure this, it’s crucial that the zeroStateConfig is consistent across all validators. To achieve this, we will share the publicMainKey with them.

### DIFF
--- a/nil/internal/collate/syncer.go
+++ b/nil/internal/collate/syncer.go
@@ -2,7 +2,6 @@ package collate
 
 import (
 	"context"
-	"encoding/json"
 	"errors"
 	"fmt"
 	"sync"
@@ -22,12 +21,10 @@ import (
 )
 
 type SyncerConfig struct {
-	Name          string
-	ShardId       types.ShardId
-	Timeout       time.Duration // pull blocks if no new blocks appear in the topic for this duration
-	BootstrapPeer *network.AddrInfo
-	ReplayBlocks  bool // replay blocks (archive node) or store headers and transactions only
-
+	Name                 string
+	ShardId              types.ShardId
+	Timeout              time.Duration // pull blocks if no new blocks appear in the topic for this duration
+	BootstrapPeer        *network.AddrInfo
 	BlockGeneratorParams execution.BlockGeneratorParams
 	ZeroState            string
 	ZeroStateConfig      *execution.ZeroStateConfig
@@ -127,13 +124,11 @@ func (s *Syncer) notify() {
 }
 
 func (s *Syncer) FetchSnapshot(ctx context.Context) error {
-	if s.config.ReplayBlocks {
-		if snapIsRequired, err := s.shardIsEmpty(ctx); err != nil {
-			return err
-		} else if snapIsRequired {
-			if err := fetchSnapshot(ctx, s.networkManager, s.config.BootstrapPeer, s.config.ShardId, s.db); err != nil {
-				return fmt.Errorf("failed to fetch snapshot: %w", err)
-			}
+	if snapIsRequired, err := s.shardIsEmpty(ctx); err != nil {
+		return err
+	} else if snapIsRequired {
+		if err := fetchSnapshot(ctx, s.networkManager, s.config.BootstrapPeer, s.config.ShardId, s.db); err != nil {
+			return fmt.Errorf("failed to fetch snapshot: %w", err)
 		}
 	}
 	return nil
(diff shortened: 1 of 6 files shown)


nil/internal/execution/zerostate.go
```diff
@@ -28,6 +28,36 @@ func init() {
...
+	Validators config.ParamValidators `yaml:"validators,omitempty"`
+	GasPrice   config.ParamGasPrice   `yaml:"gasPrice"`
+}
+
+type ZeroStateConfig struct {
+	ConfigParams ConfigParams     `yaml:"config,omitempty"`
+	Contracts    []*ContractDescr `yaml:"contracts"`
+}
+
+func CreateDefaultZeroState(mainPublicKey []byte) (string, error) {
```
olegrok, line 60:
```suggestion
func CreateDefaultZeroStateConfig(mainPublicKey []byte) (string, error) {
```

nil/services/nilservice/service.go
```diff
@@ -216,6 +216,12 @@ func createArchiveSyncers(cfg *Config, nm *network.Manager, database db.DB, logg
...
 		zeroState := execution.DefaultZeroStateConfig
+		if err := execution.LoadMainKeys(cfg.MainKeysPath); err == nil {
+			zeroState, err = execution.CreateDefaultZeroState(execution.MainPublicKey)
+			if err != nil {
+				return nil, err
+			}
+		}
 		zeroStateConfig := cfg.ZeroState
 		if len(cfg.ZeroStateYaml) != 0 {
 			zeroState = cfg.ZeroStateYaml
```
olegrok, line 227: Let's move this code into a function

nil/internal/collate/syncer.go
```diff
@@ -406,9 +344,11 @@ func (s *Syncer) GenerateZerostate(ctx context.Context) error {
 		return nil
 	}
 
-	if len(s.config.BlockGeneratorParams.MainKeysOutPath) != 0 && s.config.ShardId == types.BaseShardId {
-		if err := execution.DumpMainKeys(s.config.BlockGeneratorParams.MainKeysOutPath); err != nil {
-			return err
+	if len(s.config.BlockGeneratorParams.MainKeysPath) != 0 && s.config.ShardId == types.BaseShardId {
```
olegrok, line 347: Do you plan to update `gen-config` to use common generated zerostate?

https://github.com/NilFoundation/nil/blob/745ae21e083b8e84aae7f7aa8f4a8c349209a2f8/nil/cmd/nild/devnet.go#L126-L130

Zerg1996: Yes, but I don't really want to make it in this PR. Currently - I share file with Private+Public key between all validators. The next step will be to send only public key (it's enough for zerostate) and change gen-config with this approach.

nil/services/nilservice/service.go
```diff
@@ -599,25 +611,32 @@ func createShards(
...
+	var zeroState string
+	if len(cfg.ZeroStateYaml) != 0 {
+		zeroState = cfg.ZeroStateYaml
+	} else {
+		var err error
+		zeroState, err = execution.CreateZeroStateConfigWithMainPublicKey(cfg.MainKeysPath)
+		if err != nil {
+			return nil, err
+		}
+	}
```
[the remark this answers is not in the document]
Zerg1996: oh, yeah I see. thanks for removing unused configs!

nil/internal/collate/syncer.go
olegrok: Ok. Let's do that in the next patch

olegrok (approved):
Looks OK but you  need to fix tests

nil/internal/execution/zerostate.go
```diff
@@ -28,6 +28,39 @@ func init() {
...
+type ZeroStateConfig struct {
+	ConfigParams ConfigParams     `yaml:"config,omitempty"`
+	Contracts    []*ContractDescr `yaml:"contracts"`
+}
+
+func (cfg *ZeroStateConfig) GetValidators() []config.ListValidators {
+	return cfg.ConfigParams.Validators.Validators
+}
+
+func CreateDefaultZeroStateConfig(mainPublicKey []byte) (string, error) {
```
Un1oR, line 63: Is it ok that the parameter is not used inside the function?

PR #248, opened 2025-02-11
Move the fetcher waitgroup into badgerDB

avm:


nil/services/nilservice/service.go
```diff
@@ -464,8 +462,7 @@ func createShards(
 	funcs := make([]concurrent.Func, 0, 2*cfg.NShards)
 	pools := make(map[types.ShardId]txnpool.Pool)
 
-	var wgFetch sync.WaitGroup
-	wgFetch.Add(int(cfg.NShards))
+	database.SetFetcherCount(int(cfg.NShards))
```
olegrok, line 465: Maybe we don't need separate fetchers anymore - https://github.com/NilFoundation/nil/issues/54

And it will be enough to find a node and fetch its db without "syncronization between syncers".

avm: Are we ready to merge a PR that does this? If yes, I will close this one. If not, let's merge this and have less failing tests while we wait for the syncher unification.

nil/internal/db/badger.go
```diff
@@ -103,21 +121,25 @@ func captureStacktrace() []byte {
...
-	return db.createRoTx(ctx, db.db.NewTransactionAt(uint64(ts), false), true)
+	db.lock.RLock()
+	defer db.lock.RUnlock()
+	return db.createRoTx(ctx, db.db.NewTransactionAt(uint64(ts), false), true), nil
 }
 
 func (db *badgerDB) CreateRoTx(ctx context.Context) (RoTx, error) {
-	return db.createRoTx(ctx, db.db.NewTransaction(false), false)
+	db.lock.RLock()
+	defer db.lock.RUnlock()
```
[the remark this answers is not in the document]
avm: There is no contention here, so it's an atomic write+read, unlikely to be a problem until we start doing HFT.

PR #263, opened 2025-02-12
Add a COPYING file with the text of LGPL 3.0.

avm:


0xAleksaOpacic:
If i am not mistaken, by default it should be called LICENSE not COPYING. This way GITHUB recognize it and is able to show license on the side (doc https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/licensing-a-repository#determining-the-location-of-your-license)

avm:
> it should be called LICENSE not COPYING

erigon and go-ethereum use COPYING, github seems to find the license just fine:
https://github.com/ethereum/go-ethereum/blob/master/COPYING
https://github.com/erigontech/erigon/blob/main/COPYING

0xAleksaOpacic (approved):
LGTM! Thank you for adding it

PR #272, opened 2025-02-12
Remove old fee model artifacts

shermike:


nil/tests/config/config_test.go
```diff
@@ -133,28 +133,28 @@ func (s *SuiteConfigParams) TestConfigReadWriteGasPrice() {
...
-
-		data = s.AbiPack(s.abiTest, "testParamGasPriceEqual", gasPrice)
-		receipt = s.SendExternalTransactionNoCheck(data, s.testAddressMain)
-		s.Require().True(receipt.AllSuccess())
-
-		realGasPrice := s.readGasPrices()
-
-		s.Require().Equal(gasPrice.GasPriceScale, realGasPrice.GasPriceScale)
-	})
+	// TODO: uncomment after adding some suitable config param. Cannot use gas price, it is changed by protocol.
```
olegrok, line 143: Maybe let's drop it at all?
If you plan to add/change something it won't be possible to just uncomment it.

I guess it's better to file an issue about it

shermike: I agree. I thought about this. Let's remove it.

olegrok (approved):

PR #273, opened 2025-02-13
Show warning in playground if sa is not connected

ukorvl:
We have a problem in explorer that when user does not have a connected/deployed smart account we don't disable contract management panel and allow to deploy/import/call smart contracts (which evidently fails).

To make it easier to understand for a user what their first action should be, we need to disable contract management panel and show a warning instead, saying that smart account should be connected first.


0xAleksaOpacic (approved):
Looks good, tested! Thank you for opening this PR

PR #274, opened 2025-02-13
[docs] transfered interactive tutorials to explorer_backend

khannanov-nil:
This diff removes the `docs_interactive_tutorials_service` project and transfers its functionality to `explorer_backend`. This is done since it was decided to do the entire interactive tutorials project inside the Playground. Since the backend of the Playground may later be rewritten in Go, it makes sense to make tutorials available inside this Go backend as well.


explorer_backend/services/tutorials_db.ts
```diff
@@ -0,0 +1,60 @@
...
+  id: string;
+  text: string;
+  contracts: string;
+  stage: number;
+}
+
+db.exec(
+  `
+    CREATE TABLE TUTORIALS (
+      id TEXT,
```
KlonD90, line 16: unique should be

explorer_backend/services/tutorials_db.ts
```diff
@@ -0,0 +1,60 @@
...
+export interface Tutorial {
+  id: string;
+  text: string;
+  contracts: string;
+  stage: number;
+}
+
+db.exec(
+  `
+    CREATE TABLE TUTORIALS (
```
KlonD90, line 15: create table if not exists would be better

explorer_backend/services/tutorials_db.ts
KlonD90: better primary

explorer_backend/services/tutorials_db.ts
```diff
@@ -0,0 +1,60 @@
...
+  return { id, text, contracts, stage };
+};
+
+export const generateTutorials = async (): Promise<void> => {
+  const tutorialsSpec = JSON.parse(await fs.readFile('../tutorials/spec.json', 'utf-8'));
+
+  for (const tutorial of tutorialsSpec) {
+    const text = await fs.readFile(`../tutorials/${tutorial.text}`, 'utf-8');
+    const contracts = await fs.readFile(`../tutorials/${tutorial.contracts}`, 'utf-8');
+    const stage = Number(await fs.readFile(`../tutorials/${tutorial.stage}`, 'utf-8'));
```
KlonD90, line 57: Why stage not inside of spec? It's just a number it does have more sense to use it inside spec.

explorer_backend/services/tutorials_db.ts
```diff
@@ -0,0 +1,60 @@
...
+export const setTutorial = (text: string, contracts: string, stage: number): Tutorial => {
+  const existing = getStmt.get(text, contracts, stage);
+  if (existing) {
+    return existing;
+  }
+
+  const updateStmt = db.prepare("UPDATE TUTORIALS SET stage = 0 WHERE stage = ?");
+  updateStmt.run(stage);
+
+  const id = uuidv7();
```
KlonD90, line 45: It's better not generate this id here. Because it would be unstable and you couldn't update proper tutorial. Like If you want to modify or add step or change text.

KlonD90: And insert into on conflict(id) replace could be easier in that case

explorer_backend/routes/tutorials.ts
```diff
@@ -0,0 +1,26 @@
+import { z } from "zod";
+import { router, publicProcedure } from "../trpc";
+import { getTutorial } from "../services/tutorials_db";
+
+export const tutorialRouter = router({
```
KlonD90, line 5: also you need to list of all tutorials

explorer_backend/tutorials/spec.json
```diff
@@ -0,0 +1,5 @@
+[
+  {
```
KlonD90, line 2: it shouild be an empty array not an empty object.

khannanov-nil: Got it, fixed

explorer_backend/services/tutorials_db.ts
khannanov-nil: You are right, fixed

explorer_backend/services/tutorials_db.ts
khannanov-nil: Yes, I agree, fixed

explorer_backend/services/tutorials_db.ts
khannanov-nil: I see, makes sense. Fixed.

explorer_backend/services/tutorials_db.ts
```diff
@@ -0,0 +1,60 @@
...
+  stage: number;
+}
+
+db.exec(
+  `
+    CREATE TABLE TUTORIALS (
+      id TEXT,
+      text TEXT,
+      contracts TEXT,
+      stage INTEGER
```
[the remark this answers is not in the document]
khannanov-nil: Good point, added

explorer_backend/services/tutorials_progress_db.ts
```diff
@@ -0,0 +1,32 @@
+import Database from "better-sqlite3";
+const db = new Database("./database.db");
```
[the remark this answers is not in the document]
khannanov-nil: The namespace is not exportable from sqlite so I cannot expose it from sqlite.ts

package-lock.json
```diff
@@ -197,9 +197,9 @@
       }
     },
     "docs_interactive_tutorials_service/node_modules/@types/node": {
-      "version": "20.17.17",
-      "resolved": "https://registry.npmjs.org/@types/node/-/node-20.17.17.tgz",
```
ukorvl, line 201: Hmm, but why lockfile is changed if there is no change in dependencies?

khannanov-nil:
> it's better to use same instance sqlite. expose from `slite` db instance or move operation above it into this file but it could be done later

thank you for the help! fixed

package-lock.json
khannanov-nil: This only happened because the old `docs_interactive_tutorials_service` was not yet deleted. It is deleted now.

Issue #117, opened 2025-02-03
Explorer navigation


ukorvl:
Currently the **tabs navigation** in explorer is not perfect. The user scenario is:
Open transaction page

![Image](https://github.com/user-attachments/assets/2187dfbe-3b26-47ab-8ede-508bae237d37)

Open related transactions tab and select one of transactions clicking on its **hash**.

![Image](https://github.com/user-attachments/assets/d4a26d09-b662-483c-9bb8-917beb639dc2)

We should show the **overview** of the **selected transaction**, but instead we show the **outgoing transactions tab** of the **selected** **transaction**. it is not correct.

PR #275, opened 2025-02-13
added transaction routing to overview

gitshreevatsa:
This PR solves [this](https://github.com/NilFoundation/nil/issues/117) issue. Transaction routing to overview is fixed. Initially, when you click on a tx hash in the outgoing or incoming message tab, it would route to outgoing or incoming message tab respectively, instead of the overview of the tx hash. But upon this PR it would redirect you to the overview.

### DIFF
--- a/explorer_frontend/src/features/transaction/components/Transaction.tsx
+++ b/explorer_frontend/src/features/transaction/components/Transaction.tsx
@@ -31,6 +31,11 @@ export const Transaction = () => {
     $transactionChilds,
     fetchTransactionChildsFx.pending,
   ]);
+
+  useEffect(() => {
+    if (transaction) setActiveKey("0");
+  }, [transaction]);
+
   const [activeKey, setActiveKey] = useState<Key>("0");
   const onChangeHandler: OnChangeHandler = (currentKey) => {
     setActiveKey(currentKey.activeKey);


explorer_frontend/src/features/contracts/models/base.ts
```diff
@@ -332,3 +333,13 @@ export const removeValueInput = createEvent<number>();
 
 export const $activeComponent = createStore<ActiveComponent | null>(ActiveComponent.Deploy);
 export const setActiveComponent = createEvent<ActiveComponent>();
+
+export const $walletError = $smartAccount.map((wallet) =>
```
ukorvl, line 337: don't use name wallet

explorer_frontend/src/features/contracts/components/Deploy/DeployTab.tsx
```diff
@@ -22,45 +23,75 @@ import {
...
+    pending || !smartAccount || !endpoint || !balance || balance <= 0 || shardId === null;
+
   const [css] = useStyletron();
+  const modalOpened = useRef(false);
+
+  useEffect(() => {
+    modalOpened.current = true;
+  }, []);
+
+  const caption = useMemo(() => {
```
ukorvl, line 53: useMemo is excessive here (it optimizes nothing but takes resources for an optimization function). please use ternary `errorMessage ? ... ?`

ukorvl (approved):

PR #277, opened 2025-02-13
Throw error if in playground import nonexistent smart contract

ukorvl:
Currently we have a severe issue with importing existing contract feature in playground. User can set any address to input, even the address of nonexistent contract (any valid Hex string) and import will not fail. Moreover the fake contract will appear in the project and can cause a lot of misunderstanding.

This is fixed just by checking the code of the contract before importing it. It `getCode()` returns `0x` it means nothing is deploy at address.

Additionally this diff fixes remnants in the code of "assign smart contract" naming.

### DIFF
--- a/explorer_frontend/src/features/contracts/components/Deploy/ImportContractTab.tsx
+++ b/explorer_frontend/src/features/contracts/components/Deploy/ImportContractTab.tsx
@@ -3,56 +3,31 @@ import { Button, COLORS, FormControl, Input, LabelSmall, SPACE } from "@nilfound
 import { INPUT_KIND, INPUT_SIZE } from "@nilfoundation/ui-kit";
 import type { InputOverrides } from "baseui/input";
 import { useUnit } from "effector-react";
-import { useCallback, useEffect, useState } from "react";
 import { useStyletron } from "styletron-react";
 import { $smartAccount } from "../../../account-connector/model";
 import {
   $activeAppWithState,
-  $deployedContracts,
   $importedSmartContractAddress,
+  $importedSmartContractAddressError,
+  $importedSmartContractAddressIsValid,
   importSmartContract,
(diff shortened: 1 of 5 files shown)


explorer_frontend/src/features/contracts/components/Deploy/ImportContractTab.tsx
```diff
@@ -94,7 +96,7 @@ export const ImportContractTab = () => {
       <div>
         <Button
           onClick={() => {
-            if (!error) assignSmartContract();
+            if (!error) importSmartContract();
```
0xAleksaOpacic, line 99: Why don't we trigger validate on error, catching in UseEffect can create quite a fee side effects. Personally i don't like errors while i am typing, but when i press validate input if it's not valid show me error if it's valid execute action


https://github.com/user-attachments/assets/121c1369-c714-4e58-82e5-f3d6cfb4bc94

0xAleksaOpacic (changes requested):
Validation is not triggered on button press, i think it's one liner, but let's make sure it's fixed before merging

explorer_frontend/src/features/contracts/components/Deploy/ImportContractTab.tsx
ukorvl: Yes, that's correct 👍 . There were plenty of issues with this feture.

ukorvl:
@0xAleksaOpacic check plz how it's working now. i added ux-friendly errors and changed the mechanics of how a value of address is validated.[

https://github.com/user-attachments/assets/2b0dc36c-9881-4e3e-8631-f62f282176a9

](url)

0xAleksaOpacic (approved):
Thank you for fixing everything! It works! Tested

PR #283, opened 2025-02-13
[consensus] introduce bls

isergeyam:
This patch introduces BLS signature scheme and it's implementation using kyber. The implementation is hiddent behind interfaces, so we will be able to replace it when the time comes.

### DIFF
--- a/nil/internal/collate/syncer.go
+++ b/nil/internal/collate/syncer.go
@@ -434,7 +434,7 @@ func (s *Syncer) replayBlock(ctx context.Context, block *types.BlockWithExtracte
 	}
 
 	// Finally, write generated block into the database
-	if err = gen.Finalize(resBlock, block.Signature); err != nil {
+	if err = gen.Finalize(resBlock, &block.Signature); err != nil {
 		return fmt.Errorf("failed to finalize block: %w", err)
 	}
 
--- a/nil/internal/collate/validator.go
+++ b/nil/internal/collate/validator.go
@@ -45,7 +45,7 @@ func (s *Validator) VerifyProposal(ctx context.Context, proposal *execution.Prop
 	return res.Block, nil
 }
 
-func (s *Validator) InsertProposal(ctx context.Context, proposal *execution.Proposal, sig types.Signature) error {
+func (s *Validator) InsertProposal(ctx context.Context, proposal *execution.Proposal, sig *types.BlsAggregateSignature) error {
 	gen, err := execution.NewBlockGenerator(ctx, s.params.BlockGeneratorParams, s.txFabric,
 		&proposal.PrevBlockHash, proposal.GetMainShardHash(s.params.ShardId))
 	if err != nil {
--- a/nil/internal/config/params.go
+++ b/nil/internal/config/params.go
@@ -8,12 +8,13 @@ import (
 	"github.com/NilFoundation/nil/nil/common/check"
 	"github.com/NilFoundation/nil/nil/common/hexutil"
 	"github.com/NilFoundation/nil/nil/common/logging"
+	"github.com/NilFoundation/nil/nil/internal/crypto/bls"
 	"github.com/NilFoundation/nil/nil/internal/db"
 	"github.com/NilFoundation/nil/nil/internal/types"
 	"github.com/rs/zerolog"
 )
 
-const ValidatorPubkeySize = 33
+const ValidatorPubkeySize = 128
 
 const (
 	NameValidators = "curr_validators"
@@ -64,7 +65,7 @@ type ParamValidators struct {
 }
 
 type ValidatorInfo struct {
-	PublicKey         Pubkey        `json:"pubKey" yaml:"pubKey" ssz-size:"33"`
+	PublicKey         Pubkey        `json:"pubKey" yaml:"pubKey" ssz-size:"128"`
 	WithdrawalAddress types.Address `json:"withdrawalAddress" yaml:"withdrawalAddress"`
 }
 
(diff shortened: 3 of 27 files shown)


nil/internal/crypto/bls/kyber/kyber_test.go
```diff
@@ -0,0 +1,74 @@
...
+
+	err = aggregatedSig.Verify(aggregatedKey, msg)
+	require.NoError(t, err)
+
+	mask.SetParticipants([]uint32{0})
+	aggregatedKey, err = mask.AggregatePublicKeys()
+	require.NoError(t, err)
+
+	err = aggregatedSig.Verify(aggregatedKey, msg)
+	require.Error(t, err)
```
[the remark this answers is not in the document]
isergeyam: Done

nil/internal/crypto/bls/kyber/init.go
```diff
@@ -0,0 +1,11 @@
+package kyber
+
+import (
+	"go.dedis.ch/kyber/v3/pairing/bn256"
```
olegrok, line 4: https://github.com/dedis/kyber/tree/fe452f240eed5718b4e9813ad74446d5aa6d1ee3?tab=readme-ov-file#using-the-module

> Kyber supports Go modules, and currently has a major version of 3, which means that the import path is: go.dedis.ch/kyber/v4.

Why do you use v3?

nil/internal/collate/validator.go
```diff
@@ -58,6 +59,11 @@ func (s *Validator) InsertProposal(ctx context.Context, proposal *execution.Prop
 		return fmt.Errorf("failed to generate block: %w", err)
 	}
 
+	bv := signer.NewBlockVerifier(s.params.ShardId, s.txFabric)
+	if err := bv.VerifyBlock(ctx, res.Block, s.logger); err != nil {
+		return fmt.Errorf("failed to verify block: %w", err)
+	}
```
olegrok, line 65: Shouldn't it be done in IsValidCommittedSeal?

nil/internal/crypto/bls/kyber/init.go
isergeyam: There is a mess in the versioning in this library, but the most recent stable release is v3.1.0, so I use it

nil/internal/collate/syncer.go
```diff
@@ -321,7 +321,7 @@ func (s *Syncer) saveBlock(ctx context.Context, block *types.BlockWithExtractedD
 			Uint64(logging.FieldBlockNumber, uint64(block.Id)).
 			Stringer(logging.FieldBlockHash, block.Hash(s.config.ShardId)).
 			Stringer(logging.FieldShardId, s.config.ShardId).
-			Stringer(logging.FieldSignature, block.Signature).
+			Stringer(logging.FieldSignature, &block.Signature).
```
olegrok, line 324: Do we really need it? If `*Signature` is a Stringer then `Signature` should be also Stringer

nil/internal/config/params.go
```diff
@@ -202,6 +203,21 @@ func GetValidatorListForShard(
...
 }
 
+func GetValidatorsPubKeys(validators []ValidatorInfo, logger zerolog.Logger) ([]bls.PublicKey, error) {
+	pubkeys := make([]bls.PublicKey, 0, len(validators))
+	for _, v := range validators {
+		pk, err := bls.PublicKeyFromBytes(v.PublicKey[:])
+		if err != nil {
+			logger.Error().Err(err).
+				Hex(logging.FieldPublicKey, v.PublicKey[:]).
+				Msg("Failed to read public key")
```
olegrok, line 213: Let's avoid logging in this "too deep" place.

nil/internal/consensus/ibft/ibft.go
```diff
@@ -79,27 +83,92 @@ func (i *backendIBFT) BuildProposal(view *protoIBFT.View) []byte {
...
+		}
+		index := slices.IndexFunc(validators, func(v config.ValidatorInfo) bool {
+			return bytes.Equal(v.PublicKey[:], seal.Signer)
+		})
+		if index == -1 {
+			logger.Error().
+				Hex(logging.FieldPublicKey, seal.Signer).
+				Msg("Signer not found in validators list")
+			return nil, err
+		}
```
olegrok, line 122: Is it possible to create `map[publicKey]...`? It should be a bit more useful for such cases. And also we can use it in `IsValidValidator`

nil/internal/consensus/ibft/signer.go
```diff
@@ -1,43 +1,65 @@
...
+func (s *Signer) SignHash(hash []byte) ([]byte, error) {
+	sig, err := s.privateKey.Sign(hash)
+	if err != nil {
+		return nil, err
+	}
+	return sig.Marshal()
 }
 
-func (s *Signer) Sign(data []byte) (types.Signature, error) {
+func (s *Signer) Sign(data []byte) ([]byte, error) {
```
olegrok, line 37: Is it possible to introduce `types.BlsSignature` type?

nil/internal/consensus/ibft/ibft.go
isergeyam: Introduced PublicKeyMap

nil/internal/consensus/ibft/verifier.go
```diff
@@ -37,29 +36,31 @@ func (i *backendIBFT) IsValidValidator(msg *protoIBFT.IbftMessage) bool {
...
-		event := i.logger.Error()
-		if view := msg.GetView(); view != nil {
-			event = event.Uint64(logging.FieldHeight, view.Height).
-				Uint64(logging.FieldRound, view.Round)
-		}
-		event.Msg("Invalid signature")
+	_, ok := pubkeys.Find(config.Pubkey(msg.From))
+	if !ok {
+		logger.Error().
+			Msg("public key not found in validators list")
```
olegrok, line 58: `return false` is missed

nil/internal/types/signature.go
```diff
@@ -33,3 +34,16 @@ func (s Signature) Hex() string {
 func (s Signature) String() string {
 	return s.Hex()
 }
+
+type BlsSignature []byte
+
+type BlsAggregateSignature struct {
+	Sig  []byte `json:"sig" yaml:"sig" ssz-max:"64"`
+	Mask []byte `json:"mask" yaml:"mask" ssz-max:"128"`
```
olegrok, line 42: By default `json` serializes it like an array `[1 2 3]`. Please sure it's so and implement MarshalText, UnmarshalText methods... Or you can use hexutil.Bytes instead of []byte

nil/internal/types/signature.go
```diff
@@ -33,3 +34,16 @@ func (s Signature) Hex() string {
...
+type BlsAggregateSignature struct {
+	Sig  hexutil.Bytes `json:"sig" yaml:"sig" ssz-max:"64"`
+	Mask hexutil.Bytes `json:"mask" yaml:"mask" ssz-max:"128"`
+}
+
+func (b BlsAggregateSignature) String() string {
+	return fmt.Sprintf("BlsAggregateSignature{Sig: %x, Mask: %x}", b.Sig, b.Mask)
+}
+
+//go:generate go run github.com/NilFoundation/fastssz/sszgen --path signature.go -include ../../common/hexutil/bytes.go --objs BlsAggregateSignature
```
olegrok, line 49: Do you really need it?

nil/internal/config/params.go
```diff
@@ -202,6 +203,49 @@ func GetValidatorListForShard(
...
+	m.m[key] = index
+	pk, err := bls.PublicKeyFromBytes(key[:])
+	if err != nil {
+		return err
+	}
+	m.keys = append(m.keys, pk)
+	return nil
+}
+
+func GetValidatorsPubKeys(validators []ValidatorInfo) (*PublicKeyMap, error) {
```
olegrok, line 239:
```suggestion
func CreateValidatorsPublicKeyMap(validators []ValidatorInfo) (*PublicKeyMap, error) {
```

olegrok: "Get" is usually about some read-only operation

nil/internal/crypto/bls/kyber/init.go
```diff
@@ -0,0 +1,11 @@
...
+)
+
+var suite *bn256.Suite
+
+func init() {
+	suite = bn256.NewSuite()
+}
```
olegrok, line 11: Why not?

```suggestion
var suite *bn256.Suite = bn256.NewSuite()
```

olegrok (approved):
I left one comment

PR #285, opened 2025-02-13
[collate] drop unused config options

olegrok:
Right now syncer is responsible for zerostate generation.


avm (approved):

Zerg1996 (approved):
Thanks a lot for update params!

PR #287, opened 2025-02-14
[execution] drop unused timer field from config

olegrok:
Follow-up 44b36b686fc853d58c669c7c9a901d8031246626


Zerg1996 (approved):
Looks like we didn't use it. So yeah, thanks for PR and let's remove it.

Issue #296, opened 2025-02-14
Mapping fields not intercepted during a view call


gitshreevatsa:
Currently in the playground, when you deploy a contract that has mapping in its code and query the mapping it results in an error as:
```
Call failed

InvalidAddressError: Address "" is invalid.

- Address must be a hex value of 20 bytes (40 hex characters).
- Address must match its checksum counterpart.

Version: viem@2.22.17
```
It mentions as address as the mapping was `mapping(address => uint256)` but if you use uint256 as key then every query would be 0. Ideally when queried with a eligible key it has to return the values similar to a view function

PR #297, opened 2025-02-14
Added support to interact with a mapping via playground

gitshreevatsa:
This PR solves this [issue](https://github.com/NilFoundation/nil/issues/296) where you would get an error or a wrong default result when you query a mapping. This PR would enable this functionality as initially the `input.name` from the `abi` was checked but mapping usually have no name and hence this would not take any args. This PR will add a check to see if the `input.name` is empty string and if so it would take into consideration that it is a mapping and hence extract the arguments in a different manner.

### DIFF
--- a/explorer_frontend/src/features/contracts/init.ts
+++ b/explorer_frontend/src/features/contracts/init.ts
@@ -341,8 +341,14 @@ sample({
           if (typeof input.name !== "string") {
             continue;
           }
-          const name = input.name;
-          args.push(callParams[name] || "");
+          if (input.name === "") {
+            for (const key in callParams) {
+              args.push(callParams[key] || "");
+            }
+          } else {
+            const name = input.name;
+            args.push(callParams[name] || "");
+          }
         }
       }
     }


KlonD90:
Does should be done with updating `Method.tsx`? If I've got this correct. Could you give an ABI example to test it out?

gitshreevatsa:
You can use this contract on the playground to test it out completely, where you call function `deposit` with some value and then in function `balances` you enter the value of `i` which would be `1` and you will see `0` instead of deposited value in wei. 

```solidity
contract Counter is NilBase {
  
    uint256 private value; // Stores the current counter value
    uint256 public i;
    mapping (uint256 => uint256) public balances;
  
    // Increments the counter by 1 and emits a ValueChanged event
    function increment() public {
        value += 1;
    }

    // Returns the current value of the counter
    function getValue() public view returns (uint256) {
        return value;
    }

    function deposit() payable public {
      i++;
      balances[i] += msg.value;
    }

 }
 
```

The ABI:

```json
[
	{
		"inputs": [
			{
				"internalType": "uint256",
				"name": "",
				"type": "uint256"
			}
		],
		"name": "balances",
		"outputs": [
			{
				"internalType": "uint256",
				"name": "",
				"type": "uint256"
			}
		],
		"stateMutability": "view",
		"type": "function"
	},
	{
		"inputs": [],
		"name": "deposit",
		"outputs": [],
		"stateMutability": "payable",
		"type": "function"
	},
	{
		"inputs": [],
		"name": "getValue",
		"outputs": [
			{
				"internalType": "uint256",
				"name": "",
				"type": "uint256"
			}
		],
		"stateMutability": "view",
		"type": "function"
	},
	{
		"inputs": [],
		"name": "i",
		"outputs": [
			{
				"internalType": "uint256",
				"name": "",
				"type": "uint256"
			}
		],
		"stateMutability": "view",
		"type": "function"
	},
	{
		"inputs": [],
		"name": "increment",
		"outputs": [],
		"stateMutability": "nonpayable",
		"type": "function"
	}
]

```

If you see here, `balances` is a type mapping. 
```
"inputs": [
			{
				"internalType": "uint256",
				"name": "",
				"type": "uint256"
			}
		],
		"name": "balances",
		"outputs": [
			{
				"internalType": "uint256",
				"name": "",
				"type": "uint256"
			}
		],
		"stateMutability": "view",
		"type": "function"
	},
```
And has no value to key `name`

Rolaman (approved):

PR #302, opened 2025-02-15
[rpc] log client RPC events

Gezort:
this PR adds some logging based on RPC request headers. mostly for debugging and understanding whether client functionality works properly as we release new versions of our CLI / wallet / nil.js

also it reworks a bit our mechanism for event writing to ClickHouse. since we reused `zerolog` interface here (which is rly convenient), there's one small caveat: we need to control global logging level. otherwise event logging may be occasionally disabled

### DIFF
--- a/nil/common/logging/fields.go
+++ b/nil/common/logging/fields.go
@@ -51,4 +51,8 @@ const (
 	FieldHeight    = "height"
 	FieldRound     = "round"
 	FieldType      = "type"
+
+	FieldClientType    = "clientType"
+	FieldClientVersion = "clientVersion"
+	FieldUid           = "uid"
 )
--- a/nil/internal/execution/block_generator.go
+++ b/nil/internal/execution/block_generator.go
@@ -129,7 +129,7 @@ func (g *BlockGenerator) collectGasPrices(prevBlockHash common.Hash, shards []co
 
 			block, err := db.ReadBlock(roTx, shardId, shardHash)
 			if err != nil {
-				logger.Err(err).
+				g.logger.Err(err).
 					Stringer(logging.FieldShardId, shardId).
 					Msg("Get gas price from shard: failed to read last block")
 				gasPrice.Shards[shardId] = *types.DefaultGasPrice.Uint256
@@ -174,21 +174,21 @@ func (g *BlockGenerator) GenerateZeroState(zeroStateYaml string, config *ZeroSta
 	return res.Block, nil
 }
 
-func (g *BlockGenerator) prepareExecutionState(proposal *Proposal, logger zerolog.Logger) error {
+func (g *BlockGenerator) prepareExecutionState(proposal *Proposal) error {
 	if g.executionState.PrevBlock != proposal.PrevBlockHash {
 		esJson, err := g.executionState.MarshalJSON()
 		if err != nil {
-			logger.Err(err).Msg("Failed to marshal execution state")
+			g.logger.Err(err).Msg("Failed to marshal execution state")
 			esJson = nil
 		}
 		//nolint:musttag
 		proposalJson, err := json.Marshal(proposal)
 		if err != nil {
-			logger.Err(err).Msg("Failed to marshal block proposal")
+			g.logger.Err(err).Msg("Failed to marshal block proposal")
 			proposalJson = nil
 		}
 
-		logger.Debug().
+		g.logger.Debug().
 			Stringer("expected", g.executionState.PrevBlock).
 			Stringer("got", proposal.PrevBlockHash).
 			RawJSON("executionState", esJson).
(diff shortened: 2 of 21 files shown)


nil/common/logging/logger.go
```diff
@@ -139,6 +139,14 @@ func NewLogger(component string) zerolog.Logger {
 		Logger()
 }
 
+func NewMetricsLogger(component, table string) zerolog.Logger {
+	return NewLogger(component).With().
+		Bool("store_to_clickhouse", true).
+		Str("database", "metrics").
+		Str("table", table).
```
olegrok, line 146: I'm not sure you solve your problem properly.
Don't you need separate type of logger with custom exporter directly to clickhouse?

nil/internal/execution/block_generator.go
```diff
@@ -174,21 +174,21 @@ func (g *BlockGenerator) GenerateZeroState(zeroStateYaml string, config *ZeroSta
 	return res.Block, nil
 }
 
-func (g *BlockGenerator) prepareExecutionState(proposal *Proposal, logger zerolog.Logger) error {
+func (g *BlockGenerator) prepareExecutionState(proposal *Proposal) error {
 	if g.executionState.PrevBlock != proposal.PrevBlockHash {
 		esJson, err := g.executionState.MarshalJSON()
 		if err != nil {
-			logger.Err(err).Msg("Failed to marshal execution state")
+			g.logger.Err(err).Msg("Failed to marshal execution state")
```
olegrok, line 181: Such changes are really good and I propose to move them into separate patch to merge out of scope of this review.

nil/services/rpc/jsonrpc/send_transaction.go
```diff
@@ -19,12 +22,33 @@ func (api *APIImpl) SendRawTransaction(ctx context.Context, encoded hexutil.Byte
...
+	// occasionally gone when global log level is increased
+	log := api.clientEventsLog.Warn().
+		Stringer(logging.FieldShardId, shardId).
+		Str(logging.FieldRpcMethod, "eth_sendRawTransaction").
+		Str(logging.FieldClientType, headers.Get("Client-Type")).
+		Str(logging.FieldClientVersion, headers.Get("Client-Version")).
+		Str(logging.FieldUid, headers.Get("X-UID")) //nolint:canonicalheader
+
 	if err != nil {
+		log.Msg("finished with err")
```
olegrok, line 42: Usually we add error itself
```suggestion
		log.Err(err).Msg("finished with err")
```

nil/common/logging/logger.go
Gezort: wdym? exactly this way of logging to CH was proposed by @Zerg1996 in another review earlier

nil/services/rpc/jsonrpc/send_transaction.go
```diff
@@ -19,12 +22,33 @@ func (api *APIImpl) SendRawTransaction(ctx context.Context, encoded hexutil.Byte
...
+
+	headers, ok := ctx.Value(transport.HeadersContextKey).(http.Header)
+	if !ok {
+		headers = http.Header{}
+		headers.Add("Client-Type", "failed to extract headers from context")
+	}
+
+	// even though warn here looks weird these logs shouldn't be
+	// occasionally gone when global log level is increased
+	log := api.clientEventsLog.Warn().
```
olegrok, line 34: Why `warn`?

nil/internal/execution/block_generator.go
Gezort: well, it's just a couple of lines and more about codestyle. I propose to keep it here to avoid unnecessary long CI checks :(

nil/common/logging/logger.go
olegrok: Oh... I see...

https://github.com/NilFoundation/nil/blob/f23f70942199fc7d64b72aba79a8386bdd0aad97/nil/services/journald_forwarder/main.go#L222

nil/services/rpc/jsonrpc/send_transaction.go
olegrok: Trace/Debug/Info

Gezort: I left the comment describing why it's preferable to keep warn level here

Gezort:
wallet tests are arleady fixed in main. just don't want to rebase once again without strict necessity

Zerg1996 (approved):

nil/common/logging/fields.go
```diff
@@ -51,4 +51,8 @@ const (
 	FieldHeight    = "height"
 	FieldRound     = "round"
 	FieldType      = "type"
+
+	FieldClientType    = "clientType"
+	FieldClientVersion = "clientType"
```
olegrok, line 56:
```suggestion
	FieldClientVersion = "clientVersion"
```

PR #303, opened 2025-02-16
[sync committee] Enable skip mode

akokoshn:
**Add mode `skip-proof` for prover:**  run without actual proof. It useful when need to check whole pipeline without powerful prover network.

**Allow start sync committee service without L1 contract access** for check main functionality even no access to L1 contract.

### DIFF
--- a/nil/cmd/proof_provider/main.go
+++ b/nil/cmd/proof_provider/main.go
@@ -54,6 +54,7 @@ func addFlags(cmd *cobra.Command, cfg *cmdConfig) {
 	cmd.Flags().StringVar(&cfg.TaskListenerRpcEndpoint, "own-endpoint", cfg.TaskListenerRpcEndpoint, "own rpc server endpoint")
 	cmd.Flags().StringVar(&cfg.DbPath, "db-path", "proof_provider.db", "path to database")
 	cmd.Flags().BoolVar(&cfg.Telemetry.ExportMetrics, "metrics", cfg.Telemetry.ExportMetrics, "export metrics via grpc")
+	cmd.Flags().IntVar(&cfg.SkipRate, "skip", cfg.SkipRate, "rate of skip tasks, will skip N from 10, where N is value of option (0 means no skip). Possible values: [0,10]")
 	logLevel := cmd.Flags().String("log-level", "info", "log level: trace|debug|info|warn|error|fatal|panic")
 
 	cmd.PreRun = func(cmd *cobra.Command, args []string) {
--- a/nil/services/synccommittee/proofprovider/task_handler.go
+++ b/nil/services/synccommittee/proofprovider/task_handler.go
@@ -12,24 +12,49 @@ import (
 )
 
 type taskHandler struct {
-	taskStorage storage.TaskStorage
-	timer       common.Timer
-	logger      zerolog.Logger
+	taskStorage   storage.TaskStorage
+	resultStorage storage.TaskResultStorage
+	skipRate      int
+	taskNum       int
+	timer         common.Timer
+	logger        zerolog.Logger
 }
 
 func newTaskHandler(
-	taskStorage storage.TaskStorage, timer common.Timer, logger zerolog.Logger,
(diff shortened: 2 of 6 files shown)


nil/services/synccommittee/prover/commands/commands.go
```diff
@@ -32,12 +33,17 @@ type AfterCommandExecuted interface {
...
+	binary      string
+	outDir      string
+	isSkipProof bool
 }
 
 func makeCmdCommon(config CommandConfig) cmdCommon {
-	return cmdCommon{binary: config.ProofProducerBinary, outDir: config.OutDir}
+	binary := config.ProofProducerBinary
+	if config.IsSkipProof {
+		binary = "echo"
```
makxenov, line 44: What do you think about simple return of success result before executing `taskHandler.HandleImpl()` in case of `isSkipProof`? In this case we won't have to bear in mind particular commands that has side effects. I don't think that printing command arguments adds much in testing of no-proof pipeline, but we definitely lose in code quality here.

akokoshn: 1. Debug purpose: i want to be able check if command line is correct
2. Still need to generate binary output for `merge proofs`

makxenov: 1. You can run specific task with `--skip-proof=false`. It is more reliable then retrieving arguments from stdout.
2. The output is trivial, so it can be generated outside the scope of `merge proofs` command, just before returning of the result.

nil/services/synccommittee/proofprovider/task_handler.go
```diff
@@ -12,26 +12,51 @@ import (
...
 }
 
-func (h *taskHandler) Handle(ctx context.Context, _ types.TaskExecutorId, task *types.Task) error {
+func (h *taskHandler) Handle(ctx context.Context, executorId types.TaskExecutorId, task *types.Task) error {
 	if (task.TaskType != types.ProofBlock) && (task.TaskType != types.AggregateProofs) {
 		return types.NewTaskErrNotSupportedType(task.TaskType)
 	}
 
+	// skip task
+	taskIdx := h.taskNum%10
```
makxenov, line 42: It seems that `h.taskNum` is always zero for now, so is `taskIdx`

nil/cmd/proof_provider/main.go
```diff
@@ -54,6 +54,7 @@ func addFlags(cmd *cobra.Command, cfg *cmdConfig) {
 	cmd.Flags().StringVar(&cfg.TaskListenerRpcEndpoint, "own-endpoint", cfg.TaskListenerRpcEndpoint, "own rpc server endpoint")
 	cmd.Flags().StringVar(&cfg.DbPath, "db-path", "proof_provider.db", "path to database")
 	cmd.Flags().BoolVar(&cfg.Telemetry.ExportMetrics, "metrics", cfg.Telemetry.ExportMetrics, "export metrics via grpc")
+	cmd.Flags().IntVar(&cfg.SkipRate, "skip", cfg.SkipRate, "rate of skip tasks, will skip value from 10, 0 means no skip. Possible values: [0,10]")
```
makxenov, line 57: It makes sense to specify that we skip only proof-provider tasks.

makxenov, line 57: > will skip value from 10

meaning of `value`  is not clear here

akokoshn: proof-provider handle only proof-provider tasks (proof batch), we can't to skip something also

nil/services/synccommittee/proofprovider/task_handler.go
akokoshn: thanks!, i miss increment

nil/cmd/proof_provider/main.go
makxenov: Sorry, can see now that this option is exclusively for proof-provider. Please correct the PR description also, there is still prover in it.

PR #305, opened 2025-02-17
Add playground rpc api healthcheck

ukorvl:
When rpc api is inaccessible or user has an invalid rpc url, we want to let them know something is wrong.
We are going to display a warning if api is inaccessible and disable some UI features as well.

RPC check is made every 10 seconds or on any endpoint change. If user is not on the page the check is not made for the sake of optimizing resources.

### DIFF
--- a/explorer_frontend/src/features/healthcheck/init.ts
+++ b/explorer_frontend/src/features/healthcheck/init.ts
@@ -1,0 +1,37 @@
+import { combine, sample } from "effector";
+import { interval } from "patronum";
+import { $endpoint } from "../account-connector/model";
+import { loadedPlaygroundPage } from "../code/model";
+import { playgroundRoute, playgroundWithHashRoute } from "../routing/routes/playgroundRoute";
+import { $isPageVisible, $rpcIsHealthy, checkRpcHealthFx, pageVisibilityChanged } from "./model";
+
+$isPageVisible.on(pageVisibilityChanged, (_, isVisible) => isVisible);
+document.addEventListener("visibilitychange", () => {
+  pageVisibilityChanged(document.visibilityState === "visible");
+});
+
+const { tick } = interval({
+  timeout: 1000 * 10,
+  start: loadedPlaygroundPage,
+  leading: true,
+});
+
+sample({
+  clock: tick,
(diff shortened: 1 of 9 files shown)


explorer_frontend/src/features/healthcheck/init.ts
```diff
@@ -0,0 +1,30 @@
+import { sample } from "effector";
+import { interval } from "patronum";
+import { $endpoint } from "../account-connector/model";
+import { loadedPlaygroundPage } from "../code/model";
+import { $isPageVisible, $rpcIsHealthy, checkRpcHealthFx, pageVisibilityChanged } from "./model";
+
+$isPageVisible.on(pageVisibilityChanged, (_, isVisible) => isVisible);
+document.addEventListener("visibilitychange", () => {
```
KlonD90, line 8: it's better to be in react useEffect because that's ok for production but will be a million of event listener in case of dev mode

explorer_frontend/src/features/healthcheck/init.ts
```diff
@@ -0,0 +1,30 @@
...
+  timeout: 1000 * 10,
+  start: loadedPlaygroundPage,
+  leading: true,
+});
+
+sample({
+  clock: tick,
+  target: checkRpcHealthFx,
+  source: $endpoint,
+  filter: $isPageVisible,
```
KlonD90, line 22: also would be good to have isPageVisible and that is actually playground or related page. you could check this with $opened on route instance

KlonD90 (approved):

explorer_frontend/src/features/healthcheck/init.ts
KlonD90: Do you try to update this file in this case? It's usually do not delete previous listener

explorer_frontend/src/features/healthcheck/init.ts
ukorvl: Done

explorer_frontend/src/features/healthcheck/init.ts
ukorvl: Yep, checked again with this file and with changes in UI - no listeners overflow

PR #311, opened 2025-02-17
[syncer] drop per-shard snapshot bootstrap

olegrok:
Anyway we decided that all validator/archive_nodes replicate all data of all shards. So no need to support per-shard snapshotting.


Zerg1996 (approved):
Thanks for PR!

PR #312, opened 2025-02-17
Set initial random shard-id in playground

ukorvl:
We had a discussion about shard id and decided to change a little the mechanics of shard id.
What we want:
- When user opens modal initially and there is no previously used shard id, we need to set as a default value any random number in the range from `0` to `shards amount`.
- When user opens modal second time, we need to preserve previously used shard id, because it is more probable theat user is going to use it again.
- Important notes - always trigger validation on default value

### DIFF
--- a/explorer_frontend/src/features/contracts/init.ts
+++ b/explorer_frontend/src/features/contracts/init.ts
@@ -55,9 +55,11 @@ import {
   setImportedSmartContractAddressError,
   setImportedSmartContractAddressIsValid,
   setParams,
+  setRandomShardId,
   setShardId,
   setValueInput,
   toggleActiveKey,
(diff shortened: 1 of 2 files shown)


0xAleksaOpacic (approved):
Overall it works, thanks for adding this! I am just thinking do we want to re-trigger on every deploy button. Currently we generate random shard number on page load

ukorvl:
> Overall it works, thanks for adding this! I am just thinking do we want to re-trigger on every deploy button. Currently we generate random shard number on page load

We don't usually have an info about the amount of shards on page load.
For me re-triggering every time user press a deploy button is ok, especially because we have a filter to generate random shard id only if it is null.

Issue #304, opened 2025-02-16
Playground wallet focused input UI issue


ukorvl:
![Image](https://github.com/user-attachments/assets/7e2bc13b-7db2-4480-a47a-d1eddd6b3a20)

As you can see on the screenshot, input outline display is incorrect in focused state and partially hidden. Typically it happens when a container covers its content and the easiest solution is to add some padding to the content so it is not hidden by container (remove the same value from the container to keep dimensions).

ukorvl:
The other important problem is that a user can not type anything to the input.

ukorvl:
It appeared, this was intended to make an input non-interactive.
So, let's make it non focusable as well :D

New DOD: this input does not have focus styles. When user focuses input - nothing changes.

ukorvl:
Example 

https://github.com/user-attachments/assets/ad0076e7-ed2a-4f8b-9c0b-22d7ab1f81fa

PR #313, opened 2025-02-17
Fix issue with playground RPC input hidden border

gitshreevatsa:
This PR solves the bug mentioned [here](https://github.com/NilFoundation/nil/issues/304). Currently, the RPC input, after connecting the wallet is a readOnly input and would confuse users form trying to edit as it appears as a normal input box where you can click on it and the borders get highlighted, just as every input box getting active. This PR would make the RPC input box inactive and not allow for focussing so that it behaves as a non-interactive UI component, removing a confusion for the interaction with the input box.

### DIFF
--- a/explorer_frontend/src/features/account-connector/components/MainScreen.tsx
+++ b/explorer_frontend/src/features/account-connector/components/MainScreen.tsx
@@ -156,9 +156,10 @@ const MainScreen = () => {
               overrides={{
                 Root: {
                   style: {
-                    flex: 1, // Makes input take full available width
+                    flex: 1,
                     height: "48px",
                     background: COLORS.gray700,
+                    boxShadow: "none",
                     ":hover": {
                       background: COLORS.gray600,
                     },
@@ -166,7 +167,6 @@ const MainScreen = () => {
                 },
               }}
             />
-
             {/* Copy Button */}
             <Button
               kind={BUTTON_KIND.secondary}


explorer_frontend/src/features/account-connector/components/MainScreen.tsx
```diff
@@ -156,9 +159,13 @@ const MainScreen = () => {
               overrides={{
                 Root: {
                   style: {
-                    flex: 1, // Makes input take full available width
+                    flex: 1,
                     height: "48px",
                     background: COLORS.gray700,
+                    paddingLeft: "12px", // Add left padding
+                    paddingRight: "12px", // Add right padding
+                    maxWidth: "80%", // Ensure it doesn't stretch too far
```
ukorvl, line 167: remove all comments plz

ukorvl: why don't we need comments here. First, most of them have no semantic load, like `"add right padding"`.
Second, to write a comments in css (even css in js) is a bad idea, because typically css itself is very declarative, properties names are meaningful and easy to understand. If a person knows css they need no comments to read it.
And, eventually, from my POV - code that needs lots of comments is not very good, because a good code is documenting itself.

ukorvl: We can only leave a comment in css if the property value is strange/extraordinary, the solution does not look clear and we wanna explain the reason for the others.

explorer_frontend/src/features/account-connector/components/MainScreen.tsx
```diff
@@ -156,9 +159,13 @@ const MainScreen = () => {
...
+                    borderRadius: "8px", // Soften the edges for better UI
```
ukorvl, line 168: No, this is a standard input from ui-kit and we should not modify things like border radius. It's unified across all the app and the website.

ukorvl (changes requested):
image

 We can not do it like this. The elements should be aligned agains its container.
 Your main source of the truth should be figma template. Plz check wallet templates and see that this elements are always aligned. If we would like to introduce and changes or ideas - we need to contact our designer first.

knazarov:
When making fixes, please write enough details in the description. It's not enough to point to a ticket: the description should be self-contained. The title should also tell which bug you're fixing.

explorer_frontend/src/features/account-connector/init.ts
```diff
@@ -160,7 +160,7 @@ topUpSmartAccountBalanceFx.use(async (smartAccount) => {
 
   await faucetClient.topUpAndWaitUntilCompletion(
     {
-      smartAccount: smartAccount.address,
+      smartAccountAddress: smartAccount.address,
```
ukorvl, line 163: We already have this change in the code, probably you need to rebase

explorer_frontend/src/features/account-connector/components/MainScreen.tsx
```diff
@@ -156,17 +156,18 @@ const MainScreen = () => {
               overrides={{
                 Root: {
                   style: {
-                    flex: 1, // Makes input take full available width
+                    flex: 1,
                     height: "48px",
                     background: COLORS.gray700,
+                    width: "100%",
```
ukorvl, line 162: can you please explain why do we need `with: 100%`?

ukorvl (approved):
LGTM

PR #314, opened 2025-02-17
Fix issue with wallet top-up and sendCurrency

gitshreevatsa:
The latest patch of niljs on the main branch needs amount or value parameter as type bigint instead of type number. Fixed the occurences of these instances to the correct datatype.

### DIFF
--- a/wallet-extension/src/features/blockchain/faucet.ts
+++ b/wallet-extension/src/features/blockchain/faucet.ts
@@ -16,7 +16,7 @@ export function createFaucetClient(rpcEndpoint: string): FaucetClient {
 export async function topUpAllCurrencies(
   smartAccount: SmartAccountV1,
   faucetClient: FaucetClient,
-  amount = 10,
+  amount = 10n,
 ): Promise<void> {
   try {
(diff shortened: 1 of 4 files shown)


wallet-extension/src/pages/wallet/TopUp.tsx
```diff
@@ -91,7 +91,7 @@ export const TopUp = () => {
         selectedCurrency.label,
         finalAmount,
       );
-
+
```
olegrok, line 94: This change is redundant

wallet-extension/src/features/blockchain/faucet.ts
```diff
@@ -74,7 +74,7 @@ export async function topUpSpecificCurrency(
   if (symbol === Currency.NIL) {
     initialBalance = await fetchBalance(smartAccount);
   }
-
+  console.log(`Initial balance: ${initialBalance}`);
```
ukorvl, line 77: remove consolelog plz

ukorvl:
Generally, what is the idea behind making even small numbers of type `bigint`?

PR #318, opened 2025-02-18
[execution] unify config accessor management between gas prices and validators

olegrok:
We decided to change approach to configuration management a bit. At the moment of block generation (with height=h) we have committed known block (h-1). This block has a link to mainShardHash that is a base for config accessor.
Such approach is quite consistent and decrease chances to face with an issue when we don't have an access to main shard block because it's not replicated yet.

### DIFF
--- a/nil/internal/collate/replay_scheduler.go
+++ b/nil/internal/collate/replay_scheduler.go
@@ -68,13 +68,12 @@ func (s *ReplayScheduler) doReplay(ctx context.Context, blockId types.BlockNumbe
 	ctx, cancel := context.WithTimeout(ctx, s.params.Timeout)
 	defer cancel()
 
-	proposal, err := s.buildProposalFromPrevBlock(ctx, blockId)
+	proposal, prevBlock, err := s.buildProposalFromPrevBlock(ctx, blockId)
 	if err != nil {
 		return err
 	}
 
-	gen, err := execution.NewBlockGenerator(ctx, s.params.BlockGeneratorParams, s.txFabric,
-		&proposal.PrevBlockHash, proposal.GetMainShardHash(s.params.ShardId))
+	gen, err := execution.NewBlockGenerator(ctx, s.params.BlockGeneratorParams, s.txFabric, prevBlock)
 	if err != nil {
 		return err
 	}
@@ -107,36 +106,38 @@ func (s *ReplayScheduler) switchLastBlock(ctx context.Context, blockId types.Blo
 	return blockHash, rwTx.Commit()
 }
 
-func (s *ReplayScheduler) buildProposalFromPrevBlock(ctx context.Context, blockId types.BlockNumber) (*execution.Proposal, error) {
+func (s *ReplayScheduler) buildProposalFromPrevBlock(
+	ctx context.Context, blockId types.BlockNumber,
+) (*execution.Proposal, *types.Block, error) {
 	if s.params.ShardId == types.MainShardId {
-		return nil, errors.New("replay for masterchain is not supported")
+		return nil, nil, errors.New("replay for main shard is not supported")
 	}
 	if blockId == types.BlockNumber(0) {
-		return nil, errors.New("replay for zerostate-block is not supported")
+		return nil, nil, errors.New("replay for zerostate-block is not supported")
 	}
 
 	proposal := &execution.Proposal{PrevBlockId: blockId - 1}
 
-	// NOTE: masterchain last block isn't switched now
+	// NOTE: main shard last block isn't switched now
 	if hash, err := s.switchLastBlock(ctx, proposal.PrevBlockId); err != nil {
-		return nil, err
+		return nil, nil, err
 	} else {
 		proposal.PrevBlockHash = hash
 	}
 
 	roTx, err := s.txFabric.CreateRoTx(ctx)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 	defer roTx.Rollback()
 
 	prevBlock, err := db.ReadBlock(roTx, s.params.ShardId, proposal.PrevBlockHash)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 	blockToReplay, err := db.ReadBlockByNumber(roTx, s.params.ShardId, blockId)
 	if err != nil {
-		return nil, err
+		return nil, nil, err
 	}
 
 	proposal.MainChainHash = prevBlock.MainChainHash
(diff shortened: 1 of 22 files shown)


nil/internal/config/params.go
```diff
@@ -152,40 +152,65 @@ func mergeValidators(input []ListValidators) []ValidatorInfo {
...
+
+func GetCurrentConfigAccessorTx(tx db.RoTx, block *types.Block, shardId types.ShardId) (ConfigAccessor, error) {
+	var mainShardHash *common.Hash
+	if block != nil {
+		mainShardHash = &block.MainChainHash
+		if shardId.IsMainShard() {
+			mainShardHash = &block.PrevBlock
+			// The first block uses configuration from itself.
+			if mainShardHash.Empty() {
+				h := block.Hash(types.MainShardId)
```
Zerg1996, line 172: Can you please explain what happens here?

nil/internal/collate/replay_scheduler.go
```diff
@@ -107,36 +106,38 @@ func (s *ReplayScheduler) switchLastBlock(ctx context.Context, blockId types.Blo
 	return blockHash, rwTx.Commit()
 }
 
-func (s *ReplayScheduler) buildProposalFromPrevBlock(ctx context.Context, blockId types.BlockNumber) (*execution.Proposal, error) {
+func (s *ReplayScheduler) buildProposalFromPrevBlock(
+	ctx context.Context, blockId types.BlockNumber,
+) (*execution.Proposal, *types.Block, error) {
 	if s.params.ShardId == types.MainShardId {
-		return nil, errors.New("replay for masterchain is not supported")
+		return nil, nil, errors.New("replay for masterchain is not supported")
```
knazarov, line 113: `s/masterchain/main shard/g`

nil/internal/collate/syncer.go
```diff
@@ -413,22 +424,18 @@ func validateRepliedBlock(
 }
 
 func (s *Syncer) replayBlock(ctx context.Context, block *types.BlockWithExtractedData) error {
-	mainShardHash := block.Block.MainChainHash
-	if s.config.ShardId.IsMainShard() {
-		mainShardHash = block.Block.PrevBlock
+	roTx, err := s.db.CreateRoTx(ctx)
```
knazarov, line 427: In this and other places we create RO transactions in place. This is probably not a good thing to do as ideally we should create one top-level read-only stransaction instead. I don't think we should handle it in this ticket, but as a follow-up instead.

Zerg1996 (approved):
From my point of view - looks good!

nil/internal/collate/replay_scheduler.go
olegrok: Fixed

nil/internal/config/params.go
knazarov: I'm not sure I understand either. Didn't you already pass "prev block" as `block` parameter? What is the purpose of calling another `&block.PrevBlock` here?

olegrok: Configuration of the first block should be calculated on this block itself. Since it's the first block and you don't have any previous committed.

nil/internal/config/params.go
```diff
@@ -152,40 +152,65 @@ func mergeValidators(input []ListValidators) []ValidatorInfo {
...
+func GetCurrentConfigAccessor(ctx context.Context, database db.DB, block *types.Block, shardId types.ShardId) (ConfigAccessor, error) {
 	tx, err := database.CreateRoTx(ctx)
 	if err != nil {
 		return nil, err
 	}
 	defer tx.Rollback()
+	return GetCurrentConfigAccessorTx(tx, block, shardId)
+}
+
+func GetCurrentConfigAccessorTx(tx db.RoTx, block *types.Block, shardId types.ShardId) (ConfigAccessor, error) {
```
shermike, line 164: Can we follow the same naming here and use a name like `NewConfigAccessorFromBlock` or smth?

nil/internal/config/params.go
```diff
@@ -152,40 +152,65 @@ func mergeValidators(input []ListValidators) []ValidatorInfo {
...
+	var mainShardHash *common.Hash
+	if block != nil {
+		mainShardHash = &block.MainChainHash
+		if shardId.IsMainShard() {
```
shermike, line 168: It is not obvious why we get previous block only for main shard. Maybe it is worth adding some comments here to clarify this?

nil/internal/execution/block_generator.go
```diff
@@ -102,27 +95,48 @@ func (g *BlockGenerator) Rollback() {
...
+func (p *BlockGenerator) CollectGasPrices(prevBlockId types.BlockNumber) []types.Uint256 {
 	if !p.params.ShardId.IsMainShard() {
 		return nil
 	}
 
+	// Basically we load configuration from block.MainChainHash.
+	// But for main shard this value should be block.PrevBlock.
+	// The first block uses configuration from itself.
+	configBlockId := prevBlockId
+	if prevBlockId != 0 {
```
knazarov, line 107: Wouldn't this condition also trigger for block zero on regular shards?

olegrok: No. Because we have on the function top.
```go
        if !p.params.ShardId.IsMainShard() {
		return nil
	}
```

nil/internal/execution/block_generator.go
```diff
@@ -102,27 +95,48 @@ func (g *BlockGenerator) Rollback() {
...
+		configBlockId--
+	}
+
+	mainBlock, err := db.ReadBlockByNumber(p.rwTx, types.MainShardId, configBlockId)
```
knazarov, line 111: It is worth noting that we need to be more strict about not addressing things by "height", as in presence of forks these heights will resolve to something you may not expect. Not for this PR for sure, but just for future.

nil/internal/config/params.go
olegrok: Added "For the main shard MainChainHash is empty. So we use the hash of the previous block."

nil/internal/config/params.go
olegrok: Done

nil/internal/execution/block_generator.go
olegrok: This part can be replaced with two `ReadBlocks` - get current block by hash (lookup prevBlock by hash and then lookup its prevBlock by hash).

nil/internal/config/params.go
olegrok: What is the purpose of calling another &block.PrevBlock here

because mainShard has MainChainHash empty for itself. And we need to choose `block.PrevBlock`. Just for consistency.

knazarov (approved):

PR #322, opened 2025-02-18
[nild] Introduce caching to blockchain config

isergeyam:
This patch introduces an LRU cache for blockchain config. This cache is shared between syncers and validators. It could be useful if the validator gets behind, in which case we switch to the syncer, which can use the cached entries. It is also useful for the validators because we require the same height's config in different places throughout the validation. 

The caching is disabled in sharded suite because otherwise the cache will be shared between instances, but it is enabled in rpc suite.

### DIFF
--- a/nil/internal/config/cache.go
+++ b/nil/internal/config/cache.go
@@ -1,0 +1,173 @@
+package config
+
+import (
+	"context"
+	"sync"
+
+	"github.com/NilFoundation/nil/nil/common/check"
+	"github.com/NilFoundation/nil/nil/internal/db"
+	"github.com/NilFoundation/nil/nil/internal/types"
+	lru "github.com/hashicorp/golang-lru/v2"
+)
+
+var GlobalConfigCache *ConfigCache
+
+func InitGlobalConfigCache(nShards uint32, txFabric db.DB) error {
+	var err error
+	GlobalConfigCache, err = NewConfigCache(nShards, txFabric)
+	return err
+}
+
+func GetConfigParams(ctx context.Context, txFabric db.DB, shardId types.ShardId, height uint64) (*ConfigParams, error) {
+	if GlobalConfigCache != nil {
+		return GlobalConfigCache.GetParams(ctx, shardId, height)
+	}
+	value := &cacheValue{
+		txFabric: txFabric,
+		shardId:  shardId,
+		height:   height,
+	}
+	// SAFETY: value is not shared
+	if err := value.initUnsafe(ctx); err != nil {
+		return nil, err
+	}
+	return &value.ConfigParams, nil
+}
+
+const lruCacheSize = 16
+
+type ConfigCache struct {
+	configLru []*lru.Cache[uint64, *cacheValue]
+
+	txFabric db.DB
+}
+
+func NewConfigCache(nShards uint32, txFabric db.DB) (*ConfigCache, error) {
+	configLru := make([]*lru.Cache[uint64, *cacheValue], 0, nShards)
+	for range nShards {
+		cache, err := lru.New[uint64, *cacheValue](lruCacheSize)
+		if err != nil {
+			return nil, err
+		}
+		configLru = append(configLru, cache)
+	}
+	return &ConfigCache{
+		configLru: configLru,
+		txFabric:  txFabric,
+	}, nil
+}
+
+func (c *ConfigCache) GetParams(ctx context.Context, shardId types.ShardId, height uint64) (*ConfigParams, error) {
+	if int(shardId) >= len(c.configLru) {
+		return nil, types.NewError(types.ErrorShardIdIsTooBig)
+	}
+
+	cache := c.configLru[shardId]
+	value := &cacheValue{
+		txFabric: c.txFabric,
+		shardId:  shardId,
+		height:   height,
+	}
+
+	// Note:  this is suboptimal, but hashicorp/golang-lru doesn't provide GetOrAdd,
+	//		  there is a PR though: https://github.com/hashicorp/golang-lru/pull/170
+	cache.ContainsOrAdd(height, value)
+	value, ok := cache.Get(height)
+	check.PanicIfNot(ok)
+
+	value.init(ctx)
+	if value.err != nil {
+		// This is likely to happen if we try to get validators for a height that is not yet available.
+		// In this case, we should not cache the error, because the error is not permanent.
+		cache.Remove(height)
+		return nil, value.err
+	}
+	return &value.ConfigParams, nil
+}
+
+type ConfigParams struct {
+	ValidatorInfo []ValidatorInfo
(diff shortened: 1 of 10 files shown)


nil/internal/config/cache.go
```diff
@@ -0,0 +1,113 @@
...
+	"github.com/NilFoundation/nil/nil/internal/db"
+	"github.com/NilFoundation/nil/nil/internal/types"
+	lru "github.com/hashicorp/golang-lru/v2"
+	"github.com/rs/zerolog"
+)
+
+const lruCacheSize = 16
+
+type ConfigCache struct {
+	configLru []*lru.Cache[uint64, *cacheValue]
```
shermike, line 17: Is it possible to make LRU cache global and use it in all `ConfigAccessor`s at once without creating a separate entity `ConfigCache`?
In this case, all components will use cached config by default.

olegrok: I'd also vote for such option. It makes logic less invasive.

nil/internal/config/cache.go
```diff
@@ -0,0 +1,113 @@
...
+	if int(shardId) >= len(c.configLru) {
+		return nil, types.NewError(types.ErrorShardIdIsTooBig)
+	}
+
+	cache := c.configLru[shardId]
+	value := &cacheValue{
+		txFabric: c.txFabric,
+		shardId:  shardId,
+		height:   height,
+		logger:   logger,
```
olegrok, line 47: Do you really need it?

nil/internal/config/cache.go
```diff
@@ -0,0 +1,113 @@
...
+		return
+	}
+	v.L1BlockInfo, v.err = GetParamL1Block(configAccessor)
+}
+
+func (v *cacheValue) init(ctx context.Context) {
+	v.once.Do(func() {
+		v.initImpl(ctx)
+	})
+}
```
olegrok, line 113: Isn't it better than have `err` field in a struct?
```suggestion
func (v *cacheValue) init(ctx context.Context) error {
	var err error
	v.once.Do(func() {
		err = v.initImpl(ctx)
	})
	return err
}
```

nil/internal/config/cache.go
```diff
@@ -0,0 +1,113 @@
...
+	height  uint64
+
+	err error
+
+	once sync.Once
+
+	logger zerolog.Logger
+}
+
+func (v *cacheValue) initImpl(ctx context.Context) {
```
olegrok, line 88: Let's try something like:
```go
func newValue(...) (*cacheValue, error) {
    option1, err := get(...)
    ...
    optionN, err := get(...)
    return &cacheValue{O1: option1, ..., ON: optionN}, nil
}

```

nil/internal/config/cache.go
```diff
@@ -0,0 +1,113 @@
...
+func (c *ConfigCache) GetParams(ctx context.Context, shardId types.ShardId, height uint64, logger zerolog.Logger) (*ConfigParams, error) {
+	if int(shardId) >= len(c.configLru) {
+		return nil, types.NewError(types.ErrorShardIdIsTooBig)
+	}
+
+	cache := c.configLru[shardId]
+	value := &cacheValue{
+		txFabric: c.txFabric,
+		shardId:  shardId,
+		height:   height,
```
olegrok, line 46: That also looks unused

nil/internal/config/cache.go
isergeyam: This was my initial approach, but I've stumbled upon the problem in our testing framework. We use shared memory for different nodes [here](https://github.com/NilFoundation/nil/blob/5a34529e8bec4ab2356aa96f612b8f3aad21be91/nil/tests/sharded_suite.go#L196), so the global cache, given it is instansiated with `init()` in the module, would be shared across all nodes. Do you have some workaround in mind for that?

shermike: You can disable global cache for testing. It doesn't really matter there, because it only affects performance.

nil/internal/config/cache.go
isergeyam: The idea here is to get params only once for concurrent execution, that's why we create a value object that is emplaced in the cache(if doesn't exist), and then we get the value again, so after that every concurrent participant has the same value object(the same pointer), and then we do sync.Once in that object.

nil/internal/config/cache.go
isergeyam: The idea here is to get params only once for concurrent execution, that's why we create a value object that is emplaced in the cache(if doesn't exist), and then we get the value again, so after that every concurrent participant has the same value object(the same pointer), and then we do sync.Once in that object.

nil/internal/config/cache.go
isergeyam: We want to share the error for concurrent participants, see the comment below

nil/internal/config/cache.go
isergeyam: Done

nil/internal/config/cache.go
```diff
@@ -0,0 +1,175 @@
...
+		return GlobalConfigCache.GetParams(ctx, shardId, height)
+	}
+	value := &cacheValue{
+		txFabric: txFabric,
+		shardId:  shardId,
+		height:   height,
+	}
+	// SAFETY: value is not shared
+	value.initUnsafe(ctx)
+	if value.err != nil {
```
shermike, line 32: Why do we save `error` in the `cacheValue`? It seems that it is used only here, thus maybe just return error from `initUnsafe`?

isergeyam: We need to store it in case of concurrent execution so that all concurrent participants receive the same error that actually happened during sync.Once execution, but in this case it is actually not needed, fixed

nil/tests/sharded_suite.go
```diff
@@ -330,6 +334,8 @@ func (s *ShardedSuite) StartRPCNode(dhtBootstrapByValidators DhtBootstrapByValid
 		netCfg.DHTBootstrapPeers = slices.Collect(common.Transform(slices.Values(s.Instances), getShardAddress))
 	}
 	cfg.RpcNode.ArchiveNodeList = archiveNodes
+	// Disable config cache because otherwise it would be shared between instances
+	cfg.DisableConfigCache = true
```
olegrok, line 338: Can we disable it by default and enable only for nilservice (where we create default config)?

Then we will have only one point where we enable it. Otherwise it's easy to forget about it in tests. And spend hours debugging what happens.

olegrok: Maybe we want to test this cache somehow in fact. Otherwise there are no guarantees that it actually works.

isergeyam: Ok, done, now we have it disabled by default. It is enabled in rpc suite, so it is being tested

Issue #39, opened 2025-01-29
Fix versioning


isergeyam:
After moving to the new repository, the old scheme that we used (git revisions for versioning) doesn't work, so we need a dirty hack like [this](https://github.com/NilFoundation/nil/pull/36) to make it work for the current release. We should discuss what kind of versioning we want; I think the best way would be to use semantic versioning like we do for NilJS.

PR #329, opened 2025-02-19
[nild] Remove BaseRevision

Zerg1996:
I think all our clients already update they cli to a new versions, that's why we don't need to handle corner case with old [revision](https://github.com/NilFoundation/nil/issues/39).

### DIFF
--- a/nil/services/rpc/internal/http/connection.go
+++ b/nil/services/rpc/internal/http/connection.go
@@ -3,14 +3,19 @@ package http
 import (
 	"io"
 	"net/http"
+	"strconv"
 	"time"
+
+	"github.com/NilFoundation/nil/nil/common/check"
+	"github.com/NilFoundation/nil/nil/common/version"
 )
 
 const (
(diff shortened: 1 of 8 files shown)


nil/services/rpc/internal/http/connection.go
```diff
@@ -3,18 +3,20 @@ package http
...
 )
 
 const (
 	MaxRequestContentLength  = 1024 * 1024 * 32 // 32MB
-	currentHeight            = 356
-	minSupportedRevision     = version.BaseRevision + currentHeight
 	minSupportedNiljsVersion = "0.24.0"
 )
 
+var minSupportedRevision = 1
```
knazarov, line 18: Keep this as a constant. We do need to set this manually (likely)

Zerg1996: Done!

knazarov (approved):
image

Gezort (approved):

PR #332, opened 2025-02-19
[execution] make main shard validation self-sufficient 

olegrok:
Before this patch we had following schema:
Main shard proposer should choose main shard hashes and send them to validators.
Validators update main shard config with corresponding gas prices of that shards.

Such schema has following pitfalls: we have all syncers dependent from each other
because main shard should be able to read their gas prices. And other shards
need a config from main shard. Such approach makes synchronization quite complex:
main shard should wait 1+ shards and 1+ shards should wait for the main shard.

After this patch validation of main shard will be independent because
we include gas prices of shard blocks into proposal.
Reverse operation is performed by using gas prices from configuration.

### DIFF
--- a/nil/internal/collate/block_listener.go
+++ b/nil/internal/collate/block_listener.go
@@ -212,13 +212,15 @@ func SetRequestHandler(
 		}
 		defer tx.Rollback()
 
+		acc := accessor.RawAccess(tx, shardId).
+			GetBlock().
+			WithOutTransactions().
+			WithInTransactions().
+			WithChildBlocks().
+			WithConfig()
+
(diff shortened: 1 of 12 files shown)


shermike:
Do I understand correctly that now each proposal will have a whole copy of config params?

olegrok:
> Do I understand correctly that now each proposal will have a whole copy of config params?

For now it has only gas prices

shermike:
> For now it has only gas prices

My bad, I meant `BlockWithExtractedData`, which is used in `Validator::InsertProposal`.

olegrok:
Also we do that only for a main shard.

shermike (approved):

PR #335, opened 2025-02-19
Allow to specify pprof port in nild command-line arguments

knazarov:
Currently we run multiple instances in production, and pprof port is always 6600 without any ability to override it. This means that only one instance (randomly) succeeds at opening this port first, and we can only connect to that specific instance to get profiling information.

Right now we're trying to figure out a few bugs that happen where an ability to take stack trace of all threads is very important. Thus, this PR makes pprof port configurable.

### DIFF
--- a/nil/cmd/nild/main.go
+++ b/nil/cmd/nild/main.go
@@ -36,7 +36,7 @@ func main() {
 
 	logging.ApplyComponentsFilter(logFilter)
 
-	profiling.Start(profiling.DefaultPort)
+	profiling.Start(cfg.PprofPort)
 
 	database, err := openDb(cfg.DB.Path, cfg.DB.AllowDrop, logger)
 	check.PanicIfErr(err)
@@ -148,6 +148,7 @@ func parseArgs() *nildconfig.Config {
 
 	// For backward compatibility
 	rootCmd.PersistentFlags().IntVar(&cfg.RPCPort, "port", cfg.RPCPort, "http port for rpc server")
+	rootCmd.PersistentFlags().IntVar(&cfg.PprofPort, "pprof-port", cfg.PprofPort, "port to serve pprof profiling information")
 	check.PanicIfErr(rootCmd.PersistentFlags().MarkHidden("port"))
 
 	runCmd := &cobra.Command{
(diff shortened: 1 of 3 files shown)


nil/services/nilservice/config.go
```diff
@@ -43,6 +43,9 @@ type Config struct {
 	RPCPort        int                   `yaml:"rpcPort,omitempty"`
 	BootstrapPeers network.AddrInfoSlice `yaml:"bootstrapPeers,omitempty"`
 
+	// Profiling
+	PprofPort int `yaml:"pprofPort,omitempty"`
```
Gezort, line 47: you may also need to change `devnet.go` file

Gezort (approved):

nil/cmd/nild/main.go
```diff
@@ -148,6 +148,7 @@ func parseArgs() *nildconfig.Config {
 
 	// For backward compatibility
 	rootCmd.PersistentFlags().IntVar(&cfg.RPCPort, "port", cfg.RPCPort, "http port for rpc server")
+	rootCmd.PersistentFlags().IntVar(&cfg.PprofPort, "pprof-port", cfg.PprofPort, "port to serve pprof profiling information")
```
olegrok, line 151: You need to add default to `NewDefaultConfig`. Otherwise without option you will start pprof on 0 port

knazarov: Yep, you're right. Fixed.

nil/services/nilservice/config.go
knazarov: Done.

knazarov:
I haven't yet tested config generation, because I don't yet know how. But I'll do that, just to make sure the pprof port is correctly exposed.

nil/cmd/nild/main.go
```diff
@@ -36,7 +36,7 @@ func main() {
 
 	logging.ApplyComponentsFilter(logFilter)
 
-	profiling.Start(profiling.DefaultPort)
```
olegrok, line 39: So you can finally drop `profiling.DefaultPort` variable I guess

PR #340, opened 2025-02-19
set wallet input if endpoint exists

0xAleksaOpacic:
If endpoint is saved in the store  it will trigger wallet creation. Which means that if you open wallet it will show loading on endpoint page. Usually when network is live this is done fast so you don't see it, but if network is down we have strange ui where something is loading but input is empty

### DIFF
--- a/explorer_frontend/src/features/account-connector/components/EndpointScreen.tsx
+++ b/explorer_frontend/src/features/account-connector/components/EndpointScreen.tsx
@@ -11,6 +11,7 @@ import animationData from "../assets/wallet-creation.json";
 import {
   $balance,
   $balanceToken,
+  $endpoint,
   $initializingSmartAccountError,
   $initializingSmartAccountState,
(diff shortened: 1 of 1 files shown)


explorer_frontend/src/features/account-connector/components/EndpointScreen.tsx
```diff
@@ -99,6 +102,12 @@ const EndpointScreen = () => {
     }
   }, [smartAccount, balance, balanceToken]);
 
+  useEffect(() => {
```
ukorvl, line 105: What about providing $endpoint as a default value here?
`const [inputValue, setInputValue] = useState("");`

0xAleksaOpacic: Much better, you are right. Fixed

ukorvl (approved):

PR #341, opened 2025-02-19
Add main keys to configuration

Zerg1996:
In the future we need to get rid of generation mainKeys inside our cluster. But for now for replication - the best way to share them inside configuration and generate new one if we don't do it.

### DIFF
--- a/nil/cmd/nild/devnet.go
+++ b/nil/cmd/nild/devnet.go
@@ -2,6 +2,8 @@
 package main
 
 import (
+	"crypto/ecdsa"
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -9,6 +11,7 @@ import (
 	"github.com/NilFoundation/nil/nil/cmd/nild/nildconfig"
 	"github.com/NilFoundation/nil/nil/common/check"
 	"github.com/NilFoundation/nil/nil/internal/config"
+	nilcrypto "github.com/NilFoundation/nil/nil/internal/crypto"
 	"github.com/NilFoundation/nil/nil/internal/db"
 	"github.com/NilFoundation/nil/nil/internal/execution"
 	"github.com/NilFoundation/nil/nil/internal/keys"
@@ -127,11 +130,28 @@ func (devnet devnet) generateZeroState(nShards uint32, servers []server) (*execu
 			})
 		}
 	}
-	return &execution.ZeroStateConfig{
-		ConfigParams: execution.ConfigParams{
-			Validators: config.ParamValidators{Validators: validators},
-		},
-	}, nil
+	mainKeyPath := devnet.spec.NildCredentialsDir + "/keys.yaml"
+	_, mainPublicKey, err := execution.LoadMainKeys(mainKeyPath)
+	if err != nil && !errors.Is(err, os.ErrNotExist) {
+		return nil, err
+	}
+	if err != nil {
+		var mainPrivateKey *ecdsa.PrivateKey
+		mainPrivateKey, mainPublicKey, err = nilcrypto.GenerateKeyPair()
+		if err != nil {
+			return nil, err
+		}
+		if err := execution.DumpMainKeys(mainKeyPath, mainPrivateKey, mainPublicKey); err != nil {
+			return nil, err
+		}
+	}
+	zeroState, err := execution.CreateDefaultZeroStateConfig(mainPublicKey)
+	if err != nil {
+		return nil, err
+	}
+	zeroState.ConfigParams.Validators = config.ParamValidators{Validators: validators}
+
+	return zeroState, nil
 }
 
 func genDevnet(cmd *cobra.Command, args []string) error {
@@ -236,6 +256,7 @@ func (spec devnetSpec) makeServers(nodeSpecs []nodeSpec, basePort int, pprofBase
 		}
 		servers[i].credsDir = fmt.Sprintf("%s/%s-%d", spec.NildCredentialsDir, service, i)
 		servers[i].workDir = fmt.Sprintf("%s/%s-%d", baseDir, service, i)
+
 		var err error
 		servers[i].identity, err = spec.EnsureIdentity(servers[i])
 		if err != nil {
--- a/nil/internal/collate/syncer.go
+++ b/nil/internal/collate/syncer.go
@@ -30,7 +30,6 @@ type SyncerConfig struct {
 	ShardId         types.ShardId
 	Timeout         time.Duration // pull blocks if no new blocks appear in the topic for this duration
 	BootstrapPeers  []network.AddrInfo
-	ZeroState       string
 	ZeroStateConfig *execution.ZeroStateConfig
 }
 
@@ -357,23 +356,14 @@ func (s *Syncer) GenerateZerostate(ctx context.Context) error {
 		return nil
 	}
 
-	if len(s.config.BlockGeneratorParams.MainKeysPath) != 0 && s.config.ShardId == types.BaseShardId {
-		if err := execution.LoadMainKeys(s.config.BlockGeneratorParams.MainKeysPath); err != nil {
-			if err := execution.DumpMainKeys(s.config.BlockGeneratorParams.MainKeysPath); err != nil {
-				return err
-			}
-		}
-	}
-
 	s.logger.Info().Msg("Generating zero-state...")
-
 	gen, err := execution.NewBlockGenerator(ctx, s.config.BlockGeneratorParams, s.db, nil, nil)
 	if err != nil {
 		return err
 	}
 	defer gen.Rollback()
 
-	if _, err := gen.GenerateZeroState(s.config.ZeroState, s.config.ZeroStateConfig); err != nil {
+	if _, err := gen.GenerateZeroState(s.config.ZeroStateConfig); err != nil {
 		return err
 	}
 	return nil
--- a/nil/internal/execution/testaide.go
+++ b/nil/internal/execution/testaide.go
@@ -4,25 +4,35 @@ package execution
 
 import (
 	"context"
+	"crypto/ecdsa"
 	"testing"
 
 	"github.com/NilFoundation/nil/nil/common"
+	"github.com/NilFoundation/nil/nil/common/check"
 	"github.com/NilFoundation/nil/nil/internal/config"
 	"github.com/NilFoundation/nil/nil/internal/contracts"
+	nilcrypto "github.com/NilFoundation/nil/nil/internal/crypto"
 	"github.com/NilFoundation/nil/nil/internal/db"
 	"github.com/NilFoundation/nil/nil/internal/types"
 	"github.com/stretchr/testify/require"
+)
+
+var (
+	MainPrivateKey   *ecdsa.PrivateKey
+	MainPublicKey    []byte
+	DefaultGasCredit = types.Gas(DefaultGasLimit).ToValue(types.DefaultGasPrice)
+	DefaultSendValue = types.GasToValue(200_000_000)
 )
 
 const (
 	DefaultGasLimit = 100_000
 )
 
-var (
-	DefaultGasCredit = types.Gas(DefaultGasLimit).ToValue(types.DefaultGasPrice)
-
-	DefaultSendValue = types.GasToValue(200_000_000)
-)
+func init() {
+	var err error
+	MainPrivateKey, MainPublicKey, err = nilcrypto.GenerateKeyPair()
+	check.PanicIfErr(err)
+}
 
 func GenerateZeroState(t *testing.T, ctx context.Context,
 	shardId types.ShardId, txFabric db.DB,
@@ -35,15 +45,14 @@ func GenerateZeroState(t *testing.T, ctx context.Context,
 	require.NoError(t, err)
 	defer g.Rollback()
 
-	zerostateCfg, err := ParseZeroStateConfig(DefaultZeroStateConfig)
+	zerostateCfg, err := CreateDefaultZeroStateConfig(MainPublicKey)
 	require.NoError(t, err)
 	zerostateCfg.ConfigParams = ConfigParams{
 		GasPrice: config.ParamGasPrice{
 			Shards: []types.Uint256{*types.NewUint256(10), *types.NewUint256(10), *types.NewUint256(10)},
 		},
 	}
-
-	block, err := g.GenerateZeroState("", zerostateCfg)
+	block, err := g.GenerateZeroState(zerostateCfg)
 	require.NoError(t, err)
 	require.NotNil(t, block)
 	return block.Hash(shardId)
--- a/nil/internal/execution/zerostate.go
+++ b/nil/internal/execution/zerostate.go
@@ -16,22 +16,6 @@ import (
 	"gopkg.in/yaml.v3"
 )
 
-var (
-	MainPrivateKey *ecdsa.PrivateKey
-	MainPublicKey  []byte
-)
-
-var DefaultZeroStateConfig string
-
-func init() {
-	var err error
-	MainPrivateKey, MainPublicKey, err = nilcrypto.GenerateKeyPair()
-	check.PanicIfErr(err)
-
-	DefaultZeroStateConfig, err = CreateDefaultZeroStateConfig(MainPublicKey)
-	check.PanicIfErr(err)
-}
-
 type ContractDescr struct {
 	Name     string         `yaml:"name"`
 	Address  *types.Address `yaml:"address,omitempty"`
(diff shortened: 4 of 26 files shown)


nil/services/nilservice/service.go
```diff
@@ -369,6 +365,17 @@ func CreateNode(ctx context.Context, name string, cfg *Config, database db.DB, i
 			}
 		}
 
+		if cfg.ZeroState.MainKeys.MainPrivateKey != "" && cfg.ZeroState.MainKeys.MainPublicKey != "" {
+			execution.MainPrivateKey, err = crypto.HexToECDSA(cfg.ZeroState.MainKeys.MainPrivateKey)
```
Un1oR: Hell, I think mutable global variable is the root of our problems.
Since there is no normal constant semantics in freaking Go (like in C++), these variables are basically `var`. But they were "logical" constants, initialized in `init()`. Now we start actually rewriting them in a completely different part of the program 🤯 Honestly, I think we should get rid of such constants (now it's not even a "singleton", it's even worse!) and fix the whole code to explicitly pass the corresponding values.

Zerg1996: I'm agree with that. Still this PR mostly about generating mainKeys inside nil gen config and we can't properly start archive nodes without that. So get rid of global variables take some time and I think it's better to make it in another PR.

nil/cmd/nild/devnet.go
```diff
@@ -125,10 +127,17 @@ func (devnet devnet) generateZeroState(nShards uint32, servers []server) (*execu
...
+	if err != nil {
+		return nil, err
+	}
+	mainKeys := execution.MainKeys{MainPrivateKey: "0x" + nilcrypto.PrivateKeyToEthereumFormat(mainPrivateKey), MainPublicKey: hexutil.Encode(mainPublicKey)}
+
 	return &execution.ZeroStateConfig{
 		ConfigParams: execution.ConfigParams{
 			Validators: config.ParamValidators{Validators: validators},
 		},
+		MainKeys: mainKeys,
```
olegrok, line 140: I'm a bit confused here. I'd suggest to generate yaml config file and share it between all instances.
You don't need to pass `MainPrivateKey` somewhere. All you need for zerostate is only public key.
You can just dump private key into credentials directory and reuse it if it exists.

Zerg1996: yes, you are right. fixed it and create keys file.

nil/internal/collate/syncer.go
```diff
@@ -358,11 +358,9 @@ func (s *Syncer) GenerateZerostate(ctx context.Context) error {
 		return nil
 	}
 
-	if len(s.config.BlockGeneratorParams.MainKeysPath) != 0 && s.config.ShardId == types.BaseShardId {
```
olegrok, line 361: At least for now you don't need to load keys for everything except base shard. Why do you need this change?

nil/cmd/nild/devnet.go
```diff
@@ -229,6 +238,10 @@ func (spec devnetSpec) makeServers(nodeSpecs []nodeSpec, basePort int, baseHTTPP
 		servers[i].credsDir = fmt.Sprintf("%s/%s-%d", spec.NildCredentialsDir, service, i)
 		servers[i].workDir = fmt.Sprintf("%s/%s-%d", baseDir, service, i)
 		var err error
+
+		if err := execution.DumpMainKeys(server.MainKeysFile(servers[i])); err != nil {
```
olegrok, line 242: Isn't it enough to dump keys to single common place (not for each server)?

nil/internal/collate/syncer.go
Zerg1996: Because now the main purpose of this keys - to make validator execute some transactions using this case. This means, that if I run only node with shard0+2, I'm also want to use MainPrivateKey to execute transactions.

olegrok: That's pretty strange. We use this key only as a key for main smart account. It's not related to validators anyhow.

olegrok: However, seems I got all you meant

Zerg1996: Okay. But who can use this key? 
If we load it only for baseShard only validator on BaseShard can use it. That’s also strange behavior.

nil/internal/collate/syncer.go
```diff
@@ -358,11 +358,9 @@ func (s *Syncer) GenerateZerostate(ctx context.Context) error {
...
+	if len(s.config.BlockGeneratorParams.MainKeysPath) != 0 {
 		if err := execution.LoadMainKeys(s.config.BlockGeneratorParams.MainKeysPath); err != nil {
-			if err := execution.DumpMainKeys(s.config.BlockGeneratorParams.MainKeysPath); err != nil {
-				return err
-			}
+			s.logger.Debug().Str("path", s.config.BlockGeneratorParams.MainKeysPath).Msg("Failed to load main keys, using default")
```
Un1oR, line 363: I think it would be better to have a warning level for this log

nil/cmd/nild/devnet.go
```diff
@@ -245,6 +258,10 @@ func (srv server) NetworkKeysFile() string {
 	return srv.credsDir + "/network-keys.yaml"
 }
 
+func (srv server) MainKeysFile() string {
```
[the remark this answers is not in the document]
Zerg1996: Yes, it'll be removed and place directly to /etc/nild/key.yaml file.

nil/internal/execution/testaide.go
```diff
@@ -4,25 +4,87 @@ package execution
...
+  address: {{ .L1BlockInfoAddress }}
+  value: 0
+  contract: system/L1BlockInfo
+`
+
+	res, err := common.ParseTemplate(zerostate, map[string]interface{}{
+		"MainSmartAccountAddress": types.MainSmartAccountAddress.Hex(),
+		"L1BlockInfoAddress":      types.L1BlockInfoAddress.Hex(),
+		"MainPublicKey":           hexutil.Encode(MainPublicKey),
+		"MainSmartAccountPubKey":  hexutil.Encode(MainPublicKey),
```
olegrok, line 75: There is no such option in template

nil/internal/execution/zerostate.go
```diff
@@ -93,22 +74,44 @@ contracts:
...
+	}
 
-	return common.ParseTemplate(zerostate, map[string]interface{}{
+	res, err := common.ParseTemplate(zerostate, map[string]interface{}{
 		"MainSmartAccountAddress": types.MainSmartAccountAddress.Hex(),
 		"L1BlockInfoAddress":      types.L1BlockInfoAddress.Hex(),
-		"MainPublicKey":           hexutil.Encode(MainPublicKey),
-		"MainSmartAccountPubKey":  hexutil.Encode(MainPublicKey),
+		"MainPublicKey":           hexutil.Encode(mainPublicKey),
+		"MainSmartAccountPubKey":  hexutil.Encode(mainPublicKey),
```
olegrok, line 89: The same. You can remove it.

nil/internal/execution/zerostate.go
```diff
@@ -52,15 +36,12 @@ type ConfigParams struct {
...
 type ZeroStateConfig struct {
-	ConfigParams ConfigParams     `yaml:"config,omitempty"`
-	Contracts    []*ContractDescr `yaml:"contracts"`
-}
-
-func (cfg *ZeroStateConfig) GetValidators() []config.ListValidators {
-	return cfg.ConfigParams.Validators.Validators
+	ConfigParams  ConfigParams     `yaml:"config,omitempty"`
+	Contracts     []*ContractDescr `yaml:"contracts"`
+	MainPublicKey string           `yaml:"mainPublicKey"`
```
olegrok, line 41: Let's simplify code a bit:
```diff
diff --git a/nil/internal/execution/execution_state_test.go b/nil/internal/execution/execution_state_test.go
index 7e1f1a41..830fbcc4 100644
--- a/nil/internal/execution/execution_state_test.go
+++ b/nil/internal/execution/execution_state_test.go
@@ -652,7 +652,7 @@ contracts:
 
 	zeroState, err := ParseZeroStateConfig(zerostateCfg)
 	require.NoError(b, err)
-	zeroState.MainPublicKey = hexutil.Encode(MainPublicKey)
+	zeroState.MainPublicKey = MainPublicKey
 
 	params := NewBlockGeneratorParams(1, 2)
 
diff --git a/nil/internal/execution/testaide.go b/nil/internal/execution/testaide.go
index dc5c4695..86fa5f7f 100644
--- a/nil/internal/execution/testaide.go
+++ b/nil/internal/execution/testaide.go
@@ -82,7 +82,7 @@ contracts:
 
 	DefaultZeroStateConfig, err = ParseZeroStateConfig(res)
 	check.PanicIfErr(err)
-	DefaultZeroStateConfig.MainPublicKey = hexutil.Encode(MainPublicKey)
+	DefaultZeroStateConfig.MainPublicKey = MainPublicKey
 }
 
 func GenerateZeroState(t *testing.T, ctx context.Context,
@@ -103,7 +103,7 @@ func GenerateZeroState(t *testing.T, ctx context.Context,
 			Shards: []types.Uint256{*types.NewUint256(10), *types.NewUint256(10), *types.NewUint256(10)},
 		},
 	}
-	zerostateCfg.MainPublicKey = hexutil.Encode(MainPublicKey)
+	zerostateCfg.MainPublicKey = MainPublicKey
 
 	block, err := g.GenerateZeroState(zerostateCfg)
 	require.NoError(t, err)
diff --git a/nil/internal/execution/zerostate.go b/nil/internal/execution/zerostate.go
index eae1f77d..d4bb9629 100644
--- a/nil/internal/execution/zerostate.go
+++ b/nil/internal/execution/zerostate.go
@@ -38,7 +38,7 @@ type ConfigParams struct {
 type ZeroStateConfig struct {
 	ConfigParams  ConfigParams     `yaml:"config,omitempty"`
 	Contracts     []*ContractDescr `yaml:"contracts"`
-	MainPublicKey string           `yaml:"mainPublicKey"`
+	MainPublicKey hexutil.Bytes    `yaml:"mainPublicKey"`
 }
 
 func CreateDefaultZeroStateConfig(mainPublicKey []byte) (*ZeroStateConfig, error) {
@@ -100,7 +100,7 @@ contracts:
 	if err != nil {
 		return nil, err
 	}
-	zeroStateConfig.MainPublicKey = hexutil.Encode(mainPublicKey)
+	zeroStateConfig.MainPublicKey = mainPublicKey
 
 	return zeroStateConfig, nil
 }
@@ -212,18 +212,13 @@ func (es *ExecutionState) GenerateZeroState(stateConfig *ZeroStateConfig) error
 			return err
 		}
 
-		mainPublicKey, err := hexutil.Decode(stateConfig.MainPublicKey)
-		if err != nil {
-			return err
-		}
-
 		args := make([]any, 0)
 		for _, arg := range contract.CtorArgs {
 			switch arg := arg.(type) {
 			case string:
 				switch {
 				case arg == "MainPublicKey":
-					args = append(args, mainPublicKey)
+					args = append(args, stateConfig.MainPublicKey[:])
 				case arg[:2] == "0x":
 					args = append(args, hexutil.FromHex(arg))
 				default:
diff --git a/nil/internal/execution/zerostate_test.go b/nil/internal/execution/zerostate_test.go
index 075b0a84..36a38043 100644
--- a/nil/internal/execution/zerostate_test.go
+++ b/nil/internal/execution/zerostate_test.go
@@ -8,7 +8,6 @@ import (
 	"testing"
 
 	"github.com/NilFoundation/nil/nil/common"
-	"github.com/NilFoundation/nil/nil/common/hexutil"
 	"github.com/NilFoundation/nil/nil/internal/abi"
 	"github.com/NilFoundation/nil/nil/internal/config"
 	"github.com/NilFoundation/nil/nil/internal/contracts"
@@ -125,7 +124,7 @@ config:
 
 	zeroState, err := ParseZeroStateConfig(configYaml)
 	require.NoError(t, err)
-	zeroState.MainPublicKey = hexutil.Encode(MainPublicKey)
+	zeroState.MainPublicKey = MainPublicKey
 
 	err = state.GenerateZeroState(zeroState)
 	require.NoError(t, err)
@@ -137,7 +136,7 @@ config:
 	require.NoError(t, err)
 	zeroState, err = ParseZeroStateConfig(configYaml)
 	require.NoError(t, err)
-	zeroState.MainPublicKey = hexutil.Encode(MainPublicKey)
+	zeroState.MainPublicKey = MainPublicKey
 
 	err = state.GenerateZeroState(zeroState)
 	require.NoError(t, err)
@@ -149,7 +148,7 @@ config:
 	require.NoError(t, err)
 	zeroState, err = ParseZeroStateConfig(configYaml)
 	require.NoError(t, err)
-	zeroState.MainPublicKey = hexutil.Encode(MainPublicKey)
+	zeroState.MainPublicKey = MainPublicKey
 
 	err = state.GenerateZeroState(zeroState)
 	require.NoError(t, err)
@@ -172,7 +171,7 @@ contracts:
 	require.NoError(t, err)
 	zeroState, err = ParseZeroStateConfig(configYaml)
 	require.NoError(t, err)
-	zeroState.MainPublicKey = hexutil.Encode(MainPublicKey)
+	zeroState.MainPublicKey = MainPublicKey
 
 	err = state.GenerateZeroState(zeroState)
 	require.NoError(t, err)
@@ -203,7 +202,7 @@ contracts:
 	require.NoError(t, err)
 	zeroState, err = ParseZeroStateConfig(configYaml2)
 	require.NoError(t, err)
-	zeroState.MainPublicKey = hexutil.Encode(MainPublicKey)
+	zeroState.MainPublicKey = MainPublicKey
 
 	err = state.GenerateZeroState(zeroState)
 	require.Error(t, err)
@@ -227,7 +226,7 @@ contracts:
 	require.NoError(t, err)
 	zeroState, err = ParseZeroStateConfig(configYaml3)
 	require.NoError(t, err)
-	zeroState.MainPublicKey = hexutil.Encode(MainPublicKey)
+	zeroState.MainPublicKey = MainPublicKey
 
 	err = state.GenerateZeroState(zeroState)
 	require.NoError(t, err)
```

nil/internal/execution/zerostate.go
```diff
@@ -93,22 +74,43 @@ contracts:
...
+}
+
+func (cfg *ZeroStateConfig) GetValidators() []config.ListValidators {
+	return cfg.ConfigParams.Validators.Validators
 }
 
-func DumpMainKeys(fname string) error {
-	keys := MainKeys{"0x" + nilcrypto.PrivateKeyToEthereumFormat(MainPrivateKey), hexutil.Encode(MainPublicKey)}
+func DumpMainKeys(fname string, mainPrivateKey *ecdsa.PrivateKey, mainPublicKey []byte) error {
+	keys := MainKeys{"0x" + nilcrypto.PrivateKeyToEthereumFormat(mainPrivateKey), hexutil.Encode(mainPublicKey)}
```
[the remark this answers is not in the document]
Zerg1996: Yes, thanks. You are right. That's look much better.

nil/cmd/nild/devnet.go
```diff
@@ -127,11 +128,21 @@ func (devnet devnet) generateZeroState(nShards uint32, servers []server) (*execu
...
-		},
-	}, nil
+	mainPrivateKey, mainPublicKey, err := nilcrypto.GenerateKeyPair()
+	if err != nil {
+		return nil, err
+	}
+	if err := execution.DumpMainKeys(devnet.spec.NildCredentialsDir+"/keys.yaml", mainPrivateKey, mainPublicKey); err != nil {
+		return nil, err
+	}
+	zeroState, err := execution.CreateDefaultZeroStateConfig(nil)
```
olegrok, line 138: Why not

```suggestion
	zeroState, err := execution.CreateDefaultZeroStateConfig(mainPublicKey)
```

olegrok (approved):

nil/internal/execution/testaide.go
```diff
@@ -35,15 +96,16 @@ func GenerateZeroState(t *testing.T, ctx context.Context,
 	require.NoError(t, err)
 	defer g.Rollback()
 
-	zerostateCfg, err := ParseZeroStateConfig(DefaultZeroStateConfig)
+	zerostateCfg := DefaultZeroStateConfig
```
knazarov, line 99: Shouldn't we copy this value?

nil/internal/execution/zerostate.go
```diff
@@ -125,29 +127,21 @@ func DumpMainKeys(fname string) error {
 	return err
 }
 
-func LoadMainKeys(fname string) error {
+func LoadMainKeys(fname string) (*ecdsa.PrivateKey, []byte, error) {
```
knazarov, line 130: Why do we return private key as a proper type and public key as sequence of bytes?

knazarov (approved):
A few nitpicks, but I don't see problems with this.

nil/internal/execution/testaide.go
Zerg1996: Changed to `zerostateCfg, err := CreateDefaultZeroStateConfig(MainPublicKey)`

nil/internal/execution/zerostate.go
Zerg1996: Yes, will be fixed in next PR.

PR #350, opened 2025-02-20
Sync Committee Proposer: State root initialization

zadykian:
### Sync Committee Proposer: State root initialization

Modified the Proposer's initialization logic to prevent service termination in case of an L1 state root initialization error

### DIFF
--- a/nil/services/synccommittee/core/proposer.go
+++ b/nil/services/synccommittee/core/proposer.go
@@ -89,9 +89,9 @@ func NewProposer(
 
 	p.logger = srv.WorkerLogger(logger, p)
 
-	err := p.tryGetRollupContractWrapper(ctx)
+	err := p.initRollupContractWrapper(ctx)
 	if err != nil {
-		logger.Error().Msgf("rollup contract wrapper is not initialized: %s", err)
+		logger.Error().Err(err).Msgf("rollup contract wrapper is not initialized")
 	}
 
 	return p, nil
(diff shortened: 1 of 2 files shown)


nil/services/synccommittee/core/proposer.go
```diff
@@ -122,53 +121,63 @@ func (p *proposer) Run(ctx context.Context, started chan<- struct{}) error {
...
-	if storedStateRoot == nil || *storedStateRoot != latestStateRoot {
-		err = p.storage.SetProvedStateRoot(ctx, latestStateRoot)
-		if err != nil {
-			return fmt.Errorf("failed set proved state root: %w", err)
+		if err := p.updateStoredStateRoot(ctx, latestStateRoot); err != nil {
+			return err
 		}
+
+	default:
+		p.logger.Info().Stringer("stateRoot", storedStateRoot).Msg("proved state root value is valid")
```
akokoshn, line 180: may be better print it with debug level?

nil/services/synccommittee/core/proposer.go
```diff
@@ -122,53 +121,63 @@ func (p *proposer) Run(ctx context.Context, started chan<- struct{}) error {
...
 	latestStateRoot, err := p.getLatestProvedStateRoot(ctx)
-	if err != nil {
-		return fmt.Errorf("failed get current contract state root: %w", err)
-	}
 
 	switch {
-	case latestStateRoot == common.EmptyHash:
-		p.logger.Info().
-			Msg("L1 not available, will be used stored proved state root")
+	case err != nil:
```
akokoshn, line 152: If we'll not update storedStateRoot here (not initialized yet, and L1 contract not accessible), `TryGetNextProposalData` returns error, and fail SC process

akokoshn: no, seems process continue works

PR #355, opened 2025-02-20
[execution] Fix nested sendRequests

shermike:
If the callee of `sendRequest` also calls `sendRequest`, it leads to incorrect inheritance of the request chain.
`sendRequest` should not wait for another `sendRequest`, it should wait for only `awaitCall`.
Three new test cases covering these statements have been added.

### DIFF
--- a/nil/tests/async_await/async_await_test.go
+++ b/nil/tests/async_await/async_await_test.go
@@ -12,6 +12,7 @@ import (
 	"github.com/NilFoundation/nil/nil/internal/network"
 	"github.com/NilFoundation/nil/nil/internal/types"
 	"github.com/NilFoundation/nil/nil/services/nilservice"
+	"github.com/NilFoundation/nil/nil/services/rpc"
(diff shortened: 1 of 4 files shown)


nil/tests/async_await/async_await_test.go
```diff
@@ -4,21 +4,22 @@ import (
...
 	"github.com/NilFoundation/nil/nil/services/nilservice"
+	"github.com/NilFoundation/nil/nil/services/rpc"
 	"github.com/NilFoundation/nil/nil/services/rpc/jsonrpc"
 	"github.com/NilFoundation/nil/nil/tests"
 	"github.com/stretchr/testify/suite"
 )
 
 type SuiteAsyncAwait struct {
-	tests.ShardedSuite
+	tests.RpcSuite
```
olegrok, line 22: I'd suggest to enable it back just to be sure that cross-shard messaging here works

olegrok (approved):

PR #360, opened 2025-02-21
[types] extend block info with round and validator index

olegrok:
Info about the previous validator will allow you to choose the next one more honestly.

### DIFF
--- a/nil/internal/consensus/ibft/ibft.go
+++ b/nil/internal/consensus/ibft/ibft.go
@@ -32,7 +32,7 @@ type ConsensusParams struct {
 type validator interface {
 	BuildProposal(ctx context.Context) (*execution.Proposal, error)
 	VerifyProposal(ctx context.Context, proposal *execution.Proposal) (*types.Block, error)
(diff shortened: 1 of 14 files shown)


nil/internal/consensus/ibft/ibft.go
```diff
@@ -165,8 +165,18 @@ func (i *backendIBFT) InsertProposal(proposal *protoIBFT.Proposal, committedSeal
 		return // error is logged in buildSignature
 	}
 
-	if err := i.validator.InsertProposal(i.ctx, proposalBlock, sig); err != nil {
-		logger.Error().Err(err).Msg("failed to insert proposal")
+	prevProposerIndex := i.getPrevProposer(height)
+	_, proposerIndex, err := i.calcProposer(height, proposal.Round, prevProposerIndex)
+	if err != nil {
+		logger.Error().Err(err).Msg("Failed to calculate current proposal")
```
dmtrskv, line 171: propos*er*

dmtrskv (approved):

nil/internal/consensus/ibft/ibft.go
olegrok: Thanks!

PR #362, opened 2025-02-21
[explorer] drop unused files

olegrok:


knazarov (approved):
I'm fine with this. Hope you've grepped the codebase and made sure the types aren't really used.

explorer_frontend/src/types.ts
```diff
@@ -1,98 +0,0 @@
...
-  utime: number;
-}
-
-export type ContractState = {
-  decoded: null | Record<string, string | boolean>;
-  balance: string;
-  state: string;
-};
-
-export type App = {
```
ukorvl, line 93: App type is used in many places. The other types in this file are redundant.

olegrok: Restored. Thanks!

PR #364, opened 2025-02-21
Sync Committee: Block Storage Reset

zadykian:
### Sync Committee: Block Storage Reset

As part of the Soft State Reset logic, introduced several changes to the `BlockStorage` type:

* Implemented `BlockStorage.ResetProgress(hash)` method, which performs a state reset starting from the given main block hash (including stored exec shard blocks and the latest fetched refs);

* Added new tests: `Test_ResetProgress_Block_Does_Not_Exists()` and `Test_ResetProgress()` with multiple test cases for block reset logic;

* BlockStorage.getBlockEntry(id) method now accepts an additional `required bool` parameter and returns `scTypes.ErrBlockNotFound` error if the required block is not found;

* Added `scTypes.ErrBlockNotFound` to the list of errors ignored by `badgerRetryRunner`;

#### Blocks reset will be triggered by `taskStateChangeHandler` in a follow-up PR

### DIFF
--- a/nil/services/synccommittee/internal/storage/block_storage.go
+++ b/nil/services/synccommittee/internal/storage/block_storage.go
@@ -7,6 +7,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"iter"
 	"time"
 
 	"github.com/NilFoundation/nil/nil/common"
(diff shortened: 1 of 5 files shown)


nil/services/synccommittee/internal/storage/block_storage.go
```diff
@@ -520,15 +620,53 @@ func (bs *BlockStorage) getBlockEntry(tx db.RoTx, id scTypes.BlockId) (*blockEnt
...
+
+func (bs *BlockStorage) deleteBlock(tx db.RwTx, entry *blockEntry) error {
+	parentId := scTypes.ParentBlockId(&entry.Block)
+	err := tx.Delete(blockParentIdxTable, parentId.Bytes())
+	if err != nil && !errors.Is(err, db.ErrKeyNotFound) {
+		return fmt.Errorf("failed to delete parent idx entry, parentId=%s: %w", parentId, err)
+	}
+
+	blockId := scTypes.IdFromBlock(&entry.Block)
+	if err := tx.Delete(blocksTable, blockId.Bytes()); err != nil {
```
akokoshn, line 654: For now it should works, but after update fetching, we'll have to recursive delete parents of the block

nil/services/synccommittee/internal/storage/block_storage.go
```diff
@@ -496,20 +498,118 @@ func (bs *BlockStorage) putLatestFetchedBlockTx(tx db.RwTx, shardId types.ShardI
...
+	defer tx.Rollback()
+
+	startingId := scTypes.NewBlockId(types.MainShardId, firstMainHashToPurge)
+
+	for entry, err := range bs.getChainSequence(tx, startingId) {
+		if err != nil {
+			return err
+		}
+
+		if scTypes.IdFromBlock(&entry.Block) == startingId {
```
akokoshn, line 527: seems can be moved out of loop

akokoshn (approved):

nil/services/synccommittee/internal/storage/block_storage.go
zadykian: The idea here is to delete the whole batch at once from storage after state change is submitted to L1.

So, recursion shouldn't be required - we'll just need to call `deleteMainBlockWithChildren(...)` for each chunk in the batch

nil/services/synccommittee/internal/storage/block_storage.go
zadykian: done

PR #366, opened 2025-02-21
Fix fetching tx details in playground logs

ukorvl:
Before this patch we had an issue with infinite fetching transaction details in case of an error.
This diff adds `clearInterval(intervalId);` to stop trying fetch transaction if once it threw an error.
Also it brings minor styling improvements for an error message.

### DIFF
--- a/explorer_frontend/src/features/logs/hooks/useRepeatedGetTxByHash.ts
+++ b/explorer_frontend/src/features/logs/hooks/useRepeatedGetTxByHash.ts
@@ -6,6 +6,7 @@ import type { Transaction } from "../../transaction/types/Transaction";
 export const useRepeatedGetTxByHash = (
   txHash: Hex,
   interval = 1000,
+  retries = 3,
 ): {
   data: Transaction | null;
   error: boolean;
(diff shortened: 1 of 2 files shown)


explorer_frontend/src/features/logs/hooks/useRepeatedGetTxByHash.ts
```diff
@@ -34,6 +34,7 @@ export const useRepeatedGetTxByHash = (
       } catch (err) {
         setError(true);
         setLoading(false);
+        clearInterval(intervalId);
```
0xAleksaOpacic, line 37: Why are you closing interval on error? We don't want retry?

ukorvl: Yes, retry here makes a little sense, because if there is an error the reason is likely not to go by itself (becasue it is an error on backends side or error with connection). What i mean - we better display "Error while fetching tx details" in logs than spam with requests.
I don't see a case where retry can definitely help. If something is down on backend - it is not going to work in next 5-10 seconds with high probability. That's why i didn't add even 2-3 retries.

ukorvl: Ok, i thought about it and decided that you are correct. I'll add 3 retries. Actually, the issue was that there were unlimited retires, but 3-5 of them is a decent idea.

KlonD90 (approved):

PR #374, opened 2025-02-22
[execution] use deterministic time

olegrok:
We had problems with consensus. That's because the same EVM input produced different output that led to hash mismatch error. It was pretty hard to debug this issue. It broke consensus on load gen tests because uniswap sometimes operates with a time.

This patch removes using of `time.Now()` that breaks an idea of repeatable blockchain history.

Our block.Timestamp is always zero. So I suggest to use block.Id as a logical time. It also monotonically increases.

### DIFF
--- a/nil/internal/execution/state.go
+++ b/nil/internal/execution/state.go
@@ -10,7 +10,6 @@ import (
 	"math"
 	"math/big"
 	"sort"
-	"time"
 
 	"github.com/NilFoundation/nil/nil/common"
 	"github.com/NilFoundation/nil/nil/common/assert"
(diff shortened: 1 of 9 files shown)


nil/internal/execution/state.go
```diff
@@ -213,17 +212,21 @@ func NewEVMBlockContext(es *ExecutionState) (*vm.BlockContext, error) {
 
 	currentBlockId := uint64(0)
 	var header *types.Block
+	time := uint64(0)
 	if err == nil {
 		header = data.Block()
 		currentBlockId = header.Id.Uint64() + 1
+		// TODO: we need to use header.Timestamp instead of but it's always zero for now.
```
Gezort, line 219: we need to either exclude it from hashed fields and make non-zero or completely remove from the block

Gezort (approved):

PR #381, opened 2025-02-24
[network] add custom prefix to network protocols

olegrok:
Seems we have a conflict for archive node test. Exact reason why messages signed by different validators are accepted is a seprate task. But to make test more stable let's add custom prefix to our network protocols.

By default it's `/nil`.


olegrok:
> It looks like let's always use network prefix?

The idea was to fix archive node test via introducing custom prefixes for separate tests

PR #389, opened 2025-02-24
Remove generation from yaml

Zerg1996:
We don't need to set zeroStateConfig as Yaml string and later parse it to ZeroStateConfig structure.

### DIFF
--- a/nil/tests/async_await/async_await_test.go
+++ b/nil/tests/async_await/async_await_test.go
@@ -5,7 +5,6 @@ import (
 	"testing"
 
 	"github.com/NilFoundation/nil/nil/common"
-	"github.com/NilFoundation/nil/nil/common/hexutil"
 	"github.com/NilFoundation/nil/nil/internal/abi"
 	"github.com/NilFoundation/nil/nil/internal/contracts"
 	"github.com/NilFoundation/nil/nil/internal/execution"
@@ -27,7 +26,6 @@ type SuiteAsyncAwait struct {
 	counterAddress1 types.Address
 	abiTest         *abi.ABI
 	abiCounter      *abi.ABI
-	zerostateCfg    string
 	accounts        []types.Address
 }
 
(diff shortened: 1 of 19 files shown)


nil/tests/async_await/async_await_test.go
```diff
@@ -54,46 +53,21 @@ func (s *SuiteAsyncAwait) SetupSuite() {
...
 	nShards := uint32(3)
 	port := 10425
 
-	zeroState, err := execution.ParseZeroStateConfig(s.zerostateCfg)
+	smartAccountValue, err := types.NewValueFromDecimal("100000000000000")
 	s.Require().NoError(err)
-	zeroState.MainPublicKey = execution.MainPublicKey
+	zeroState := &execution.ZeroStateConfig{
+		Contracts: []*execution.ContractDescr{
+			{Name: "MainSmartAccount", Contract: "SmartAccount", Address: &types.MainSmartAccountAddress, Value: smartAccountValue, CtorArgs: []interface{}{hexutil.Encode(execution.MainPublicKey)}},
```
olegrok, line 63:
```suggestion
			{Name: "MainSmartAccount", Contract: "SmartAccount", Address: &types.MainSmartAccountAddress, Value: smartAccountValue, CtorArgs: []any{hexutil.Encode(execution.MainPublicKey)}},
```

olegrok, line 63: It should be possible to avoid `hexutil.Encode` here

nil/tests/async_await/async_await_test.go
```diff
@@ -54,46 +53,21 @@ func (s *SuiteAsyncAwait) SetupSuite() {
...
+			{Name: "Test0", Contract: "tests/RequestResponseTest", Address: &s.testAddress0, Value: smartAccountValue},
+			{Name: "Test1", Contract: "tests/RequestResponseTest", Address: &s.testAddress1, Value: types.Value0},
+			{Name: "Counter0", Contract: "tests/Counter", Address: &s.counterAddress0, Value: smartAccountValue},
+			{Name: "Counter1", Contract: "tests/Counter", Address: &s.counterAddress1, Value: smartAccountValue},
+		},
+		MainPublicKey: execution.MainPublicKey,
```
olegrok, line 69: Do you still need it? You pass it as a constructor argument

nil/tests/cli_call/cli_call_test.go
```diff
@@ -33,21 +32,14 @@ func (s *SuiteCliTestCall) SetupSuite() {
...
-  contract: tests/Test
-`, s.testAddress.Hex())
-
-	s.zeroState, err = execution.ParseZeroStateConfig(zerostateCfg)
-	s.Require().NoError(err)
-	s.zeroState.MainPublicKey = execution.MainPublicKey
 }
 
 func (s *SuiteCliTestCall) SetupTest() {
+	s.zeroState = &execution.ZeroStateConfig{
```
olegrok, line 38: You probably don't need a struct member anymore
```suggestion
	zeroState := &execution.ZeroStateConfig{
```

nil/tests/async_await/async_await_test.go
Zerg1996: I totally get rid of this field in config.

olegrok (approved):

PR #395, opened 2025-02-24
[collate] prevent "Transaction conflict" error because of syncer/consensus race

olegrok:
The main idea is to make validator as an entry point for proposal insertion.
We insert proposal in two cases: consensus and block replay.

Sometimes we had an error "Transaction conflict" caused by race between
that two modules. So it seems reasonable to add a lock here to
avoid confusing error messages and execution transaction two times.

### DIFF
--- a/nil/internal/collate/syncer.go
+++ b/nil/internal/collate/syncer.go
@@ -7,7 +7,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/NilFoundation/nil/nil/common"
 	"github.com/NilFoundation/nil/nil/common/assert"
 	"github.com/NilFoundation/nil/nil/common/check"
 	"github.com/NilFoundation/nil/nil/common/logging"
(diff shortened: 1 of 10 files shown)


nil/internal/collate/syncer.go
```diff
@@ -69,31 +60,14 @@ func NewSyncer(cfg *SyncerConfig, db db.DB, networkManager *network.Manager) (*S
...
-	}
-	if err == nil {
-		return block, hash, err
-	}
-	return nil, common.EmptyHash, nil
-}
-
 func (s *Syncer) shardIsEmpty(ctx context.Context) (bool, error) {
-	block, _, err := s.readLastBlock(ctx)
+	block, _, err := s.validator.GetLastBlock(ctx)
```
knazarov, line 70: I know this has worked like this for a while, but it may be a source of inconsistency if we read such things outside of transactions. You can read the last block to be one thing initially and then at some point down the line it will be different.

Zerg1996 (approved):
Thanks for PR! But it'll be awesome if you explain why we need `GetLastBlock` with mutex inside.

nil/internal/collate/validator.go
```diff
@@ -2,26 +2,91 @@ package collate
...
+	if err != nil && !errors.Is(err, db.ErrKeyNotFound) {
+		return nil, common.EmptyHash, err
+	}
+	if err == nil {
+		return block, hash, err
+	}
+	return nil, common.EmptyHash, nil
+}
+
+func (s *Validator) GetLastBlock(ctx context.Context) (*types.Block, common.Hash, error) {
```
knazarov, line 77: I'm not completely sure that putting fine-grained locks into individual functions is a good idea. Can we do something like a lock around all operations with the validator? It's very easy to miss funny edge cases otherwise.

knazarov (approved):
I'm generally OK with the approach, because it's better than what we have now (spamming conflicts to the logs). However, it would be good to make the locking around validator better (taking one lock for example or something like that). But do let me know if you disagree.

knazarov:
> Thanks for PR! But it'll be awesome if you explain why we need `GetLastBlock` with mutex inside.

I guess this is because this function can be called from multiple threads and we just want to return consistent results. This is why we take an "RW" lock.

nil/internal/collate/syncer.go
olegrok: This function is called once on instance startup. Here we check that data is empty. This function expects some binary response (we have some blocks in db or not).
Below we also don't care about it because we need block.Id for log message.

nil/internal/collate/validator.go
olegrok: I'll think about it. Probably all public functions should have lock management and call only one `<functionName>Unlocked` function inside.

You are basically right but I'm a bit confused by choosing correct approach here.

PR #397, opened 2025-02-24
Introduce external transactions ordering

shermike:
This PR introduces ordering of transactions in txpool. It is preliminary version, final one should be much more sophisticated, probably similar to Erigon's version.
Currently, it orders transactions by their effective priority fee. This applies only to transactions sent to different contracts. Transactions on the same contract are ordered using seqno. It doesn't take into account gaps of seqno and current seqno of the account.


nil/internal/collate/proposer.go
```diff
@@ -432,3 +433,73 @@ func (p *proposer) handleTransactionsFromNeighbors(tx db.RoTx) error {
 	p.proposal.CollatorState = state
 	return nil
 }
+
+type TransactionWithFee struct {
+	*types.Transaction
+	EffectivePriorityFee types.Value
+}
+
+func (p *proposer) sortExternalTransactions(poolTxns []*types.Transaction) []*types.Transaction {
```
olegrok, line 442: Shouldn't we do that right on txpool stage?

nil/internal/execution/state.go
```diff
@@ -28,7 +28,7 @@ import (
 
 var logger = logging.NewLogger("execution")
 
-const TraceBlocksEnabled = false
+const TraceBlocksEnabled = true
```
olegrok, line 31: debug?

nil/internal/collate/proposer.go
```diff
@@ -432,3 +433,73 @@ func (p *proposer) handleTransactionsFromNeighbors(tx db.RoTx) error {
...
+			heap.Push(&byPriorityFee, accTxs[0])
+			bySeqno[best.To] = accTxs[1:]
+		}
+		resTxns = append(resTxns, best.Transaction)
+	}
+
+	return resTxns
+}
+
+type TxByPriorityFee []*TransactionWithFee
```
olegrok, line 482: You can try to use https://github.com/ethereum/go-ethereum/blob/7d8aca95d28c4e8560a657fd1ff7852ad4eee72c/common/prque/prque.go

shermike: Hmm, I don't see how it can be better solution here. It operates with `item` struct which require priority to be int64. Maybe I miss something?

nil/internal/collate/proposer.go
shermike: I think Proposer is a better place for ordering transactions, since txpool is quite low level component which is responsible only for storing transactions and exchanging them with peers.

nil/internal/collate/proposer.go
olegrok: I just suggested to use some existing solution. If you think that it's not applicable for our case then let's stay them as is.

nil/internal/collate/proposer.go
olegrok: You can take a look on txpool in geth/erigon they do such ordering.

To do it on proposer layer properly you need to extract all transactions (thousands/tens of thousands of them) and then sort all this batch.
Or you can peek top N from txpool that already added all messages to sorted heap in a appropriate way

shermike: Hmm, sounds reasonable, will try to move it to the pool.

olegrok: Initially I implemented our txpool based on erion's one.

But we didn't have any priorities at that time so I just put all transactions into queue instead of heap. Now it seems a time to change it.

nil/services/txnpool/meta_txn.go
```diff
@@ -0,0 +1,34 @@
...
+
+import (
+	"github.com/NilFoundation/nil/nil/common"
+	"github.com/NilFoundation/nil/nil/internal/execution"
+	"github.com/NilFoundation/nil/nil/internal/types"
+)
+
+type metaTxn struct {
+	*types.TxnWithHash
+	hash                 common.Hash
```
olegrok, line 11: But `TxnWithHash` already includes hash

nil/services/txnpool/txnpool.go
```diff
@@ -375,16 +424,25 @@ func (p *TxnPool) Peek(ctx context.Context, n int) ([]*types.TxnWithHash, error)
...
 	}
+
 	return res, nil
 }
 
 func (p *TxnPool) TransactionCount() int {
-	p.lock.Lock()
-	defer p.lock.Unlock()
-	return p.queue.Size()
+	txns, err := p.Peek(context.Background(), 1000000)
```
olegrok, line 443: Let's drop context argument if Peek doesn't use it

nil/services/txnpool/txnpool.go
```diff
@@ -371,20 +420,29 @@ func (p *TxnPool) removeCommitted(bySeqno *ByReceiverAndSeqno, txns []*types.Tra
...
 	}
+
 	return res, nil
 }
 
 func (p *TxnPool) TransactionCount() int {
-	p.lock.Lock()
-	defer p.lock.Unlock()
-	return p.queue.Size()
+	txns, err := p.Peek(1000000)
```
olegrok, line 443: That looks strange. Is it possible to implement fair count method?

olegrok: > Btw TransactionCount() is used only in tests

Let's move it to testaide.go then. If it's useful but only for tests

shermike: Moved to the test file.

nil/internal/execution/fee.go
```diff
@@ -67,11 +83,21 @@ func calculateBaseFee(baseFeePrevious types.Value, gasUsedPrevious types.Gas) ty
 	baseFeePreviousFixed := decimal.NewFromBigInt(baseFeePrevious.Int().ToBig(), 0)
 	newFee := baseFeePreviousFixed.Mul(adjustmentFactor.Mul(normalized).Add(decimal.New(1, 0)))
 
+	var resultBaseFee types.Value
```
olegrok, line 86: Then you don't need else branch
```suggestion
	resultBaseFee := types.DefaultGasPrice
```

nil/internal/execution/fee.go
```diff
@@ -53,7 +66,10 @@ func init() {
 	}
 }
 
-func calculateBaseFee(baseFeePrevious types.Value, gasUsedPrevious types.Gas) types.Value {
+func (m *MainFeeCalculator) CalculateBaseFee(prevBlock *types.Block) (types.Value, error) {
```
olegrok, line 69: Do we really need `error` return value here?

nil/internal/execution/state.go
```diff
@@ -1544,7 +1554,14 @@ func (es *ExecutionState) GetShardID() types.ShardId {
 	return es.ShardId
 }
 
-func (es *ExecutionState) CallVerifyExternal(transaction *types.Transaction, account *AccountState) *ExecutionResult {
+func (es *ExecutionState) CallVerifyExternal(transaction *types.Transaction) *ExecutionResult {
+	account, err := es.GetAccount(transaction.To)
+	check.PanicIfErr(err)
```
olegrok, line 1559: Can't we send an external message to some non-existent account?

nil/internal/execution/fee.go
shermike: Not anymore, it seems.

olegrok (approved):
Thanks!

Issue #392, opened 2025-02-24
Compiling "Handling external transactions" results in error


ukorvl:
If a user tries to compile `"Handling external transactions"` example it results in solidity compiler error.
![Image](https://github.com/user-attachments/assets/3e9fa47e-a00d-4c1f-8906-4165062a7611)

ukorvl:
Done

PR #399, opened 2025-02-25
Fixed a bug in Handling External Transactions example contract

gitshreevatsa:
This PR fixes issue [392](https://github.com/NilFoundation/nil/issues/392). It changes the state mutability of a function in the contract which would throw compiling error otherwise. This PR also fixes a bug where bytes input types in constructor were incorrectly processed, leading to unexpected behaviour which is it would result in errors such as `t.replace` is not a function . Previously, the bytes case returned true instead of the actual input value due to improper type coercion.


ukorvl (approved):
LGTM

Issue #223, opened 2025-02-07
Improve RPC Endpoint Selection Flow During Wallet Creation


0xAleksaOpacic:
## Description:
Currently, users must manually enter an RPC endpoint when creating a wallet, requiring them to copy it from external sources (e.g., TG bot). This disrupts the UX and makes the onboarding process harder.

We need a new approach where users don’t have to manually input an RPC endpoint, but instead get automatically assigned one. This requires a new flow, potentially involving a “Connect with X” button that assigns an endpoint based on the user’s credentials (e.g., username)

## Acceptance Criteria:
1. Implement an automated method for assigning RPC endpoints
2. Ensure the flow is user-friendly and does not introduce complexity
3. Sync with Design and Product teams to finalize UX changes

## Technical Notes:
1. We have a backend which already manages a set of RPC endpoints for TG bot interactions, evaluate integrating a similar mechanism in the extension
2. Consider a “Connect with X” button approach to assign endpoints dynamically

![Image](https://github.com/user-attachments/assets/12d13643-b30c-410b-b016-0a2ae64cb619)

PR #400, opened 2025-02-25
Wallet: great renaming and set default endpoint

Rolaman:
Wallet: great renaming and set default endpoint

- renamed currency to token
- set default endpoint

Fixed https://github.com/NilFoundation/nil/issues/223


wallet-extension/.env
```diff
@@ -7,4 +7,5 @@ VITE_ETH_ADDRESS=0x0001111111111111111111111111111111111112
 VITE_USDT_ADDRESS=0x0001111111111111111111111111111111111113
 VITE_BTC_ADDRESS=0x0001111111111111111111111111111111111114
 VITE_NUMBER_SHARDS=4
-VITE_NIL_EXPLORER=https://explore.nil.foundation/
\ No newline at end of file
+VITE_NIL_EXPLORER=https://explore.nil.foundation/
+VITE_RPC_URL=https://api.devnet.nil.foundation/api/wallet/05575405435085249967847709316857
```
knazarov, line 11: Is this RPC token? If so - you need to remove it and recreate.

Rolaman: agree, removed. also will create a PR for infra

knazarov (approved):

PR #402, opened 2025-02-25
Use eth_getShardIdList to get shards amount in playground

ukorvl:
This diff adds a new methods to `niljs` client  - `getShardIdList()` and `getNumShards()`. `getNumShards()` returns the amount if shards. We decided to migrate to usage of this rpc method in playground to decrease connectedness between playground and explorer backend.

In addition this diff implements `getNumShards()` on the side of rpc.

It adds `openrpc.json` to `.gitignore` because we don't want to commit auto-generated file accidentally.

### DIFF
--- a/explorer_frontend/src/features/contracts/init.ts
+++ b/explorer_frontend/src/features/contracts/init.ts
@@ -5,9 +5,8 @@ import { persist } from "effector-storage/local";
 import { debug } from "patronum";
 import type { App } from "../../types";
 import { $rpcUrl, $smartAccount } from "../account-connector/model";
-import { $solidityVersion, compileCodeFx } from "../code/model";
+import { $solidityVersion, compileCodeFx, loadedPlaygroundPage } from "../code/model";
 import { $cometaService } from "../cometa/model";
-import { $shardsAmount } from "../shards/models/model";
 import { getTokenAddressBySymbol } from "../tokens";
 import {
   $activeApp,
@@ -64,6 +63,7 @@ import {
   validateSmartContractAddressFx,
 } from "./models/base";
 import { exportApp, exportAppFx } from "./models/exportApp";
+import { $shardsAmount, getShardsAmountFx } from "./models/shardsAmount";
 
 compileCodeFx.doneData.watch(console.log);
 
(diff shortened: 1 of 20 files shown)


explorer_frontend/src/features/contracts/init.ts
```diff
@@ -651,3 +651,16 @@ sample({
   filter: $activeAppWithState.map((app) => !!app && !app.address),
   target: triggerShardIdValidation,
 });
+
+$shardsAmount.on(setShardsAmount, (_, shardsAmount) => shardsAmount);
```
KlonD90, line 655: it's better here to subscribe on effectResult. Event should be used more for user interaction side.

ukorvl: Done!

explorer_frontend/src/features/contracts/init.ts
```diff
@@ -651,3 +651,14 @@ sample({
...
   target: triggerShardIdValidation,
 });
+
+sample({
+  clock: getShardsAmountFx.doneData,
+  target: $shardsAmount
+});
+
+sample({
+  clock: merge([loadedPlaygroundPage, $rpcUrl.updates]),
```
KlonD90, line 661: you could leave $rpcUrl but add a filter to it

KlonD90 (approved):

knazarov:
Why do we need to get shard list, where we can just get shard count? Aren't all shards 1 to <num shards>?

nil/services/rpc/rawapi/local_system.go
```diff
@@ -49,3 +49,26 @@ func (api *LocalShardApi) GetShardIdList(ctx context.Context) ([]types.ShardId,
...
+	block, _, err := db.ReadLastBlock(tx, types.MainShardId)
+	if err != nil {
+		return 0, err
+	}
+
+	treeShards := execution.NewDbShardBlocksTrieReader(tx, types.MainShardId, block.Id)
+	treeShards.SetRootHash(block.ChildBlocksRootHash)
+	keys, err := treeShards.Keys()
+
+	return uint64(len(keys)), err
```
olegrok, line 73:
```
shards, err := api.GetShardIdList(ctx)
if err != nil {
    return 0, err
}
return len(shards) + 1, nil
```

ukorvl: Done

olegrok (approved):

PR #408, opened 2025-02-25
Fix conversion of amount during top-up

gitshreevatsa:
Currently, the conversion of amount for topping up is not parsed correctly, as NIL token and other tokens have different decimals. This PR fixes the conversion of the amount to bigint and number respectively for topping up of NIL and other tokens.

### DIFF
--- a/explorer_frontend/src/features/account-connector/init.ts
+++ b/explorer_frontend/src/features/account-connector/init.ts
@@ -100,7 +100,7 @@ createSmartAccountFx.use(async ({ privateKey, rpcUrl }) => {
       {
         smartAccountAddress: smartAccount.address,
         faucetAddress: faucets.NIL,
-        amount: BigInt(1e18),
+        amount: convertEthToWei(0.1),
       },
       client,
     );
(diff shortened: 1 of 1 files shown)


explorer_frontend/src/features/account-connector/init.ts
```diff
@@ -265,13 +265,20 @@ topupSmartAccountTokenFx.use(async ({ smartAccount, topupInput, faucets, rpcUrl
...
+  } else {
+    topUpAmount = Number(amount);
+  }
+
   const txHash = await faucetClient.topUpAndWaitUntilCompletion(
     {
       smartAccountAddress: smartAccount.address,
       faucetAddress: tokenFaucetAddress,
-      amount: Number(amount),
+      amount: topUpAmount,
```
ukorvl, line 281:
```suggestion
      amount: tokenFaucetAddress === faucets.NIL ? convertEthToWei(Number(amount)) : BigInt(amount)
```

ukorvl (approved):

PR #409, opened 2025-02-25
[docs] removal of hardcoded tokens

khannanov-nil:
This diff removes hardcoded RPC URL from docs' tests. It was there previously when the repo was still closed but now it's been exposed to the public.


knazarov (approved):

knazarov:
NB: these tokens were rotated and are no longer active.

PR #411, opened 2025-02-25
Fix ethers.js v6 compatibility and logging cleanup

Hopium21:
create-wallet-with-funding.ts

Old: console.log("Transaction Mined:", receipt?.hash);
To: console.log("Transaction Mined:", receipt.hash);
Reason: tx.wait() always returns a valid receipt with a hash, so optional chaining (?.) is unnecessary.

deploy-nilchain.ts

Old:const valueInHex = ethers.toQuantity(ethers.parseEther("100"));
To:const valueInHex = ethers.hexlify(ethers.parseEther("100"));
Reason: ethers.toQuantity() is deprecated in ethers.js v6. ethers.hexlify() is the correct alternative to ensure compatibility.

### DIFF
--- a/l1-contracts/create-wallet-with-funding.ts
+++ b/l1-contracts/create-wallet-with-funding.ts
@@ -53,7 +53,7 @@ async function createAndUseWallet() {
 
     // Step 4: Wait for the transaction to be mined
     const receipt = await tx.wait();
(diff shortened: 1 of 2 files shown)


Hopium21:
Hi @AndreyMlashkin 

Please take a look at the open pull request #411. It should be useful and ready for review.

Thanks!

PR #413, opened 2025-02-25
fixed missing semicolon in box-shadow styles.module.css

aso20455:
Added a semicolon after the box-shadow property to properly finalize the CSS rule


knazarov:
For nitpicks, please open an issue instead. We encourage contributions, but primarily for functional changes.

Issue #97, opened 2025-01-31
Add a metric for number of poseidon calls in a block


knazarov:
This is needed to calculate how constrained are we on poseidon. Poseidon is said to be 30k hashes/second in terms of throughput.

We can (and will) optimize the hashes and delay them as much as possible to block finalization time, so that they can be evaluated in parallel threads. But still, it's good to get a baseline.

dmtrskv:
Exact measuring requires either introduction of tracing (and thus contexts) to the basic methods, or playing tricks.
The overall load from the hashing can be seen in profiles.

Hashing was reduced to the minimum necessary amount (though no control over it).
Also, poseidon is suspected to be replaced.

PR #414, opened 2025-02-25
[optimization] Reduce block and txn hashing via reuse

dmtrskv:
Reduced the calls to the expensive `Hash()` method of `Block` and `Transaction` to more or less necessary amount.

While looking at ways to implement https://github.com/NilFoundation/nil/issues/97, I decided to optimize the code, since there aren't many places where blocks and txns are hashed.
Now transaction hashes are calculated at the following places in validators:
1. Pool: all added txns.
2. Proposer: all incoming internal txns (not forwarded; external txn hashes are taken from the pool).
3. Proposal validation: block execution.
4. Proposal commit: block execution.
5. Block execution: all incoming and outgoing (generated during execution) txns.

Changes made:
1. Proposer fetches txns from the pool along with their hashes (`MetaTxn`).
2. Pool discards txns by hash.
3. `BlockGenerationResult` obtained in and out txn hashes. Users of this struct (e.g. `BlocksTracer`) now don't calculate these hashes themselves. For that, `OutboundTransaction` obtained a hash field. Also, fixed tests that produced their own mock `BlockGenerationResult` without txns.
4. Postropessing part that writes `txn hash -> (block hash, txn num)` map uses transactions from `BlockGenerationResult` instead of parsing the trie.
5. Unnecessary hashing in logs: removed or replaced with already calculated hash.
6. Each n-th block is reported at info level by a syncher along with its hash. This is done so that archive nodes show some activity in info logs, but also because logging of the block's hash was removed from the debug log.

### DIFF
--- a/nil/services/txnpool/txnpool.go
+++ b/nil/services/txnpool/txnpool.go
@@ -16,27 +16,25 @@ import (
 
 type Pool interface {
 	Add(ctx context.Context, txns ...*types.Transaction) ([]DiscardReason, error)
-	Discard(ctx context.Context, txns []*types.Transaction, reason DiscardReason) error
+	Discard(ctx context.Context, txns []common.Hash, reason DiscardReason) error
 	OnCommitted(ctx context.Context, committed []*types.Transaction) error
 	// IdHashKnown check whether transaction with given Id hash is known to the pool
 	IdHashKnown(hash common.Hash) (bool, error)
 	Started() bool
 
-	Peek(ctx context.Context, n int) ([]*types.Transaction, error)
+	Peek(ctx context.Context, n int) ([]*types.TxnWithHash, error)
 	SeqnoToAddress(addr types.Address) (seqno types.Seqno, inPool bool)
 	TransactionCount() int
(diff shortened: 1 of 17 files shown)


nil/services/txnpool/txnpool.go
```diff
@@ -16,30 +16,34 @@ import (
...
 
-	Peek(ctx context.Context, n int) ([]*types.Transaction, error)
+	Peek(ctx context.Context, n int) ([]*MetaTxn, error)
 	SeqnoToAddress(addr types.Address) (seqno types.Seqno, inPool bool)
 	TransactionCount() int
 	Get(hash common.Hash) (*types.Transaction, error)
 }
 
-type metaTxn struct {
+type MetaTxn struct {
```
olegrok, line 31: Initially I considered this as a pure internal txpool type.
We probably can add more system fields in future that will be handled. Maybe let's add something like `TxWithHash` to `types` and inline it here?

I'd avoid leaking `MetaTxn` type outside the pool.

olegrok (approved):

PR #415, opened 2025-02-25
Remove redundant padding declaration in `.socialButton`

GooseMatrix:
The `padding` property in `.socialButton` is duplicated. The first declaration (`padding: 5px;`) is immediately overridden by the second (`padding: 20px;`), making it redundant. Removing the unnecessary declaration improves code readability, reduces confusion, and simplifies future maintenance.


knazarov:
For nitpicks, please open an issue instead. We encourage contributions, but primarily for functional changes.

PR #417, opened 2025-02-26
[explorer] implemented frontend for interactive tutorials

khannanov-nil:
This diff adds the first version of the UI for interactive tutorials. While this UI is just a placeholder, this PR also changes some business logic on the side of the `Code` feature and adds a `Tutorial` feature both of which can be reused when a different UI is added later on. Since the tutorial backend is disabled anyway, we can safely merge this.


explorer_frontend/src/features/code/model.ts
```diff
@@ -51,9 +55,17 @@ export const $shareCodeSnippetError = codeDomain.createStore<boolean>(false);
 export const setCodeSnippetEvent = codeDomain.createEvent();
 export const fetchCodeSnippetEvent = codeDomain.createEvent<string>();
 
+export const fetchTutorialEvent = codeDomain.createEvent<Tutorial>();
```
KlonD90, line 58: should move the to tutorial page model and work with them in proper init too.

explorer_frontend/src/features/code/model.ts
```diff
@@ -62,7 +74,25 @@ fetchCodeSnippetFx.use((hash) => {
...
+
+fetchAllTutorialsFx.use(() => {
+  return fetchAllTutorials();
+});
+
+export const loadedPlaygroundPage = codeDomain.createEvent<boolean>();
+
+export const loadedTutorialPage = codeDomain.createEvent<boolean>();
+
+export const userClickOnLogButton = codeDomain.createEvent();
```
KlonD90, line 89: we could reduce user from here

explorer_frontend/src/features/tutorial/model.ts
```diff
@@ -0,0 +1,23 @@
...
+export const tutorialDomain = createDomain("tutorial");
+
+export const $tutorial = tutorialDomain.createStore<Tutorial>({
+  id: "",
+  text: "",
+  contracts: "",
+  stage: 0,
+});
+export const $compiledTutorialContracts = tutorialDomain.createStore<App[]>([]);
+export const $tutorialText = $tutorial.map((tutorial) => (tutorial ? tutorial.text : ""));
```
KlonD90, line 20: do you really need this projection? it looks like tutorial is pretty immutable during page so you could just import tutorial itself and print tutorial.text

explorer_frontend/src/features/code/model.ts
```diff
@@ -62,7 +74,25 @@ fetchCodeSnippetFx.use((hash) => {
...
-export const loadedPlaygroundPage = codeDomain.createEvent();
+fetchTutorialFx.use((stage) => {
+  return fetchTutorial(stage);
+});
+
+fetchAllTutorialsFx.use(() => {
+  return fetchAllTutorials();
+});
+
+export const loadedPlaygroundPage = codeDomain.createEvent<boolean>();
```
KlonD90, line 85: It looks like we don't need boolean as argument here

explorer_frontend/src/pages/playground/init.ts
```diff
@@ -1,7 +1,14 @@
...
-import { $activeComponent, setActiveComponent } from "./model";
+import {
+  userClickOnBackButton,
+  userClickOnContractsButton,
+  userClickOnLogButton,
+} from "../../features/code/model";
+import { $activeComponent, LayoutComponent } from "./model";
 
-$activeComponent.on(setActiveComponent, (_, payload) => payload);
+$activeComponent.on(userClickOnLogButton, (_) => LayoutComponent.Logs);
```
KlonD90, line 9: it looks like you could write `() => LayoutComponent.Logs`. It was used (_, payload) because we skipped first argument and used only second one. But in this case we don't use at all arguments

explorer_frontend/src/features/logs/components/Logs.tsx
```diff
@@ -70,7 +70,7 @@ export const Logs = () => {
             }}
             kind={BUTTON_KIND.secondary}
             size={BUTTON_SIZE.compact}
-            onClick={() => setActiveComponent(LayoutComponent.Code)}
+            onClick={() => userClickOnBackButton()}
```
KlonD90, line 73: That event should be inside of proper model not import from code.

explorer_frontend/src/pages/tutorials/TutorialMobileLayout.tsx
```diff
@@ -0,0 +1,34 @@
...
+import { ContractsContainer } from "../../features/contracts";
+import { Logs } from "../../features/logs/components/Logs";
+import { TutorialText } from "../../features/tutorial/TutorialText";
+import {
+  $activeComponentTutorial,
+  TutorialLayoutComponent,
+  setActiveComponentTutorial,
+} from "./model";
+
+const featureMap = new Map();
```
KlonD90, line 13: it should be typed

explorer_backend/services/tutorials_db.ts
```diff
@@ -45,16 +46,20 @@ export const setTutorial = (text: string, contracts: string, stage: number): Tut
...
 export const generateTutorials = async (): Promise<void> => {
-  const tutorialsSpec = JSON.parse(await fs.readFile("../tutorials/spec.json", "utf-8"));
+  const tutorialsSpec = JSON.parse(
+    await fs.readFile(path.resolve("./tutorials/spec.json"), "utf-8"),
+  );
 
   for (const tutorial of tutorialsSpec) {
-    const text = await fs.readFile(`../tutorials/${tutorial.text}`, "utf-8");
-    const contracts = await fs.readFile(`../tutorials/${tutorial.contracts}`, "utf-8");
+    console.log(tutorial);
```
KlonD90, line 60: it's still broken please remove backend part or make it workable

explorer_frontend/src/pages/playground/init.ts
khannanov-nil: got it! fixed

explorer_frontend/src/pages/tutorials/TutorialMobileLayout.tsx
khannanov-nil: thanks, fixed!

explorer_frontend/src/features/tutorial/model.ts
khannanov-nil: got it, removed

explorer_frontend/src/features/code/model.ts
khannanov-nil: sure, removed!

explorer_frontend/src/features/code/model.ts
khannanov-nil: got it, removed!

explorer_frontend/src/features/code/model.ts
khannanov-nil: got it, fixed

explorer_frontend/src/types.ts
```diff
@@ -6,3 +6,10 @@ export type App = {
   sourcecode: string;
   abi: Abi;
 };
+
+export type Tutorial = {
```
ukorvl, line 10: Move this type to tutorials feature

explorer_frontend/src/features/code/Code.tsx
```diff
@@ -65,6 +70,7 @@ export const Code = () => {
 
   const noCode = code.trim().length === 0;
   const btnContent = useCompileButton();
+  console.log(isTutorial);
```
ukorvl, line 73: debug?

khannanov-nil: yep, removed

explorer_frontend/src/features/code/Code.tsx
```diff
@@ -27,14 +30,16 @@ const MemoizedCodeToolbar = memo(CodeToolbar);
...
-  ]);
+  const [code, isDownloading, errors, fetchingCodeSnippet, compiling, warnings, isTutorial] =
+    useUnit([
+      $code,
+      fetchSolidityCompiler.pending,
+      $error,
+      fetchCodeSnippetFx.pending,
+      compileCodeFx.pending,
+      $warnings,
+      isTutorialPage,
```
KlonD90, line 41: we don't need isTutorialPage in current implenetation at all

explorer_frontend/src/features/code/Code.tsx
```diff
@@ -243,10 +250,31 @@ export const Code = () => {
...
               kind={BUTTON_KIND.secondary}
               size={BUTTON_SIZE.large}
-              onClick={() => setActiveComponent(LayoutComponent.Contracts)}
+              onClick={() => {
+                сlickOnContractsButton();
+              }}
             >
               Contracts
             </Button>
+            {isTutorial && (
```
KlonD90, line 259: if you want to add this thing add as prop and use on TutorialPage level

khannanov-nil: got it, done!

explorer_frontend/src/features/code/Code.tsx
khannanov-nil: got it, fixed!

explorer_frontend/src/features/tutorial/spec.ts
```diff
@@ -0,0 +1,14 @@
+async function loadTutorials() {
+  const testTutorial = await import("./assets/testTutorial.md?raw");
```
KlonD90, line 2: if you make two imports async it's better to make it in parallel

khannanov-nil: got it, fixed

explorer_frontend/src/features/code/Code.tsx
```diff
@@ -8,24 +8,29 @@ import {
...
-import { LayoutComponent, setActiveComponent } from "../../pages/playground/model";
 import { fetchSolidityCompiler } from "../../services/compiler";
 import { getMobileStyles } from "../../styleHelpers";
 import { useMobile } from "../shared";
 import { SolidityCodeField } from "../shared/components/SolidityCodeField";
 import { CodeToolbar } from "./code-toolbar/CodeToolbar";
 import { useCompileButton } from "./hooks/useCompileButton";
 
+interface CodeProps {
+  additionalProp: JSX.Element | null;
```
KlonD90, line 28: that's kinda not helping name of variable. i propose to use name as extraMobileButtons. also it looks like you want to this will be a optional so add to name of props in the end `?` that will be optional parameter. Also JSX.Element is cool thing but if you want to be this a null in some case it's better to be ReactNode

explorer_frontend/src/features/code/model.ts
```diff
@@ -64,6 +66,16 @@ fetchCodeSnippetFx.use((hash) => {
...
 
+export const loadedTutorialPage = codeDomain.createEvent();
+
+export const сlickOnLogButton = codeDomain.createEvent();
+
+export const сlickOnContractsButton = codeDomain.createEvent();
+
+export const сlickOnBackButton = codeDomain.createEvent();
+
+export const сlickOnTutorialButton = codeDomain.createEvent();
```
KlonD90, line 77: it's better to move out this button to TutorialPage feature because it would be injected by this feature

explorer_frontend/src/features/tutorial/model.ts
```diff
@@ -0,0 +1,34 @@
...
+
+export const fetchTutorialEvent = tutorialDomain.createEvent<Tutorial>();
+export const fetchAllTutorialsEvent = tutorialDomain.createEvent<Tutorial[]>();
+
+export const fetchTutorialFx = tutorialDomain.createEffect<string, Tutorial, string>();
+export const fetchAllTutorialsFx = tutorialDomain.createEffect<void, Tutorial[], string>();
+
+fetchTutorialFx.use(async (stage) => {
+  const tutorials = await loadTutorials();
+  return tutorials.find((tutorial) => tutorial.stage === Number(stage))!;
```
KlonD90, line 28: but it's actually could be undefined still so i don't understand why you here use force type

KlonD90 (changes requested):

explorer_frontend/src/features/tutorial/model.ts
khannanov-nil: got it, handling this explicitly now

explorer_frontend/src/features/code/model.ts
khannanov-nil: sure, moved

explorer_frontend/src/features/code/Code.tsx
khannanov-nil: sure, renamed

KlonD90 (approved):

PR #422, opened 2025-02-26
Add rollback status to the block and SudoKey to the config.

avm:


olegrok:
Please add description and probably tests

Issue #424, opened 2025-02-26
Incorrect behavior of playground token inputs with multiple payable functions


ukorvl:
https://github.com/user-attachments/assets/ba261eea-ca7b-4a05-ae14-1307ceb80646

In current playground version we have an issue with tokens input. The issue is that they are "shared" between all smart contract payable functions. When we change input in one function - it is changed in all other functions.

PR #430, opened 2025-02-27
Add call args playground validation and fix multiple token inputs issue

### LINKED ISSUE #23: Sandbox client-side fields validation
Validation on the client side should be added to some fields of sandbox forms.
Mostly it relates to contract interaction form (we need a validation based on abi of the contract) but there can be som other places - check the [figma template](https://www.figma.com/design/1Wfrv3uMit3MRcFatcYEaZ/%F0%9F%92%A5-Sandbox?t=akhpRlhaDzZ2aF8p-0) 

Image

Image
For shard id there should be also an api to get the amount of the shards, implemented on backend side. It is not difficult - just add a query to clickhouse, which will return the amount of shards.

For errors text color use red200 color everywhere.

ukorvl:
fixes #23 

This diff fixes several issues related to playground.

Firs off, it fixes the problem with multiple token inputs, described in #424 by extending object, which stores the values and adding 1 more depth level to it for each of the smart contract functions.

The second issue it fixes - lack of client side smart contract function input fields validation. The diff introduces new effector domain for validation of call params, including validation of tuples.

Also, it fixes the display of result of a smart contract function call. Sometimes result can be an array of values, that are returned by function, and we want to display them using comma separator with additional space `, `.

It fixes the problem with extra spaces at the beginning of rpc url, like ` https://`, which were considered invalid, by trimming the value before validation.

Additionally it moves some modules, which is not a significant change.

https://github.com/user-attachments/assets/5a4f4d3e-729f-4b7c-a70e-b9c409860a8a


explorer_frontend/src/features/account-connector/init.ts
```diff
@@ -140,7 +140,7 @@ createSmartAccountFx.use(async ({ privateKey, rpcUrl }) => {
         {
           smartAccountAddress: smartAccount.address,
           faucetAddress: tokenFaucetAddress,
-          amount: 10,
+          amount: 10n,
```
KlonD90, line 143: that's kinda small? if it's wei it's just 0.00...0001

ukorvl: hmm, let me check.

ukorvl:
Converting this to draft because will be able to merge only after vacation.

KlonD90 (approved):

explorer_frontend/src/features/account-connector/init.ts
ukorvl: Because it is not `wei`, it is token without decimal, so it will add exactly 10 tokens.

PR #434, opened 2025-02-28
[exporter] Fix catch-up

dmtrskv:
Partial rewrite of the exporter adding idempotence to the sync which was not present, thus the previous version failed to start from scratch and sync everything succesfully.

1. Added versioning: the clickhouse db is wiped (if `--allow-db-clear`) if main shard genesis block hashes differ in blockchain and clickhouse.
2. Added `"-"` as `ch` tag value meaning that the field should be skipped. Removed signature from clickhouse serialization, because it didn't sync to clickhouse (to be fixed later, or maybe it's better this way).
3. Separated `exporter` and its `Cfg`.
4. Mostly rewritten fetchers in exporter, so that they don't intersect and don't skip data on failures.
5. Fixed single-value requests that didn't use `LIMIT 1` and/or queried `binary` instead of `id`.

Helpers:
1. `InvalidBlockNumber`.
2. `RunWithRetries`.


nil/cmd/exporter/internal/clickhouse/scheme.go
```diff
@@ -82,8 +81,44 @@ func createTableQuery(tableName, fields, engine string, primaryKeys, orderKeys [
...
+	}
+
+	var version common.Hash
+	if err := conn.QueryRow(ctx, `
+		SELECT hash
+		FROM blocks
+		WHERE shard_id = 0 AND id = 0
+	`).Scan(&version); err != nil {
+		if errors.Is(err, sql.ErrNoRows) {
+			return common.Hash{}, nil
```
olegrok, line 118: Here and below
```suggestion
			return common.EmptyHash, nil
```

olegrok (approved):

nil/cmd/exporter/internal/clickhouse/scheme.go
dmtrskv: Sorry, emergency merging for release. Will fix it afterwards

PR #436, opened 2025-03-01
Fix Typographical Errors in Documentation and Code Comments

SITADRITA1:
1. Fixing Typographical Error
src/sea.ts
Change:
Before: // After bandling, this will be the current file, so we can just return the
After: // After bundling, this will be the current file, so we can just return the
Reason:
"Bandling" is a typographical error. The correct word is "bundling," which refers to the process of combining multiple files into one.


knazarov:
Please don't submit patches that change grammar. Open issues for that instead. 
We accept contributions for significant functional changes, and small PRs such as this will be closed without merging.

If in doubt - check the CONTRIBUTING.md file.

PR #437, opened 2025-03-01
Correct Return Type for run Method to Improve Type Safety

SITADRITA1:
Fix Incorrect Return Type

src/commands/keygen.ts
Change:
Before: public async run(): Promise<Hex>
After: public async run(): Promise<void>
Reason:
The run method does not return a Hex value explicitly. The method logs values but does not return any usable value. Changing the return type to void correctly reflects that the function does not produce a meaningful return value. This improves code clarity and prevents potential type mismatches.


knazarov:
The mentioned method does return a value. The proposed patch looks incorrect.

SITADRITA1:
@knazarov  I can revise it to make it correct.

PR #440, opened 2025-03-01
[archive node] Drop on version change

dmtrskv:
On syncer initialization (called only on a single syncer on the node) the version is checked. Main shard genesis block hash serves as the version, as in the case of exporter.
If local and remote versions differ:
- if `--allow-db-clear`, the existing db is dropped and the snapshot is fetched;
- by default, an error is returned.

Added tests that restart the archive node with and without reset.

For now, before this code is on all the nodes, there is no way to fetch a remote version, so an error there is handled stoically.

### DIFF
--- a/nil/cmd/nild/main.go
+++ b/nil/cmd/nild/main.go
@@ -38,7 +38,7 @@ func main() {
 
 	profiling.Start(cfg.PprofPort)
 
-	database, err := openDb(cfg.DB.Path, cfg.DB.AllowDrop, logger)
+	database, err := openDb(cfg.DB.Path, cfg.AllowDbDrop, logger)
 	check.PanicIfErr(err)
 
 	if len(cfg.ReadThrough.SourceAddr) != 0 {
@@ -109,7 +109,7 @@ func addTelemetryFlags(fset *pflag.FlagSet, cfg *nildconfig.Config) {
 }
 
 func addAllowDbClearFlag(fset *pflag.FlagSet, cfg *nildconfig.Config) {
-	fset.BoolVar(&cfg.DB.AllowDrop, "allow-db-clear", cfg.DB.AllowDrop, "allow to clear database in case of outdated version")
+	fset.BoolVar(&cfg.AllowDbDrop, "allow-db-clear", cfg.AllowDbDrop, "allow to clear database in case of outdated version")
 }
 
 func addRpcNodeFlags(fset *pflag.FlagSet, cfg *nildconfig.Config) {
@@ -227,6 +227,12 @@ func parseArgs() *nildconfig.Config {
 
 	logging.SetupGlobalLogger(*logLevel)
 	check.PanicIfErr(logging.SetLibp2pLogLevel(*libp2pLogLevel))
+
+	// todo: remove it when we remove the old flag
+	// Support old flag for backward compatibility
+	if cfg.DB.AllowDrop {
+		cfg.Config.AllowDbDrop = cfg.DB.AllowDrop
+	}
 
 	if cfg.Replay.BlockIdLast == 0 {
 		cfg.Replay.BlockIdLast = cfg.Replay.BlockIdFirst
--- a/nil/internal/collate/bootstrap.go
+++ b/nil/internal/collate/bootstrap.go
@@ -2,15 +2,21 @@ package collate
 
 import (
 	"context"
+	"fmt"
 
+	"github.com/NilFoundation/nil/nil/common"
+	"github.com/NilFoundation/nil/nil/common/check"
 	"github.com/NilFoundation/nil/nil/common/logging"
 	"github.com/NilFoundation/nil/nil/internal/db"
 	"github.com/NilFoundation/nil/nil/internal/network"
+	"github.com/NilFoundation/nil/nil/internal/types"
 	"github.com/rs/zerolog"
 )
 
+const topicVersion = "/nil/version"
+
 func topicBootstrapShard() network.ProtocolID {
-	return network.ProtocolID("nil/snap")
+	return "/nil/snap"
 }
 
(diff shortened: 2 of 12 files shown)


nil/internal/network/req_resp.go
```diff
@@ -103,9 +103,7 @@ func (m *Manager) SendRequestAndGetResponse(ctx context.Context, peerId PeerID,
 }
 
 func (m *Manager) SetRequestHandler(ctx context.Context, protocolId ProtocolID, handler RequestHandler) {
-	logger := m.logger.With().Str(logging.FieldProtocolID, string(protocolId)).Logger()
-
-	logger.Debug().Msg("Setting request handler...")
+	logger := m.logger.With().Str(logging.FieldProtocolID, m.withNetworkPrefix(string(protocolId))).Logger()
```
dmtrskv, line 106: fixed topic name logging and removed extra log line

nil/tests/archive_node/archive_node_test.go
```diff
@@ -1,55 +1,130 @@
...
+				return b != nil
+			}, 5*time.Second, 100*time.Millisecond)
+		}
 	}
+
+	s.Run("Before", check)
+
+	// Restart the archive node
+	s.stopArchiveNode()
+	s.startArchiveNode()
```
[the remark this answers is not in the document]
olegrok: I overlooked this moment. Could we get only significant part of `t.Name()` to avoid this issue?

nil/tests/sharded_suite.go
```diff
@@ -272,11 +275,21 @@ type ArchiveNodeConfig struct {
 func (s *ShardedSuite) StartArchiveNode(params *ArchiveNodeConfig) (client.Client, network.AddrInfo) {
```
olegrok, line 275: Maybe let's accept context as an argument here?

nil/tests/archive_node/archive_node_test.go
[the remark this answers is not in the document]
dmtrskv: I'll look into it after merging this!

nil/tests/sharded_suite.go
dmtrskv: I like it being optional, so that in general case you don't need to think
the overrides are useful only in the particular test that restarts the services

nil/cmd/nild/main.go
```diff
@@ -228,6 +228,8 @@ func parseArgs() *nildconfig.Config {
 	logging.SetupGlobalLogger(*logLevel)
 	check.PanicIfErr(logging.SetLibp2pLogLevel(*libp2pLogLevel))
 
+	cfg.Config.AllowDbDrop = cfg.DB.AllowDrop
```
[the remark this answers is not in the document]
dmtrskv: True.
Made old variable deprecated and added a todo here.

nil/tests/archive_node/archive_node_test.go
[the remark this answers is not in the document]
dmtrskv: Ok, it was easy, did it here.

nil/internal/collate/bootstrap.go
```diff
@@ -2,15 +2,21 @@ package collate
...
+	"github.com/NilFoundation/nil/nil/internal/types"
 	"github.com/rs/zerolog"
 )
 
+const topicVersion = "/nil/version"
+
 func topicBootstrapShard() network.ProtocolID {
-	return network.ProtocolID("nil/snap")
+	return "/nil/snap"
 }
```
olegrok, line 20: In theory it can be also specified as a constant. To be consistent with `topicVersion`

nil/internal/collate/syncer.go
```diff
@@ -109,41 +110,105 @@ func (s *Syncer) notify(blockId types.BlockNumber) {
...
+func (s *Syncer) fetchSnapshot(ctx context.Context) error {
+	if len(s.config.BootstrapPeers) == 0 {
+		s.logger.Warn().Msg("No bootstrap peers to fetch snapshot from")
+		return nil
+	}
+
+	var err error
+	for _, peer := range s.config.BootstrapPeers {
+		err = fetchSnapshot(ctx, s.networkManager, peer, s.db, s.logger)
+		if err == nil {
```
olegrok, line 148:
```suggestion
		if err = fetchSnapshot(ctx, s.networkManager, peer, s.db, s.logger); err == nil {
```

nil/internal/network/pubsub.go
```diff
@@ -151,6 +146,10 @@ func (ps *PubSub) Subscribe(topic string) (*Subscription, error) {
 		return nil, err
 	}
 
+	logger := ps.logger.With().
+		Str(logging.FieldComponent, "sub").
+		Str(logging.FieldTopic, t.String()).
```
olegrok, line 151:
```suggestion
		Stringer(logging.FieldTopic, t).
```

nil/internal/collate/bootstrap.go
dmtrskv: I conservatively decided to leave it as is because it can be actually parameterized by the shard id later.

nil/internal/collate/syncer.go
dmtrskv: In your form, I expect `err :=` when scanning the code, that's why I moved it to an explicit separate line

nil/internal/network/pubsub.go
dmtrskv: I did it intentionally because it's not clear how come topic is Stringer and so the alternative looked somehow erroneous to me. With non-string-like types I usually expect an explicit method call returning some string.
Now I don't think that it was such a bright idea, but I already pressed merge...

Issue #79, opened 2025-01-31
Looping in case of a transaction that exceeds the gas limit per block


knazarov:
If we get a transaction whose execution requires more gas than the block limit, we will get into an [infinite loop](https://github.com/NilFoundation/nil/blob/73a21cdd3850f53ffe068bb0e61ae5a1bb767ba2/nil/internal/collate/proposer.go#L298), discarding that transaction from each block and forever trying to add it to the next.

PR #447, opened 2025-03-03
[collate] fix tx stucking when it reaches block gas limit

olegrok:
If a transaction consumed more gas than the block allowed, it just got stuck in the neighbor out tx tree. That prevented the execution of other transactions. This patch should fix it.

See also https://github.com/NilFoundation/nil/issues/79

### DIFF
--- a/nil/internal/collate/proposer.go
+++ b/nil/internal/collate/proposer.go
@@ -420,15 +420,8 @@ func (p *proposer) handleTransactionsFromNeighbors(tx db.RoTx) error {
 							Stringer(logging.FieldTransactionHash, txnHash).
 							Msg("Invalid internal transaction")
 					} else {
-						snapshot := p.executionState.Snapshot()
 						if err := p.handleTransaction(txn, txnHash, execution.NewTransactionPayer(txn, p.executionState)); err != nil {
 							return err
-						}
-
-						if !checkLimits() {
-							p.executionState.RevertToSnapshot(snapshot)
-							p.executionState.DropInTransaction()
-							break
 						}
 					}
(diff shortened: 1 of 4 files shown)


nil/internal/collate/proposer.go
```diff
@@ -414,29 +414,32 @@ func (p *proposer) handleTransactionsFromNeighbors(tx db.RoTx) error {
...
 					if err != nil {
 						return err
 					}
 					p.proposal.InternalTxnRefs = append(p.proposal.InternalTxnRefs, ref)
+
+					if !checkLimits() {
+						// Handle at least one transaction.
+						if len(p.proposal.InternalTxnRefs) > 1 {
+							p.executionState.RevertToSnapshot(snapshot)
+							p.executionState.DropInTransaction()
```
dmtrskv, line 439: We haven't added the txn if it wasn't handled. That's why this check was only in one branch of if
Also, we can just always add the last transaction by removing snapshot and revert altogether

nil/internal/collate/proposer.go
```diff
@@ -414,29 +414,32 @@ func (p *proposer) handleTransactionsFromNeighbors(tx db.RoTx) error {
 				}
 
 				if txn.To.ShardId() == p.params.ShardId {
+					snapshot := p.executionState.Snapshot()
```
Zerg1996, line 417: I'm not sure we need snapshot if validation wasn't successful.

Zerg1996 (approved):
Thanks, for PR! lgtm

shermike (approved):

PR #448, opened 2025-03-03
[HOTFIX] [collate] fix tx stucking when it reaches block gas limit

olegrok:
If a transaction consumed more gas than the block allowed, it just got stuck in the neighbor out tx tree. That prevented the execution of other transactions. This patch should fix it.


Un1oR:
Let's refer to #79 somehow

Un1oR:
By the way, won't exceeding the block size limit interfere with the proof? Or is it a temporary solution anyway, so don't worry about it?

PR #449, opened 2025-03-03
Added simple metrics for loadgen

Zerg1996:
This PR adds simple metrics for the Uniswap load generator service. It helps us gather statistics and analyze the distribution of calls from SmartAccount to Pair.

### DIFF
--- a/nil/services/nil_load_generator/service.go
+++ b/nil/services/nil_load_generator/service.go
@@ -17,6 +17,7 @@ import (
 	"github.com/NilFoundation/nil/nil/services/cliservice"
 	"github.com/NilFoundation/nil/nil/services/faucet"
 	uniswap "github.com/NilFoundation/nil/nil/services/nil_load_generator/contracts"
+	"github.com/NilFoundation/nil/nil/services/nil_load_generator/metrics"
 	"github.com/NilFoundation/nil/nil/services/rpc"
 	"github.com/NilFoundation/nil/nil/services/rpc/httpcfg"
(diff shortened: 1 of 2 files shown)


nil/services/nil_load_generator/service.go
```diff
@@ -340,32 +355,38 @@ func Run(ctx context.Context, cfg Config, logger zerolog.Logger) error {
...
 				pairsToCall, smartAccountsToCall, err := selectPairAndAccount(shardIdList)
 				if err != nil {
+					handler.RecordError(ctx)
 					return err
 				}
 				if err := parallelizeAcrossN(len(pairsToCall), func(i int) error {
+					handler.RecordFromToCall(ctx, int64(smartAccountsToCall[i]-1), int64(pairsToCall[i]-1))
 					return swap(ctx, smartAccountsToCall[i]-1, pairsToCall[i]-1, logger)
 				}); err != nil {
+					handler.RecordError(ctx)
```
olegrok, line 381: Maybe it's better to make something like

```
defer func() {
  if err != nil {
     recordError()
  }
}()
```

olegrok (approved):

nil/services/nil_load_generator/service.go
avm: It's hard to avoid accidentally declaring new `err` variables in `if err := doSomething(); err != nil` statements, shadowing the `err` variable captured in such a `defer`red func.

PR #450, opened 2025-03-03
Fix replacement in txpool

### LINKED ISSUE #446: Transaction Replacement  Rule Should Require Higher Bidding
## Context

Currently, the transaction replacement logic in the [txnpool.go](https://github.com/NilFoundation/nil/blob/55b51a89442f4dea3fcf6d074192469d1ea288aa/nil/services/txnpool/txnpool.go#L231) allows replacement if the candidate has a **higher FeeCredit**:  

```go
func shouldReplace(existing, candidate *metaTxn) bool {
	if candidate.FeeCredit.Cmp(existing.FeeCredit) <= 0 {
		return false
	}

	// Discard the previous transaction if it is the same but at a lower fee
	existingFee := existing.FeeCredit
	existing.FeeCredit = candidate.FeeCredit
	defer func() {
		existing.FeeCredit = existingFee
	}()

	return bytes.Equal(existing.Transaction.Hash().Bytes(), candidate.Hash().Bytes())
```
Simply increasing FeeCredit does not guarantee faster inclusion if the candidates' maxFeePerGas < baseFeePerGas. In fact it does nothing, it just assigns more funds to spend.  

## Proposed Fix: Use `effectivePriorityFee` for Replacement  

A transaction should only be replaced if the candidates **effectivePriorityFee** is strictly higher:  

effectivePriorityFee = min(maxPriorityFeePerGas, MaxFeePerGas -  baseFee)

To prevent micro-adjustment spam, enforce a 5% minimum increase (compared to Go-Ethereum’s 10% rule):

effectivePriorityFeeCandidate ≥ FEE_BUMP_MULTIPLIER × effectivePriorityFeeExisting

For now, we set: 
FEE_BUMP_MULTIPLIER = 1.05 

## Implementation Example

```python
def effective_priority_fee(max_fee_per_gas, base_fee, max_priority_fee_per_gas):
    return min(max_priority_fee_per_gas, max_fee_per_gas - base_fee)

def should_replace(existing, candidate, base_fee):
    existing_epf = effective_priority_fee(existing["max_fee_per_gas"], base_fee, existing["max_priority_fee_per_gas"])
    candidate_epf = effective_priority_fee(candidate["max_fee_per_gas"], base_fee, candidate["max_priority_fee_per_gas"])

    return candidate_epf >= existing_epf * FEE_BUMP_MULTIPLIER

```

#### Example: Candidate with a Higher `maxFeePerGas` but a Lower Actual Payment  

| Transaction  | maxFeePerGas | maxPriorityFeePerGas | baseFee | EffectivePriorityFeePerGas |
|-------------|-------------|----------------------|---------|----------------------|
| **Existing** | 50 Gwei     | 10 Gwei             | 40 Gwei | **10 Gwei** (min(10, 50 - 40)) |
| **Candidate** | 60 Gwei     | 5 Gwei              | 40 Gwei | **5 Gwei** (min(5, 60 - 40)) |

Even though the candidate has a **higher `maxFeePerGas` (60 > 50)**, it actually **pays less per gas unit** (EffectivePriorityFee **5 < 10**). Checking only `maxFeePerGas` would allow a worse-paying transaction to replace a better one.  

#### Example: Candidate with a Higher `maxPriorityFeePerGas` but a Lower Actual Payment  

| Transaction  | maxFeePerGas | maxPriorityFeePerGas | baseFee | EffectivePriorityFeePerGas |
|-------------|-------------|----------------------|---------|----------------------|
| **Existing** | 100 Gwei    | 10 Gwei             | 90 Gwei | **10 Gwei** (min(10, 100 - 90)) |
| **Candidate** | 95 Gwei     | 12 Gwei             | 90 Gwei | **5 Gwei** (min(12, 95 - 90)) |

The candidate has a **higher `maxPriorityFeePerGas` (12 > 10)** but actually pays **less per unit of gas (5 < 10)**. Checking only `maxPriorityFeePerGas` would allow a lower-paying transaction to replace a higher one.

### Outcomes

Ensures replacements only happen when the candidate pays more per gas unit.
Prevents replacement spam by enforcing a 5% minimum increase 
Aligns transaction replacement logic with higher bidding principles

## EDIT NOTES:

Since the current implementation computes EffectivePriorityFee, it is easier to just compare EffectivePriorityFeePerGas, there is no need to add baseFee to it to get EffectiveGasPrice

shermike:
During replacement, take into account effective priority fee, not just feeCredit.

Closes #446

### DIFF
--- a/nil/services/txnpool/txnpool.go
+++ b/nil/services/txnpool/txnpool.go
@@ -1,7 +1,6 @@
 package txnpool
 
 import (
-	"bytes"
 	"container/heap"
 	"context"
 	"fmt"
@@ -15,6 +14,11 @@ import (
 	"github.com/NilFoundation/nil/nil/internal/types"
 	"github.com/rs/zerolog"
 )
+
+// FeeBumpPercentage is the percentage of the priorityFee that a transaction must exceed to replace another transaction.
+// For example, if the priorityFee of a transaction is 100 and FeeBumpPercentage is 5, then the transaction must have a
+// priorityFee of at least 105 to replace the existing transaction.
+const FeeBumpPercentage = 5
 
 type Pool interface {
 	Add(ctx context.Context, txns ...*types.Transaction) ([]DiscardReason, error)
@@ -229,18 +233,8 @@ func (p *TxnPool) getLocked(hash common.Hash) *metaTxn {
 }
 
 func shouldReplace(existing, candidate *metaTxn) bool {
-	if candidate.FeeCredit.Cmp(existing.FeeCredit) <= 0 {
-		return false
-	}
-
-	// Discard the previous transaction if it is the same but at a lower fee
-	existingFee := existing.FeeCredit
-	existing.FeeCredit = candidate.FeeCredit
-	defer func() {
-		existing.FeeCredit = existingFee
-	}()
-
-	return bytes.Equal(existing.Transaction.Hash().Bytes(), candidate.Hash().Bytes())
+	adjustedFee := existing.effectivePriorityFee.Mul64(100 + FeeBumpPercentage).Div64(100)
+	return candidate.effectivePriorityFee.Cmp(adjustedFee) >= 0
 }
 
 func (p *TxnPool) addLocked(txn *metaTxn) DiscardReason {
@@ -251,8 +245,6 @@ func (p *TxnPool) addLocked(txn *metaTxn) DiscardReason {
 		if !shouldReplace(found, txn) {
 			return NotReplaced
 		}
-
-		p.queue.Remove(found)
 		p.discardLocked(found, ReplacedByHigherTip)
 	}
 
--- a/nil/services/txnpool/txnpool_test.go
+++ b/nil/services/txnpool/txnpool_test.go
@@ -79,12 +79,14 @@ func (s *SuiteTxnPool) addTransactionsToPoolSuccessfully(pool Pool, txn ...*type
 	s.Equal(count+len(txn), s.getTransactionCount(pool))
 }
 
-func (s *SuiteTxnPool) addTransactions(txn ...*types.Transaction) {
+func (s *SuiteTxnPool) addTransactions(txn ...*types.Transaction) []DiscardReason {
 	s.T().Helper()
 
 	reasons, err := s.pool.Add(s.ctx, txn...)
 	s.Require().NoError(err)
 	s.Require().Len(reasons, len(txn))
+
+	return reasons
 }
 
 func (s *SuiteTxnPool) getTransactions() []*types.TxnWithHash {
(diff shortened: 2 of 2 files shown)


nil/services/txnpool/txnpool.go
```diff
@@ -229,18 +228,8 @@ func (p *TxnPool) getLocked(hash common.Hash) *metaTxn {
...
-	// Discard the previous transaction if it is the same but at a lower fee
-	existingFee := existing.FeeCredit
-	existing.FeeCredit = candidate.FeeCredit
-	defer func() {
-		existing.FeeCredit = existingFee
-	}()
-
-	return bytes.Equal(existing.Transaction.Hash().Bytes(), candidate.Hash().Bytes())
+	adjustedFee := existing.effectivePriorityFee.Mul64(105).Div64(100)
+	return candidate.effectivePriorityFee.Cmp(adjustedFee) >= 0
```
avm, line 232: Why don't we check hash equality anymore?

nil/services/txnpool/txnpool.go
```diff
@@ -229,18 +228,8 @@ func (p *TxnPool) getLocked(hash common.Hash) *metaTxn {
...
-
-	// Discard the previous transaction if it is the same but at a lower fee
-	existingFee := existing.FeeCredit
-	existing.FeeCredit = candidate.FeeCredit
-	defer func() {
-		existing.FeeCredit = existingFee
-	}()
-
-	return bytes.Equal(existing.Transaction.Hash().Bytes(), candidate.Hash().Bytes())
+	adjustedFee := existing.effectivePriorityFee.Mul64(105).Div64(100)
```
avm, line 231: 105 and 100 are magic numbers. Please create constants with descriptive names.

nil/services/txnpool/txnpool.go
alex-damjanovic: There is no need for this check because a replacement transaction doesn’t have to be identical to the original. For example, if you initially submitted a transaction transferring 10 ETH but later realized you meant to send 0.1 ETH, you should be able to replace it if it is still in mempool. The key condition is that the replacement must offer a higher reward to validators, ensuring economic incentives align, and there is no spamming of replacements.

nil/services/txnpool/txnpool.go
alex-damjanovic: Agreed, it will make it easier to change in the future if needed.

shermike: It just adds 5% to existing effectivePriorityFee. Added comment about it.

alex-damjanovic (commented):
lgtm

nil/services/txnpool/txnpool.go
avm: Multiplying by 105 and dividing by 100 indeed adds 5%, I have no issue with that, but 5% is not a universal constant. If it ever changes, we'll have to hunt through the codebase, searching for incorrect assumptions. Please create a constant with a descriptive name and use it in the implementation and tests.

shermike: Done.

avm (approved):

PR #455, opened 2025-03-04
[consensus] add metrics

olegrok:
This patch adds some minimal telemetry to consensus subsystem. Right now it's BuildProposal/InsertProposal/RunSequence time measuremants, amout of validators for some specific height and counters for sent and received messages.


nil/internal/telemetry/telattr/attributes.go
```diff
@@ -27,3 +27,15 @@ func ProtocolId(id protocol.ID) attribute.KeyValue {
 func Topic(topic string) attribute.KeyValue {
 	return attribute.String(logging.FieldTopic, topic)
 }
+
+func Height(height uint64) attribute.KeyValue {
```
Zerg1996, line 31: Sorry, I missed where we use this func...

Zerg1996 (approved):

PR #462, opened 2025-03-04
[go-ibft] replace hashicorp/go-metrics with otel

olegrok:
Just adapt go-ibft for our approach with metrics.


nil/go-ibft/core/ibft.go
```diff
@@ -101,14 +105,34 @@ type IBFT struct {
...
+	if err != nil {
+		return nil, err
+	}
+
+	sequenceHistogram, err := meter.Float64Histogram("sequence.duration", metric.WithDescription("Sequence duration"))
+	if err != nil {
+		return nil, err
+	}
+
 	return &IBFT{
```
Un1oR, line 136: Maybe we could make a `NewIBFTWithMetrics` wrapper and use it in the combat code and leave it as is in the tests?

olegrok: Ok. Maybe it's better. Applied.

Un1oR (approved):
LGTM but I am not familiar with the specifics of the libraries used

Zerg1996 (approved):

PR #463, opened 2025-03-04
[loadgen] main-keys-path parameter removed

Un1oR:
`cliservice.NewService` accepts the private key of the smart account, with which the corresponding smart account management commands will work. We don't use `MainSmartAccount` for anything in the load generator, so we don't need credentials from it. Instead, we use `Faucet`, which does not require authentication.

At the same time, the presence of the flag introduces an inconvenience for local debugging. Since `MainPrivateKey` is now only saved to the filesystem when using `nil gen-configs`, we can't just do a `nil run` and then direct the load generator to that instance, we have nowhere to get the `MainPrivateKey` file. I think this adds an inconvenience that is not necessary at all in this case. Let our tools require as much bodywork as possible for a minimally working setup.

N.B. Also need to roll back the corresponding [PR](https://github.com/NilFoundation/infra/pull/586) in `infra`


olegrok (approved):
It was my question why didn't we use faucet. But we seems to use it.

PR #464, opened 2025-03-04
[otel] fix semconv version

olegrok:
To fix following issue on nildev:
```
ERROR=failed to initialize metric provider: conflicting Schema URL: https://opentelemetry.io/schemas/1.26.0 and https://opentelemetry.io/schemas/1.27.0
    JSON={"level":"error","component":"nil","error":"failed to initialize metric provider: conflicting Schema URL: https://opentelemetry.io/schemas/1.26.0 and https://opentelemetry.io/schemas/1.27.0","caller":"github.com/NilFoundation/nil/nil/services/nilservice/service.go:351","time":"2025-03-04T16:59:45Z","message":"Failed to initialize telemetry"}wq
```

### DIFF
--- a/nil/internal/telemetry/internal/resource.go
+++ b/nil/internal/telemetry/internal/resource.go
@@ -2,7 +2,7 @@ package internal
 
 import (
 	"go.opentelemetry.io/otel/sdk/resource"
-	semconv "go.opentelemetry.io/otel/semconv/v1.27.0"
+	semconv "go.opentelemetry.io/otel/semconv/v1.26.0" // Be sure otelcol support it before update
 )
 
 func NewResource(config *Config) (*resource.Resource, error) {
--- a/nix/nil.nix
+++ b/nix/nil.nix
@@ -49,7 +49,7 @@ buildGo124Module rec {
   ];
 
   # to obtain run `nix build` with vendorHash = "";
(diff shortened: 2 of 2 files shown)


nil/internal/telemetry/internal/resource.go
```diff
@@ -2,7 +2,7 @@ package internal
 
 import (
 	"go.opentelemetry.io/otel/sdk/resource"
-	semconv "go.opentelemetry.io/otel/semconv/v1.27.0"
+	semconv "go.opentelemetry.io/otel/semconv/v1.26.0"
```
Un1oR, line 5: Let's add a comment on why it's important to use this particular version

Un1oR (approved):

nil/internal/telemetry/internal/resource.go
olegrok: I'm not sure it's important. I hope it should be fixed on otelcol-contrib side. But I didn't find specific way in their documentation

Un1oR: Ok, we could just leave a warning about possible issues with the change that are worth checking out)

olegrok: It's reasonable

PR #471, opened 2025-03-05
Add a "governance" system contract that will be able to initiate rollbacks

avm:
During network launch, we may encounter misconfigurations, vulnerabilities or bugs in the validators or the provers that may result in malicious or breaking changes to the state of the blockchain. One way to deal with them at early stages of the mainnet is "rolling back" to a known good state when an authority decides to do so.

This patch adds a "governance" system contract on the main shard and a corresponding precompile, accepting rollback transactions signed by a "sudo key" (currently the same as the "main key" of the built-in system smart account). The rollback functionality itself will be implemented by following patches.

### DIFF
--- a/nil/internal/types/block.go
+++ b/nil/internal/types/block.go
@@ -50,6 +50,12 @@ type BlockData struct {
 	BaseFee             Value            `json:"gasPrice" ch:"gas_price"`
 	GasUsed             Gas              `json:"gasUsed" ch:"gas_used"`
 	L1BlockNumber       uint64           `json:"l1BlockNumber" ch:"l1_block_number"`
+
+	// Incremented after every rollback, used to prevent rollback replay attacks
+	RollbackCounter uint32 `json:"rollbackCounter" ch:"rollback_counter"`
+	// Required validator patchLevel, incremented if validator updates
+	// are required to mitigate an issue
+	PatchLevel uint32 `json:"patchLevel" ch:"patch_level"`
 }
 
 type ConsensusParams struct {
--- a/nil/internal/vm/precompiled.go
+++ b/nil/internal/vm/precompiled.go
@@ -99,6 +99,7 @@ var (
 	SendRequestAddress        = types.BytesToAddress([]byte{0xd8})
 	CheckIsResponseAddress    = types.BytesToAddress([]byte{0xd9})
 	LogAddress                = types.BytesToAddress([]byte{0xda})
+	GovernanceAddress         = types.BytesToAddress([]byte{0xdb})
 )
 
 // PrecompiledContractsPrague contains the set of pre-compiled Ethereum
@@ -140,6 +141,7 @@ var PrecompiledContractsPrague = map[types.Address]PrecompiledContract{
 	SendRequestAddress:        &sendRequest{},
 	CheckIsResponseAddress:    &checkIsResponse{},
 	LogAddress:                &emitLog{},
+	GovernanceAddress:         &governance{},
 }
 
 // RunPrecompiledContract runs and evaluates the output of a precompiled contract.
@@ -183,7 +185,7 @@ type simple struct {
 	contract SimplePrecompiledContract
 }
 
-var _ PrecompiledContract = (*simple)(nil)
+var _ ReadOnlyPrecompiledContract = (*simple)(nil)
 
 func (a *simple) RequiredGas(input []byte, state StateDBReadOnly) (uint64, error) {
 	return a.contract.RequiredGas(input), nil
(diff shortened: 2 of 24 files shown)


smart-contracts/contracts/Nil.sol
```diff
@@ -373,6 +374,20 @@ library Nil {
...
+    /**
+     * @dev Initiates a rollback
+     */
+    function rollback(
+        uint32 version,
+        uint32 counter,
+        uint32 patchLevel,
+        uint64 mainBlockId /*,
+        uint32 replayDepth,
+        uint32 searchDepth */
```
dmtrskv, line 386: a comment on what's with the comments would be nice

nil/internal/vm/precompiled.go
```diff
@@ -726,6 +722,62 @@ func (a *checkIsInternal) Run(state StateDBReadOnly, input []byte, value *uint25
...
+	}
+	return args, nil
+}
+
+type governance struct{}
+
+var _ ReadWritePrecompiledContract = (*governance)(nil)
+
+func (g *governance) RequiredGas([]byte, StateDBReadOnly) (uint64, error) {
+	return 10, nil
```
dmtrskv, line 746: I wonder about putting all these gas values as constants into a single file. Not relevant to this PR tho

nil/internal/vm/precompiled.go
```diff
@@ -726,6 +722,62 @@ func (a *checkIsInternal) Run(state StateDBReadOnly, input []byte, value *uint25
...
+	version, ok := args[0].(uint32)
+	if !ok || version != 1 {
+		return nil, types.NewVmError(types.ErrorPrecompileWrongVersion)
+	}
+
+	counter, ok1 := args[1].(uint32)
+	patchLevel, ok2 := args[2].(uint32)
+	mainBlockId, ok3 := args[3].(uint64)
+	if !ok1 || !ok2 || !ok3 {
+		return nil, types.NewVmError(types.ErrorPrecompileWrongNumberOfArguments)
```
dmtrskv, line 768: I suppose it's `ErrorAbiUnpackFailed` while the number of args is handled in your helper (thanks!)

nil/internal/vm/precompiled.go
```diff
@@ -726,6 +722,62 @@ func (a *checkIsInternal) Run(state StateDBReadOnly, input []byte, value *uint25
...
+	}
+
+	err = state.Rollback(counter, patchLevel, mainBlockId)
+
+	res := make([]byte, 32)
+	if err == nil {
+		res[31] = 1
+	}
+
+	return res, nil
```
dmtrskv, line 778: Shouldn't you return the error?

nil/internal/vm/precompiled.go
```diff
@@ -726,6 +722,62 @@ func (a *checkIsInternal) Run(state StateDBReadOnly, input []byte, value *uint25
...
+		return nil, types.NewVmError(types.ErrorPrecompileWrongCaller)
+	}
+
+	args, err := precompiledArgs("precompileRollback", input, 4)
+	if err != nil {
+		return nil, err
+	}
+
+	version, ok := args[0].(uint32)
+	if !ok || version != 1 {
```
dmtrskv, line 760: should it be the same as `validatorPatchLevel` from proposer?

nil/tests/governance/rollback_test.go
```diff
@@ -0,0 +1,55 @@
...
+		MainBlockId: 2,
+		ReplayDepth: 3,
+		SearchDepth: 4,
+	}
+
+	calldata, err := collate.CreateRollbackCalldata(params)
+	s.Require().NoError(err)
+
+	receipt := s.SendExternalTransaction(calldata, types.GovernanceAddress)
+	s.Require().NotNil(receipt)
```
dmtrskv, line 48: what does it actually check?
e.g. it doesn't check for success

nil/internal/vm/precompiled.go
avm: We may want to bump `patchLevel` but keep `version` the same – e.g. we need a higher `patchLevel` to avoid a specific attack on the blockchain, but the rollback transactions themselves are still backwards compatible and can be applied by old validators. So `version` and `patchLevel` are distinct.

nil/internal/types/block.go
```diff
@@ -48,6 +48,8 @@ type BlockData struct {
 	BaseFee             Value            `json:"gasPrice" ch:"gas_price"`
 	GasUsed             Gas              `json:"gasUsed" ch:"gas_used"`
 	L1BlockNumber       uint64           `json:"l1BlockNumber" ch:"l1_block_number"`
+	RollbackCounter     uint32           `json:"rollbackCounter" ch:"rollback_counter"`
```
olegrok, line 51: Where this value is set/updated?

nil/internal/types/block.go
```diff
@@ -48,6 +48,8 @@ type BlockData struct {
...
+	PatchLevel          uint32           `json:"patchLevel" ch:"patch_level"`
```
shermike, line 52: Could you please add some comments about the purpose of these variables? I know we don't comment most of the variables here, but let's start doing it. It will definitely help to read the code.

olegrok (approved):

PR #476, opened 2025-03-06
relax block gen unix socket permissions

oclaw:
0o666 is too strict mode which makes impossible to operate on the domain socket after nil_block_generator finishes its work. Relaxing it to owner RW priveleges

### DIFF
--- a/nil/cmd/nil_block_generator/internal/commands/config.go
+++ b/nil/cmd/nil_block_generator/internal/commands/config.go
@@ -13,6 +13,7 @@ const (
 	configFilePath = "config.json"
 	nilDBPath      = "test.db"
 	fileMode       = 0o666
+	directoryMode  = 0o744
 )
 
 type Contract struct {
@@ -167,7 +168,7 @@ func ShowCalls(writer io.StringWriter) error {
 }
 
 func GetSockPath() (string, error) {
-	err := os.Mkdir(socketDir, fileMode)
+	err := os.Mkdir(socketDir, directoryMode)
 	if err != nil {
 		return "", err
 	}


nil/cmd/nil_block_generator/internal/commands/config.go
```diff
@@ -12,7 +12,7 @@ const (
 	socketDir      = "nil_block_generator_tmp"
 	configFilePath = "config.json"
 	nilDBPath      = "test.db"
-	fileMode       = 0o666
+	fileMode       = 0o744
```
avm, line 15: We are adding 0o100 (execution by owner) and retracting 0o022 (write by group and others). How does that help?

oclaw: I am not sure how it works but without execution rights block generator fails to use this socket after creation. The other rights are just not needed for the case

oclaw: The original issue I was trying to solve is that nil_block_generator now requires super user privileges to run. And it helps :)

avm: `fileMode` is also used for the config file on line 106. Let's separate those constants and add a comment about the mysterious behavior of the exec bit.

oclaw: good point, done

nil/cmd/nil_block_generator/internal/commands/config.go
```diff
@@ -167,7 +168,7 @@ func ShowCalls(writer io.StringWriter) error {
 }
 
 func GetSockPath() (string, error) {
-	err := os.Mkdir(socketDir, fileMode)
+	err := os.Mkdir(socketDir, sockFileMode)
```
avm, line 171: OK, it's a directory, that explains the exec bit! I suggest we call this `directoryMode` and set it to `0o777`.

oclaw: deal :)

PR #478, opened 2025-03-06
[txpool] Fix bug in transactions retrieving

shermike:
`Peek` method should be constant and should not modify content of txpool. But it did by calling `heap.Push(q, txn)`, because txn refers to the real metaTx and its `bestIndex` is changed.
Fixed by cloning it. Also added reset of `bestIndex` during popping from the queue.

### DIFF
--- a/nil/services/txnpool/meta_txn.go
+++ b/nil/services/txnpool/meta_txn.go
@@ -22,6 +22,15 @@ func newMetaTxn(txn *types.Transaction, baseFee types.Value) *metaTxn {
 	}
 }
 
+func (m *metaTxn) Clone() *metaTxn {
+	return &metaTxn{
+		TxnWithHash:          m.TxnWithHash,
+		effectivePriorityFee: m.effectivePriorityFee,
+		bestIndex:            m.bestIndex,
+		valid:                m.valid,
+	}
+}
+
 func (m *metaTxn) IsValid() bool {
 	return m.valid
 }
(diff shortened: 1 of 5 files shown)


olegrok (approved):

Un1oR:
How about adding more explicit diagnostics?
```diff
--- a/nil/services/txnpool/queue.go
+++ b/nil/services/txnpool/queue.go
@@ -2,6 +2,7 @@ package txnpool
 
 import (
        "container/heap"
+       "github.com/NilFoundation/nil/nil/common/assert"
 
        "github.com/NilFoundation/nil/nil/common/check"
 )
@@ -43,6 +44,9 @@ func (p *TxnQueue) Push(x any) {
 func (p *TxnQueue) Pop() any {
        old := p.txns
        n := len(old)
+       if assert.Enable {
+               check.PanicIfNotf(n > 0, "Failed to pop txn from empty queue")
+       }
        item := old[n-1]
        old[n-1] = nil // avoid memory leak
        p.txns = old[0 : n-1]
@@ -50,6 +54,9 @@ func (p *TxnQueue) Pop() any {
 }
 
 func (p *TxnQueue) Remove(txn *metaTxn) {
+       if assert.Enable {
+               check.PanicIfNotf(p.Len() > 0, "Failed to remove txn from empty queue")
+       }
        if txn.bestIndex >= 0 {
                heap.Remove(p, txn.bestIndex)
                txn.bestIndex = -1
```
Or even unconditionally

avm:
> We don't need it in `Pop` method because it is called only from `Remove`.

Maybe rename it to `pop` then?

shermike:
> Maybe rename it to `pop` then?

We can't because this interface is required by `Heap` package: https://github.com/golang/go/blob/master/src/container/heap/heap.go#L63

PR #479, opened 2025-03-06
[collate] add wait main shard block logic to block replay 

olegrok:
It's possible case that during block replay specific block.mainShardHash
is not available. That shouldn't happen normally since we replay main shard
really fast.
But because of network issues or something else it can be a problem.
So this patch adds simple logic that tries to wait expected block using
subscriptions.

### DIFF
--- a/nil/internal/collate/validator.go
+++ b/nil/internal/collate/validator.go
@@ -39,6 +39,10 @@ type Validator struct {
 	lastBlock     *types.Block
 	lastBlockHash common.Hash
 
+	subsMutex sync.Mutex
+	subsId    uint64
+	subs      map[uint64]chan types.BlockNumber
+
 	logger zerolog.Logger
 }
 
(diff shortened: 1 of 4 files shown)


nil/internal/collate/validator.go
```diff
@@ -335,6 +347,13 @@ func (s *Validator) replayBlockUnlocked(ctx context.Context, block *types.BlockW
 		return err
 	}
 
+	if !s.params.ShardId.IsMainShard() {
+		// Main shard should have its previous block
+		if err := s.mainShardValidator.WaitForBlock(ctx, prevBlock.MainChainHash); err != nil {
```
dmtrskv, line 352: maybe explain a bit more why do we wait for a main shard block and why it has to be this one?

dmtrskv (approved):

nil/internal/collate/validator.go
olegrok: This block contains config required to replay this block.
I'll put a comment tomorrow then.

PR #480, opened 2025-03-06
Implement nix for packaging rollup-contracts v2

AndreyMlashkin:
This PR adds rollup-contracts v2 to the deb-package of nil


nix/rollup-bridge-contracts.nix
```diff
@@ -0,0 +1,58 @@
...
+  buildPhase = ''
+    echo "Installing soljson"
+    (cd create-nil-hardhat-project; bash install_soljson.sh ${soljson26})
+    export BIOME_BINARY=${biome}/bin/biome
+
+    cd rollup-bridge-contracts
+    pwd
+    cp .env.example .env
+
+    export GETH_PRIVATE_KEY=002f28996b406c557ff579766af59ba66a3f103b8b90de6e9baad8ae211c0071
```
knazarov, line 45: Why not put it directly to .env.example?

rollup-bridge-contracts/contracts/NilAccessControl.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: MIT
-pragma solidity 0.8.27;
+pragma solidity 0.8.26;
```
akokoshn, line 2: `pragma solidity 0.8.27` is not supported by compiler?

akokoshn (commented):
could you please add description and squash  commits

rollup-bridge-contracts/contracts/NilAccessControl.sol
AndreyMlashkin: It should match to the sol compiler in nix and I fails to download the newer version.
Downgrading to the minor version should not have any effect

nix/rollup-bridge-contracts.nix
AndreyMlashkin: Good point, corrected

akokoshn (approved):

rollup-bridge-contracts/.env.example
```diff
@@ -6,8 +6,8 @@ L1_RPC_ENDPOINT="http://localhost:8545"
 WALLET_ADDRESS=""
 
 GETH_RPC_ENDPOINT="http://localhost:8545"
-GETH_PRIVATE_KEY=""
-GETH_WALLET_ADDRESS=""
+GETH_PRIVATE_KEY="002f28996b406c557ff579766af59ba66a3f103b8b90de6e9baad8ae211c0071"
```
knazarov, line 9: To not alarm anyone at these values, put a comment here that these are just for test and not in any way related to production.

knazarov (approved):
One comment but otherwise LGTM

PR #488, opened 2025-03-06
[network] Introduce relay + refactor cmd flags

dmtrskv:
https://github.com/NilFoundation/nil/issues/489
The cluster should be available for libp2p connections from the public network (mainly from archive nodes).
Here is the first part of the implementation: adding a relay service.
The service is very basic, and, apart from basic stuff like logging and telemetry, only sets itself up as a libp2p relay service and waits to be killed.
I'm going to test it with cluster nodes that explicitly connect to it as a relay (also, auto-relay feature is implemented in libp2p, but I'm not sure about its feature of advertising relayed nodes). It should be enough for the initial setup.
Next PR will present the working setup.

In order to reduce boilerplate and reuse similar cli interfaces, I've introduced `cobrax` and functions in `telemetry` and `network` that add common flags to the cli args.
This PR is mostly the refactoring, and `relay` mainly serves as an example of how the result is used to compose a new service in a cleaner code.

### DIFF
--- a/nil/internal/network/config.go
+++ b/nil/internal/network/config.go
@@ -10,10 +10,14 @@ type PeerID = peer.ID
 type Config struct {
 	PrivateKey PrivateKey `yaml:"-"`
 
+	KeysPath string `yaml:"keysPath,omitempty"`
+
 	Prefix      string `yaml:"prefix,omitempty"`
 	IPV4Address string `yaml:"ipv4,omitempty"`
 	TcpPort     int    `yaml:"tcpPort,omitempty"`
 	QuicPort    int    `yaml:"quicPort,omitempty"`
+
+	Relay bool `yaml:"relay,omitempty"`
 
 	DHTEnabled        bool          `yaml:"dhtEnabled,omitempty"`
 	DHTBootstrapPeers AddrInfoSlice `yaml:"dhtBootstrapPeers,omitempty"`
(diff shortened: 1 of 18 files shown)


nil/internal/network/config.go
```diff
@@ -1,32 +1,50 @@
 package network
 
 import (
+	"github.com/NilFoundation/nil/nil/common/check"
 	dht "github.com/libp2p/go-libp2p-kad-dht"
 	"github.com/libp2p/go-libp2p/core/peer"
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
```
olegrok, line 8: I'd suggest to move it into your cobrax module.

IMO it's a bit strange to have CLI dependencies in our indernal/core modules.

nil/internal/telemetry/telemetry.go
```diff
@@ -21,6 +22,10 @@ func NewDefaultConfig() *Config {
 	}
 }
 
+func AddFlags(fset *pflag.FlagSet, config *Config) {
```
olegrok, line 25: The same

nil/internal/network/config.go
dmtrskv: moved cobrax to internal (common is a weird package really)
moved these flags to cmdflags (so that cobrax doesn't depend on network and telemetry)

PR #491, opened 2025-03-06
[pubsub] Disable flaky test until better solution

dmtrskv:
The pub-sub test is extremely prone to failing in CI, yet often passes locally.
Example: https://github.com/NilFoundation/nil/actions/runs/13705312486/job/38329100031?pr=488
Here I disable the "complex scenario" but keep the code so that it can be run locally if desired.

This way of testing libp2p really makes little sense. At the basic level, libp2p works. Complex cases can only work on a real network anyway.

### DIFF
--- a/nil/internal/network/pubsub_test.go
+++ b/nil/internal/network/pubsub_test.go
@@ -94,6 +94,9 @@ func (s *PubSubSuite) TestTwoHosts() {
 }
 
 func (s *PubSubSuite) TestComplexScenario() {
+	// todo: this test often fails in CI but works locally
+	s.T().SkipNow()
+
 	const n = 5
 	const centralHost = 3
 


nil/internal/network/pubsub_test.go
```diff
@@ -94,6 +94,9 @@ func (s *PubSubSuite) TestTwoHosts() {
 }
 
 func (s *PubSubSuite) TestComplexScenario() {
+	// todo: this test often fails in CI but works locally
+	s.T().SkipNow()
```
olegrok, line 98: Don't it disable whole test suite?

dmtrskv: It would have been weird.
`Execution will continue at the next test or benchmark`

olegrok: Yes. But from Go point of view we run only one test here. And it's 
https://github.com/NilFoundation/nil/blob/8c1e8cc4b9c73ec55465695fcb1dac60c741d5fd/nil/internal/network/pubsub_test.go#L230

olegrok: But it should work https://github.com/stretchr/testify/issues/473#issuecomment-522370318

olegrok (approved):

PR #504, opened 2025-03-07
[execution] increase default balance for default faucets

olegrok:
Balance before: ~10^14
Default gas price: ~10^7
Gas to process one faucet request: ~10^4

That mean faucet can perform only 1k transactions. That's extremely small even for our test environment. So let's increase default balance of faucets in 10^6 times.


avm (approved):
I wonder if we need a constant for a unit of currency commeasurable with gas prices, like gwei in ETH...

PR #506, opened 2025-03-07
Implement check for rollup contracts

AndreyMlashkin:


nix/rollup-bridge-contracts.nix
```diff
@@ -45,8 +45,11 @@ stdenv.mkDerivation rec {
     npx hardhat clean && npx hardhat compile
   '';
 
+  doCheck = enableTesting;
```
akokoshn, line 48: seems it useless till we have no unit tests

AndreyMlashkin: fixed

PR #510, opened 2025-03-08
refactor(synccommittee): Use `execution.StateDB` instead of custom vm.StateDB

x-mass:
Refactored the tracer to use `execution.StateDB` instead of a custom `vm.StateDB` implementation. This change enables comparison between the original block hash and the traced block hash, ensuring all operations remain consistent.

Key changes:
- Introduced `OnTxStart` and `OnTxEnd` hooks in the tracing package, moving all per-transaction tracer calls into these hooks.
- Modified the `GetContract` method in the debug API to include proof of absence when a contract does not exist. This proof is either the longest matched path with the queried contract path or the root node if no path matches at all. This ensures we can prove the absence of even the beginning of such a path in the trie. Includes adjustment to `mpt.BuildProof` function.
- Refactored the ZKEVM tracer to avoid relying on values unavailable in the EVM tracing hook context. To fetch the current code when needed, added a `Code()` method to `tracing.OpContext`.
- Significantly changed the MPT tracer: it now computes differences between tries by traversing a trie with the initial root and comparing it against a trie with the current root. This eliminates the need to track changed values during tracing—only the initial and current roots are required, along with the underlying database storing all accessed values.


nil/services/synccommittee/prover/tracer/evm_tracer.go
```diff
@@ -0,0 +1,210 @@
...
+
+	"github.com/NilFoundation/nil/nil/common"
+	"github.com/NilFoundation/nil/nil/internal/tracing"
+	"github.com/NilFoundation/nil/nil/internal/types"
+	"github.com/NilFoundation/nil/nil/internal/vm"
+)
+
+type transactionTraceContext struct {
+	evm *vm.EVM // EVM instance re-running current transaction
+	// code      []byte      // currently executed code
```
oclaw, line 14: remove?

nil/services/synccommittee/prover/tracer/evm_tracer.go
```diff
@@ -0,0 +1,210 @@
...
+	// codeHash  common.Hash // hash of this.code
+	rwCounter *RwCounter // inherited from TracerStateDB, sequential RW operations counter
```
oclaw, line 16: Looks like the comment is irrelevant now

nil/services/synccommittee/prover/tracer/evm_tracer.go
```diff
@@ -0,0 +1,210 @@
...
+
+	return nil
+}
+
+type Tracer struct {
+	stateDb   vm.StateDB
+	Traces    ExecutionTraces
+	Stats     *Stats
+	rwCounter *RwCounter // sequential RW operations counter
+	// Reinited for each transaction
```
oclaw, line 129: I do not see now where it is reset between txs now. Is it relevant info?

nil/services/synccommittee/prover/tracer/evm_tracer.go
```diff
@@ -0,0 +1,210 @@
...
+	txnTraceCtx *transactionTraceContext
+	txnCtr      uint
```
oclaw, line 131: curTxIdx?

nil/internal/vm/interpreter.go
```diff
@@ -183,6 +188,7 @@ func (in *EVMInterpreter) Run(contract *Contract, input []byte, readOnly bool) (
 			if err == nil {
 				return
 			}
+			var _ tracing.OpContext = callContext
```
oclaw, line 191: Is it used somehow?

nil/services/synccommittee/prover/tracer/evm_tracer.go
```diff
@@ -0,0 +1,210 @@
...
+			defer verifyIntegrity()
+
+			if err := t.txnTraceCtx.processOpcode(t.Stats, pc, op, gas, scope, returnData); err != nil {
+				err = fmt.Errorf("pc: %d opcode: %X, gas: %d, cost: %d, mem_size: %d bytes, stack: %d items, ret_data_size: %d bytes, depth: %d cause: %w",
+					pc, op, gas, cost, len(scope.MemoryData()), len(scope.StackData()), len(returnData), depth, err,
+				)
+
+				// tracer by default should not affect the code execution but since we only run code to collect the traces - we should know
+				// about any failure as soon as possible instead of continue running
+				panic(managedTracerFailureError{underlying: err})
```
oclaw, line 190: This code is now seem to be handled by default StateDB logic so we will not be able to differ tracer error from any other panic happened during reexecution. Is there any option to keep tracking tracer errors in new impl?

nil/services/synccommittee/prover/tracer/evm_tracer.go
x-mass: Yep, it is done inside `OnTxEnd` hook

nil/services/synccommittee/prover/tracer/evm_tracer.go
x-mass: This error will be put into transaction execution result, we could determine the fact it was raised from tracer.
https://github.com/NilFoundation/nil/blob/2c3b38d8a6f4dcb21467f919dd56aae5b8591fd2/nil/internal/execution/state.go#L985

nil/internal/vm/interpreter.go
x-mass: No, removed

nil/services/synccommittee/prover/tracer/evm_tracer.go
oclaw: Yes, but now it is just a panic from the execution error. So some of our code in sync committee error handling may fail to determine the cause of an error. So I suggest keep extraction of the actual resaon inside tracer-related code (as it were before) instead of leaving it to the calling code
cc @zadykian

nil/services/synccommittee/prover/tracer/evm_tracer.go
oclaw: Maybe I got blind but where? :) 
Current code is 
```
		OnTxEnd: func(tx *types.Transaction, err types.ExecError) {
			if err := t.saveTransactionTraces(); err != nil {
				panic(err)
			}
			t.resetTxnTrace()
		}
```
And resetTxnTrace only sets context to nil. And rwCounter seems to be persistent between transactions

nil/services/synccommittee/prover/tracer/evm_tracer.go
x-mass: Added method `CollectErrors` which concats all errors occured during opcodes tracing

nil/services/synccommittee/prover/tracer/evm_tracer.go
x-mass: `rwCounter` is incremented from tracers. There was an issue with `curTxIdx` not incremented, though.

nil/services/synccommittee/prover/tracer/evm_tracer.go
```diff
@@ -0,0 +1,209 @@
...
+
+			// debug-only: ensure that tracer impl did not change any data from the EVM context
+			verifyIntegrity := assertEVMStateConsistent(pc, scope, returnData)
+			defer verifyIntegrity()
+
+			if err := t.txnTraceCtx.processOpcode(t.Stats, pc, op, gas, scope, returnData); err != nil {
+				err = fmt.Errorf("pc: %d opcode: %X, gas: %d, cost: %d, mem_size: %d bytes, stack: %d items, ret_data_size: %d bytes, depth: %d cause: %w",
+					pc, op, gas, cost, len(scope.MemoryData()), len(scope.StackData()), len(returnData), depth, err,
+				)
+				t.tracingErrors = append(t.tracingErrors, err)
```
oclaw, line 182: We still need to interrupt tracing here. If we have encountered an error during tracing once - all the following trace might be screwed

x-mass: Now we keep the first ocurred error and skip further tracing if it is not `nil`. If there was a panic after that error, we still include it into resulting error.

nil/internal/execution/state.go
```diff
@@ -313,13 +323,37 @@ func NewExecutionState(tx any, shardId types.ShardId, params StateParams) (*Exec
 	return res, res.initTries()
 }
 
+type DbContractAccessor struct {
+	*ContractTrie
+}
+
+func (ca *DbContractAccessor) GetContract(addr types.Address) (*types.SmartContract, error) {
+	ret, err := ca.Fetch(addr.Hash())
+	return ret, err
```
olegrok, line 332:
```suggestion
	return ca.Fetch(addr.Hash())
```

nil/services/synccommittee/prover/tracer/types.go
```diff
@@ -5,19 +5,20 @@ import (
...
-	AddExpOps(ops []ExpOp)
-	AddKeccakOps(ops []KeccakBuffer)
-	AddZKEVMStates(states []ZKEVMState)
-	AddCopyEvents(events []CopyEvent)
-	AddContractBytecode(addr types.Address, code []byte)
-	SetMptTraces(mptTraces *mpttracer.MPTTraces)
-}
-
-type executionTracesImpl struct {
+// type ExecutionTraces interface {
```
olegrok, line 8: Let's drop if it's unused

nil/internal/execution/state.go
```diff
@@ -611,6 +632,7 @@ func (es *ExecutionState) GetState(addr types.Address, key common.Hash) (common.
 }
 
 func (es *ExecutionState) SetState(addr types.Address, key common.Hash, val common.Hash) error {
+	// nonsense already here
```
olegrok, line 635: nonsense?

nil/services/synccommittee/prover/tracer/traces_collector.go
```diff
@@ -0,0 +1,391 @@
...
+	// Currently, MPT traces will contain only the last traced block. Since there is no MPT circuit yet, it's not a big deal.
+	tc.initMptTracer(shardId, prevBlock.Id, prevBlock.SmartContractsRoot)
+	// Initialize MPT tracer on first call
+	// if tc.mptTracer == nil {
+	// 	tc.initMptTracer(prevBlock.Id, prevBlock.SmartContractsRoot)
+	// }
+	// if tc.mptTracer.RootHash() != prevBlock.SmartContractsRoot {
+	// 	return nil, fmt.Errorf("MPT tracer root hash (%s) doesn't match previous block's smart contracts root (%s)",
+	// 		tc.mptTracer.RootHash(), prevBlock.SmartContractsRoot)
+	// }
```
shermike, line 196: Commented code

nil/services/synccommittee/prover/tracer/types.go
```diff
@@ -31,50 +19,68 @@ type executionTracesImpl struct {
...
+func (tr *ExecutionTraces) SetMptTraces(mptTraces *mpttracer.MPTTraces) {
 	tr.MPTTraces = mptTraces
 }
 
+// Append adds `other` to the end of traces slices, adds kv pairs from `otherTrace` maps
+func (tr *ExecutionTraces) Append(other *ExecutionTraces) {
+	if tr.MPTTraces != nil {
+		panic("you should not merge MPT traces, call `SetMptTraces` once at the end")
+	}
+	tr.MPTTraces = other.MPTTraces
```
akokoshn, line 69: It's merge MPT traces or replace?

nil/services/synccommittee/prover/tracer/traces_collector.go
```diff
@@ -0,0 +1,383 @@
...
+	}
+
+	// Handle genesis block
+	if currentDbgBlock.Id == 0 {
+		// TODO: prove genesis block generation?
+		return nil, ErrCantProofGenesisBlock
+	}
+
+	// Ensure blocks are sequential
+	if tc.lastTracedBlock != nil && currentDbgBlock.Id != *tc.lastTracedBlock+1 {
```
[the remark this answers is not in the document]
x-mass: Because MPT tracer requires state root at the beginning of the block `N+1` to be the same as at the end of block `N`. So we could get a single MPT traces for all processed blocks within a shard

nil/services/synccommittee/prover/tracer/types.go
x-mass: It is the way to ensure we don't miss any MPT traces (`SetMptTraces` is called at the end, not in the middle of tracing), and that containing traces for 
Not clear how MPT traces will look at final version, I decided not to spend time on merging MPT traces from different blocks. Perhaps we will abandon the idea of having a single MPT trace for multiple blocks in favor of per-block traces.

nil/internal/execution/state.go
```diff
@@ -787,7 +805,10 @@ func (es *ExecutionState) ContractExists(address types.Address) (bool, error) {
 		(storageRoot != common.EmptyHash), nil // non-empty storage
 }
 
-func (es *ExecutionState) AddInTransactionWithHash(transaction *types.Transaction, hash common.Hash) {
+func (es *ExecutionState) AddInTransactionWithHash(
+	transaction *types.Transaction,
+	hash common.Hash,
+) {
```
shermike, line 811: Why this change? It seems that the line is no longer than 120 chars.

nil/internal/execution/state.go
```diff
@@ -948,18 +976,28 @@ func (es *ExecutionState) AddOutTransaction(
...
+	)
 
 	return txn, nil
 }
 
-func (es *ExecutionState) sendBounceTransaction(txn *types.Transaction, execResult *ExecutionResult) (bool, error) {
+func (es *ExecutionState) sendBounceTransaction(
+	txn *types.Transaction,
+	execResult *ExecutionResult,
+) (bool, error) {
```
shermike, line 1000: The same - line length is no longer than 120.
And actually the same thing is in many places below.

nil/internal/execution/state.go
```diff
@@ -1651,7 +1718,9 @@ func (es *ExecutionState) CalculateGasForwarding(initialAvailValue types.Value)
 		}
 		if !availValue.IsZero() {
 			// If there is some remaining value due to division inaccuracy, credit it to the first transaction.
-			remainingFwdTransactions[0].FeeCredit = remainingFwdTransactions[0].FeeCredit.Add(availValue)
+			remainingFwdTransactions[0].FeeCredit = remainingFwdTransactions[0].FeeCredit.Add(
+				availValue,
+			)
```
shermike, line 1723: The same, why this change?

nil/internal/execution/state.go
x-mass: I guess, these are leftovers from my experiments with auto formatter. Removed them.

shermike (approved):
Execution part LGTM.

akokoshn (approved):

oclaw (approved):

PR #511, opened 2025-03-09
Enable faucet service in RPC node

shermike:


nil/client/direct_client.go
```diff
@@ -27,16 +27,18 @@ type DirectClient struct {
...
 var _ Client = (*DirectClient)(nil)
 
-func NewEthClient(ctx context.Context, db db.ReadOnlyDB, nShards types.ShardId, txnPools map[types.ShardId]txnpool.Pool, logger zerolog.Logger) (*DirectClient, error) {
-	var err error
-	localShardApis := make(map[types.ShardId]rawapi.ShardApi)
-	for shardId := range nShards {
-		localShardApis[shardId], err = rawapi.NewLocalRawApiAccessor(shardId, rawapi.NewLocalShardApi(shardId, db, txnPools[shardId]))
-		if err != nil {
-			return nil, err
+func NewEthClient(ctx context.Context, db db.ReadOnlyDB, nShards types.ShardId, txnPools map[types.ShardId]txnpool.Pool, logger zerolog.Logger, localApi *rawapi.NodeApiOverShardApis) (*DirectClient, error) {
```
olegrok, line 30: Is it possible to move creation of localApi outside this function.
E.g. `txnPools` and `nShards` arguments are only used for this purpose.

shermike: Done

nil/services/nilservice/service.go
```diff
@@ -462,7 +462,7 @@ func CreateNode(ctx context.Context, name string, cfg *Config, database db.DB, i
 
 			var cl client.Client
 			if cfg.Cometa != nil || cfg.IsFaucetApiEnabled() {
-				cl, err = client.NewEthClient(ctx, database, types.ShardId(cfg.NShards), txnPools, logger)
+				cl, err = client.NewEthClient(ctx, database, logger, rawApi)
```
olegrok, line 465: We usually pass logger as the last argument

olegrok (approved):

PR #514, opened 2025-03-10
Blockchain indexer RPC

Rolaman:
Changes:

- RPC: get address actions with paging
- Choose DB: badger or clickhouse
- Renaming: exporter to indexer

The main goal is to support transaction indexer to return all transfers and other actions by address. Can be used by sandbox, explorer and wallet


nil/client/direct_client.go
```diff
@@ -116,7 +116,29 @@ func (c *DirectClient) GetDebugBlocksRange(
 	fullTx bool,
 	batchSize int,
 ) ([]*jsonrpc.DebugRPCBlock, error) {
-	panic("Not supported")
```
olegrok, line 119: Why do you need it?

nil/cmd/exporter/fetch_block_test.go
```diff
@@ -1,4 +1,4 @@
-package internal
+package main
```
olegrok, line 1: That's pretty strange change why do you need it?

nil/cmd/nild/main.go
```diff
@@ -159,6 +160,7 @@ func parseArgs() *nildconfig.Config {
 	runCmd.Flags().StringVar(
 		&cfg.ValidatorKeysPath, "validator-keys-path", cfg.ValidatorKeysPath, "path to write validator keys")
 	runCmd.Flags().BoolVar(&cfg.EnableDevApi, "dev-api", cfg.EnableDevApi, "enable development API")
+	runCmd.Flags().StringVar(&cfg.IndexerConfig, "indexer-config", "", "path to Indexer config")
```
olegrok, line 163: Why do we need indexer inside nild?

nil/services/nilservice/service.go
```diff
@@ -41,7 +43,7 @@ import (
 const syncTimeoutFactor = 5
 
 func startRpcServer(
-	ctx context.Context,
+	parentCtx context.Context,
```
olegrok, line 46: What was wrong with a previous name?

nil/services/indexer/types/types.go
```diff
@@ -0,0 +1,83 @@
...
+
+type AddressActionKind uint8
+
+const (
+	SendEth AddressActionKind = iota
+	ReceiveEth
+	SmartContractCall
+)
+
+func (k AddressActionKind) String() string {
```
olegrok, line 30: You probably can use `stringer` for that

nil/services/indexer/types/types.go
```diff
@@ -0,0 +1,83 @@
...
+		return "SendEth"
+	case ReceiveEth:
+		return "ReceiveEth"
+	case SmartContractCall:
+		return "SmartContractCall"
+	}
+	panic("unknown AddressActionKind")
+}
+
+func (k *AddressActionKind) Set(input string) error {
```
olegrok, line 42: I'd suggest to use case-insensitive options

test-indexer-config.yml
```diff
@@ -0,0 +1,2 @@
+use-badger: true
```
olegrok, line 1: I guess we shouldn't commit this in the repository root.
You can add readme for specific command. Like it's done for the stresser: https://github.com/NilFoundation/nil/tree/main/nil/services/stresser/examples

nil/services/indexer/service_test.go
```diff
@@ -0,0 +1,191 @@
...
+	suite.Suite
+
+	service *Service
+	client  client.ClientMock
+	ctx     context.Context
+	dbPath  string
+}
+
+func (s *SuiteServiceTest) SetupTest() {
+	s.ctx = context.Background()
```
olegrok, line 27: Context should be cancelable. And you need to call `cancel` at `TearDownTest`

nil/services/indexer/service_test.go
```diff
@@ -0,0 +1,191 @@
...
+func (s *SuiteServiceTest) TearDownTest() {
+	// Clean up the temporary directory
+	err := os.RemoveAll(filepath.Dir(s.dbPath))
+	s.Require().NoError(err)
+}
+
+func (s *SuiteServiceTest) TestGetAddressActions() {
+	// Create test addresses
+	addr1 := types.Address{1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20}
+	addr2 := types.Address{2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21}
```
olegrok, line 54: Please use `HexToAddress` instead

nil/services/indexer/badger/badger.go
```diff
@@ -0,0 +1,453 @@
...
+				decoded:    receipt,
+				sszEncoded: sszEncodedBlock.Receipts[receiptIndex],
+			}
+		}
+
+		if current, exists := shardLatest[block.ShardId]; !exists || block.Block.Id > current {
+			shardLatest[block.ShardId] = block.Block.Id
+		}
+
+		key := makeBlockKey(block.ShardId, block.Block.Id)
```
olegrok, line 81: You can use ShardedTable abstraction

nil/services/indexer/badger/badger.go
```diff
@@ -0,0 +1,453 @@
...
+		value, err := json.Marshal(blocks[blockIndex])
```
olegrok, line 82: Why JSON, why not SSZ?

nil/services/indexer/badger/badger.go
```diff
@@ -0,0 +1,453 @@
...
+}
+
+func getTransactionStatus(receipt *types.Receipt) types2.AddressActionStatus {
+	if receipt.Success {
+		return types2.Success
+	}
+	return types2.Failed
+}
+
+func storeAddressAction(tx *badger.Txn, address types.Address, action *types2.AddressAction) error {
```
olegrok, line 162: We usually don't use raw badger. We use `db.*` (e.g. `db.RwTx`) abstractions

nil/services/indexer/service.go
```diff
@@ -0,0 +1,129 @@
...
+	"fmt"
+
+	"github.com/NilFoundation/nil/nil/client"
+	"github.com/NilFoundation/nil/nil/common/logging"
+	"github.com/NilFoundation/nil/nil/internal/db"
+	"github.com/NilFoundation/nil/nil/internal/types"
+	"github.com/NilFoundation/nil/nil/services/indexer/badger"
+	"github.com/NilFoundation/nil/nil/services/indexer/clickhouse"
+	driver2 "github.com/NilFoundation/nil/nil/services/indexer/driver"
+	types2 "github.com/NilFoundation/nil/nil/services/indexer/types"
```
olegrok, line 14: indexer_types or something similar. `types2` is a bit weird.

olegrok (changes requested):

nil/client/direct_client.go
Rolaman: Used by Indexer.FetchBlocks(...) method

olegrok: It should use RPC methods

nil/cmd/exporter/fetch_block_test.go
Rolaman: cause, there's main file in the same package. it is cmd

nil/cmd/nild/main.go
Rolaman: We want to be able to fetch indexed data via main nild json rpc

olegrok: It's very strange. What is motivation to do that?
What's wrong with a previous approach?

nil/services/indexer/types/types.go
Rolaman: as i see in nil repo, it is common practice to implement String() for this case

nil/services/indexer/service.go
Rolaman: agree, fixed

nil/services/indexer/service_test.go
Rolaman: agree, fixed

test-indexer-config.yml
Rolaman: removed

nil/services/indexer/types/types.go
Rolaman: done

nil/client/direct_client.go
Rolaman: yes, but indexer can be run in nild. in this case direct_client is used

niljs/src/clients/IndexerClient.ts
```diff
@@ -0,0 +1,48 @@
...
+  ReceiveEth = "ReceiveEth",
+  SmartContractCall = "SmartContractCall",
+}
+
+export enum AddressActionStatus {
+  Success = "Success",
+  Failed = "Failed",
+}
+
+export { IndexerClient };
```
KlonD90, line 48: it doesn't make sense. or export all things when it's defined like export enum or export all things in the end. now it's just mixed

niljs/src/clients/IndexerClient.ts
```diff
@@ -0,0 +1,48 @@
+import type { IAddress } from "../signers/types/IAddress.js";
+import type { Hex } from "../types/Hex.js";
+import { BaseClient } from "./BaseClient.js";
+import type { IClientBaseConfig } from "./types/Configs.js";
+
+class IndexerClient extends BaseClient {
+  // biome-ignore lint/complexity/noUselessConstructor: may be useful in the future
```
KlonD90, line 7: just add it in the future. skip it now

nil/client/direct_client.go
olegrok: Why do we need to embed it into nild at all?

Is it for development?

Rolaman: No, would like to integrate indexer API into nild.
Thread: https://nilfoundation.slack.com/archives/C03B7K8NZ46/p1743169690986629?thread_ts=1743147905.060269&cid=C03B7K8NZ46

it will be used by wallet and explorer clients

nil/client/direct_client.go
```diff
@@ -127,7 +149,29 @@ func (c *DirectClient) GetBlocksRange(
 	fullTx bool,
 	batchSize int,
 ) ([]*jsonrpc.RPCBlock, error) {
-	panic("Not supported")
+	if from >= to {
+		return nil, nil
+	}
+
+	result := make([]*jsonrpc.RPCBlock, 0)
+	for curBlockId := from; curBlockId < to; curBlockId += types.BlockNumber(batchSize) {
```
olegrok, line 157:
```suggestion
	for curBlockId := from; curBlockId < to; curBlockId ++ {
```
You can remove nested loop.

This part is relevant for http queries but here you don't send anything over network.

nil/client/direct_client.go
```diff
@@ -116,7 +116,29 @@ func (c *DirectClient) GetDebugBlocksRange(
 	fullTx bool,
 	batchSize int,
 ) ([]*jsonrpc.DebugRPCBlock, error) {
-	panic("Not supported")
+	if from >= to {
+		return nil, nil
+	}
+
+	result := make([]*jsonrpc.DebugRPCBlock, 0)
+	for curBlockId := from; curBlockId < to; curBlockId += types.BlockNumber(batchSize) {
```
olegrok, line 124: The same. You don't need nested loop here. Iterate just one by one

nil/services/indexer/badger/badger.go
```diff
@@ -0,0 +1,456 @@
...
+	receipts map[common.Hash]receiptWithSSZ,
+) error {
+	for _, txn := range block.InTransactions {
+		receipt, exists := receipts[txn.Hash()]
+		if !exists {
+			return fmt.Errorf("receipt not found for transaction %s", txn.Hash())
+		}
+
+		baseAction := indexertypes.AddressAction{
+			Hash:      txn.Hash(),
```
olegrok, line 127: It's better to save hash to variable once and use it several times. Hash calculation is quite heavy operation

nil/services/indexer/types/types.go
```diff
@@ -0,0 +1,83 @@
...
+	case ReceiveEth:
+		return "receiveeth"
+	case SmartContractCall:
+		return "smartcontractcall"
+	}
+	panic("unknown AddressActionKind")
+}
+
+func (k *AddressActionKind) Set(input string) error {
+	switch input {
```
olegrok, line 43:
```suggestion
	switch strings.ToLower(input) {
```

nil/services/nilservice/service.go
```diff
@@ -113,13 +115,33 @@ func startRpcServer(
 	}
 
 	if cfg.Cometa != nil {
-		cmt, err := cometa.NewService(ctx, cfg.Cometa, client)
+		cmt, err := cometa.NewService(indexerCtx, cfg.Cometa, client)
```
olegrok, line 118: It's pretty strange to pass `indexerCtx` to cometa

olegrok: Why did you rename it at all?

nil/client/direct_client.go
Rolaman: resolved in the thread

nil/cmd/nild/main.go
Rolaman: resolved above

nil/services/indexer/badger/badger.go
Rolaman: changed to db.*

nil/services/indexer/badger/badger.go
Rolaman: just to keep compatible with clickhouse and badger db

nil/services/indexer/badger/badger.go
Rolaman: done, via tables

nil/services/nilservice/service.go
Rolaman: fixed

niljs/src/clients/IndexerClient.ts
Rolaman: fixed

niljs/src/clients/IndexerClient.ts
Rolaman: removed

nil/client/direct_client.go
Rolaman: fixed

nil/services/indexer/badger/badger.go
Rolaman: fixed

nil/services/indexer/types/types.go
Rolaman: done

nil/services/indexer/badger/badger.go
```diff
@@ -0,0 +1,435 @@
...
+}
+
+func makeAddressActionTimestampKey(address types.Address, timestamp uint64) []byte {
+	key := make([]byte, len(address)+8)
+	copy(key[0:], address[:])
+	binary.BigEndian.PutUint64(key[len(address):], timestamp)
+	return key
+}
+
+func (b *BadgerDriver) FetchAddressActions(
```
olegrok, line 196: Is it possible to pass context here?

nil/services/indexer/service.go
```diff
@@ -0,0 +1,129 @@
...
+	c.DbPath = v.GetString("db-path")
+	c.DbName = v.GetString("db-name")
+	c.DbUser = v.GetString("db-user")
+	c.DbPassword = v.GetString("db-password")
+	return true
+}
+
+func NewService(ctx context.Context, cfg *Config, client client.Client) (*Service, error) {
+	s := &Service{
+		client: client,
```
olegrok, line 82: Is it used somewhere? Can't we remove it?

nil/services/nilservice/service.go
```diff
@@ -128,6 +130,26 @@ func startRpcServer(
...
 
+	if cfg.Indexer != nil {
+		idx, err := indexer.NewService(ctx, cfg.Indexer, client)
+		if err != nil {
+			return fmt.Errorf("failed to create indexer service: %w", err)
+		}
+		apiList = append(apiList, idx.GetRpcApi())
+
+		check.PanicIfErr(err)
+		go func() {
```
olegrok, line 141: Ideally you need to run it via concurrent module. As other services in nilservice run.

olegrok (commented):
Thanks for your changes. It's significantly better.

nil/services/nilservice/service.go
Rolaman: agree, changed to concurrent module

nil/services/indexer/service.go
Rolaman: agree, removed

nil/services/indexer/badger/badger.go
Rolaman: agree, added context

nil/services/nilservice/service.go
```diff
@@ -128,6 +130,29 @@ func startRpcServer(
...
+		check.PanicIfErr(err)
+		task := concurrent.MakeTask(
+			"indexer",
+			func(ctx context.Context) (err error) {
+				err2 := indexer.StartIndexer(ctx, &indexer.Cfg{
+					Client:        client,
+					IndexerDriver: idx.Driver,
+					BlocksChan:    make(chan *driver.BlockWithShardId, 1000),
+				})
+				return err2
```
olegrok, line 149:
```suggestion
				return indexer.StartIndexer(ctx, &indexer.Cfg{
					Client:        client,
					IndexerDriver: idx.Driver,
					BlocksChan:    make(chan *driver.BlockWithShardId, 1000),
				})
```

olegrok (approved):

PR #518, opened 2025-03-10
[GH-#] Make sure that the errors are valid UTF-8 strings

Un1oR:
Faced errors in the cluster log:
```
{
  "level": "error",
  "component": "network",
  "p2pIdentity": "16Uiu2HAmQ62Uh4CaACYPp3Lj9igQmg7F6ciSTszbMUzP238kEpv9",
  "protocolId": "/nil/shard/1/rawapi/GetFullBlockData",
  "error": "failed to pack Protobuf response: string field contains invalid UTF-8",
  "caller": "github.com/NilFoundation/nil/nil/internal/network/manager.go:199",
  "time": "2025-03-10T14:48:00Z",
  "message": "Failed to handle request"
}
```

### DIFF
--- a/nil/internal/execution/state.go
+++ b/nil/internal/execution/state.go
@@ -10,10 +10,12 @@ import (
 	"math"
 	"math/big"
 	"sort"
+	"unicode/utf8"
 
 	"github.com/NilFoundation/nil/nil/common"
 	"github.com/NilFoundation/nil/nil/common/assert"
 	"github.com/NilFoundation/nil/nil/common/check"
+	"github.com/NilFoundation/nil/nil/common/hexutil"
 	"github.com/NilFoundation/nil/nil/common/logging"
 	"github.com/NilFoundation/nil/nil/internal/abi"
 	"github.com/NilFoundation/nil/nil/internal/config"
@@ -1225,6 +1227,9 @@ func decodeRevertTransaction(data []byte) string {
 	var revString string
 	if index := bytes.IndexByte(data, 0); index > 0 {
 		revString = string(data[:index])
+		if !utf8.ValidString(revString) {
+			return "Not a UTF-8 string: " + hexutil.Encode(data[:index])
+		}
 	}
 	return revString
 }
--- a/nil/services/rpc/rawapi/pb/conversion.go
+++ b/nil/services/rpc/rawapi/pb/conversion.go
@@ -3,10 +3,12 @@ package pb
 import (
 	"encoding/binary"
 	"errors"
+	"unicode/utf8"
 
 	"github.com/NilFoundation/nil/nil/common"
 	"github.com/NilFoundation/nil/nil/common/check"
 	"github.com/NilFoundation/nil/nil/common/hexutil"
+	"github.com/NilFoundation/nil/nil/common/logging"
 	"github.com/NilFoundation/nil/nil/common/sszx"
 	"github.com/NilFoundation/nil/nil/internal/db"
 	"github.com/NilFoundation/nil/nil/internal/types"
@@ -14,6 +16,8 @@ import (
 	rpctypes "github.com/NilFoundation/nil/nil/services/rpc/types"
 	"github.com/NilFoundation/nil/nil/services/txnpool"
 )
+
+var Logger = logging.NewLogger("pb_conversion")
 
 // Hash converters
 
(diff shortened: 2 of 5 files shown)


olegrok (changes requested):
Still I'm not sure that we need to append bytecode to error message at all

Un1oR:
@olegrok what exactly are you suggesting? Completely remove the `revString` calculation block?

olegrok:
> @olegrok what exactly are you suggesting? Completely remove the `revString` calculation block?

I suggest to understand a reason why did it happen not just try to fix a symptom.

I added a test that reproduces "OutOfGasStorage" error. But it passes. There is no any garbage at the end of error message - https://github.com/NilFoundation/nil/pull/519

Un1oR:
Ok, got you
But isn't it more valuable to fix the problem on the prod now, and look deeper into the causes later?

nil/internal/execution/state.go
```diff
@@ -1225,6 +1227,9 @@ func decodeRevertTransaction(data []byte) string {
 	var revString string
 	if index := bytes.IndexByte(data, 0); index > 0 {
 		revString = string(data[:index])
+		if !utf8.ValidString(revString) {
+			return hexutil.Encode(data[:index])
```
shermike, line 1231: Please add some prefix for better understanding that it is not a utf string.
```suggestion
			return "Not a UTF string: "+hexutil.Encode(data[:index])
```

shermike:
Also, please add the following regression test:
```diff
diff --git a/nil/contracts/solidity/tests/Test.sol b/nil/contracts/solidity/tests/Test.sol
index 4c31fc81..d9128890 100644
--- a/nil/contracts/solidity/tests/Test.sol
+++ b/nil/contracts/solidity/tests/Test.sol
@@ -13,6 +13,10 @@ contract Test is NilBase {
 
     constructor() payable {}
 
+    function garbageInRequire(bool f, string memory m) public pure {
+        require(f, m);
+    }
+
     function emitEvent(uint a, uint b) public payable {
         emit testEvent(a, b);
     }
diff --git a/nil/tests/basic/basic_test.go b/nil/tests/basic/basic_test.go
index 89229c63..21543ffa 100644
--- a/nil/tests/basic/basic_test.go
+++ b/nil/tests/basic/basic_test.go
@@ -978,6 +978,24 @@ func (s *SuiteRpc) TestBatch() {
        s.Require().ErrorContains(err, "batch limit 100 exceeded")
 }
 
+func (s *SuiteRpc) TestRequire() {
+       abi, err := contracts.GetAbi(contracts.NameTest)
+       s.Require().NoError(err)
+
+       payload := contracts.GetDeployPayload(s.T(), contracts.NameTest)
+
+       addr, receipt := s.DeployContractViaMainSmartAccount(2, payload, tests.DefaultContractValue)
+       s.Require().True(receipt.AllSuccess())
+
+       data := []byte{1, 2, 3, 4, 5, 6, 7, 8, 9, 0}
+
+       calldata := s.AbiPack(abi, "garbageInRequire", false, string(data))
+
+       receipt = s.SendExternalTransactionNoCheck(calldata, addr)
+
+       s.Require().False(receipt.AllSuccess())
+}
```
Replace `TestRequire` with something that checks non-utf message.

olegrok:
I also have a test that reproduces the same issue - https://github.com/NilFoundation/nil/pull/519

Un1oR:
Added a test and did the processing when converting to Protobuf surviving invalid data.

nil/services/rpc/rawapi/pb/conversion.go
```diff
@@ -15,6 +17,8 @@ import (
 	"github.com/NilFoundation/nil/nil/services/txnpool"
 )
 
+var Logger = logging.NewLogger("pc_conversion")
```
olegrok, line 20: pc or pb?

nil/services/rpc/rawapi/pb/conversion.go
```diff
@@ -199,6 +203,10 @@ func (e *Error) PackProtoMessage(err error) *Error {
 func packErrorMap(errors map[common.Hash]string) map[string]*Error {
 	result := make(map[string]*Error, len(errors))
 	for key, value := range errors {
+		if !utf8.ValidString(value) {
+			Logger.Error().Str("key", key.String()).Msg("invalid UTF-8 string in error map")
```
olegrok, line 207:
```suggestion
			Logger.Error().Stringer("key", key).Msg("invalid UTF-8 string in error map")
```

nil/services/rpc/rawapi/pb/conversion.go
```diff
@@ -199,6 +203,13 @@ func (e *Error) PackProtoMessage(err error) *Error {
 func packErrorMap(errors map[common.Hash]string) map[string]*Error {
 	result := make(map[string]*Error, len(errors))
 	for key, value := range errors {
+		if !utf8.ValidString(value) {
+			Logger.Error().
+				Stringer("key", key).
+				Str("value", hexutil.Encode([]byte(value))).
```
olegrok, line 209:
```suggestion
				Hex("value", []byte(value)).
```

olegrok (approved):

PR #519, opened 2025-03-10
test: add a test that checks correctness of OutOfGasStorage error

olegrok:
Just to be sure that contract returned valid output.


olegrok:
Suppressed by https://github.com/NilFoundation/nil/pull/518

PR #521, opened 2025-03-11
[go-ibft] add more metrics for specific stages

olegrok:
This patch adds histograms for "new round", "commit" and "prepare" stages. There is also "fin" stage. But it seems to be empty.


Zerg1996 (approved):
Why don't you want to use measurer like in block_generation duration?
```
	measurer, err := telemetry.NewMeasurer(
		meter, "block_generation", telattr.ShardId(shardId),
	)
```

But as hist implementation for me looks good!

olegrok:
> Why don't you want to use measurer like in block_generation duration?

I'd like to keep `go-ibft` implementation independent on our internal packages. At least for now.

PR #523, opened 2025-03-11
[tests] disable consensus for synccommittee test to prevent CI failures

olegrok:
For some reasons we see following failures in CI:
```
2025-03-11T08:47:24.4638935Z nil> 2025-03-11 08:47:09 DBG [consensus]         ../../../../go-ibft/core/ibft.go:444 > block proposal accepted shardId=[1]
2025-03-11T08:47:24.4639316Z nil> 2025-03-11 08:47:09 DBG [consensus]         ../../../../go-ibft/core/ibft.go:448 > pre-prepare message multicasted shardId=[2]
2025-03-11T08:47:24.4639642Z nil> 2025-03-11 08:47:09 DBG [consensus]         ../../../../go-ibft/core/ibft.go:840 > enter: prepare state shardId=[2]
2025-03-11T08:47:24.4639763Z nil> signal: killed
2025-03-11T08:47:24.4640119Z nil> FAIL       github.com/NilFoundation/nil/nil/services/synccommittee/prover/tracer   232.386s
2025-03-11T08:47:24.4640588Z nil> ok         github.com/NilFoundation/nil/nil/services/synccommittee/prover/tracer/internal/mpttracer        1.405s
```

Seems test consumes a lot of resourses. Probably it should be discovered. But for now the best option is to disable consensus.

### DIFF
--- a/nil/services/synccommittee/core/service_test.go
+++ b/nil/services/synccommittee/core/service_test.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/NilFoundation/nil/nil/common"
 	"github.com/NilFoundation/nil/nil/common/logging"
-	"github.com/NilFoundation/nil/nil/internal/collate"
 	"github.com/NilFoundation/nil/nil/internal/db"
 	"github.com/NilFoundation/nil/nil/services/nilservice"
 	rpctest "github.com/NilFoundation/nil/nil/services/rpc"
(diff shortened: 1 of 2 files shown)


nil/services/synccommittee/core/service_test.go
```diff
@@ -39,8 +38,8 @@ func (s *SyncCommitteeTestSuite) SetupSuite() {
 	nilserviceCfg := &nilservice.Config{
 		NShards:              s.nShards,
 		HttpUrl:              s.url,
-		Topology:             collate.TrivialShardTopologyId,
-		CollatorTickPeriodMs: 100,
+		CollatorTickPeriodMs: 200,
```
dmtrskv, line 41: btw have you tried changing 100 to 200 in `PatchConfigWithTestDefaults`? it will work for all rpc tests

dmtrskv (approved):

nil/services/synccommittee/core/service_test.go
olegrok: No need to change it everywhere. 
Here the main idea is to decrease block production rate to reduce SC tracer memory pressure

PR #526, opened 2025-03-11
implement basic geth test

AndreyMlashkin:
implement basic geth test


nix/rollup-bridge-contracts.nix
```diff
@@ -45,9 +48,38 @@ stdenv.mkDerivation rec {
     npx hardhat clean && npx hardhat compile
   '';
 
+  doCheck = enableTesting;
+  checkPhase = ''
+    export WALLET_ADDRESS=0xc8d5559BA22d11B0845215a781ff4bF3CCa0EF89
```
akokoshn, line 53: not sure about hardcoded these values here

akokoshn (changes requested):
This functionality should be checked on nil_dev environment.
The CI is good place for run unit tests.

nix/rollup-bridge-contracts.nix
AndreyMlashkin: Corrected

PR #531, opened 2025-03-11
relayer: event listener impl

oclaw:
**For scared ones**: 1.5k LoC of changes here are just .abi.json content for the L1BridgeMessenger contract that should be generated automatically in future :)

This PR adds an initial version of the relayer service. 

By spec the service is responsible for:
- Fetching ongoing and historical events from the L1BridgeMessenger contract deployed to the L1 (`EventListener` component)
- Monitoring L1 block finalization (`FinalityEnsurer` component)
- Signing and forwarding finalized L1 events to the L2 (`TransactionSender` component)

Initial version of the service is a `sync_committee_relayer` binary with `EventListener` component implementation

Here is brief description of the component functionality:
- Connect to L1
- Check which historical events has to be fetched (based on stored last processed block number) and fetch them
- Subscribe to new events receiving
- Save every fetched event to the database (BadgerDB) with updating reference to the last processed block number

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,7 +8,7 @@ GOBUILD = GOPRIVATE="$(GOPRIVATE)" $(GO) build $(GO_FLAGS)
 GO_DBG_BUILD = GOPRIVATE="$(GOPRIVATE)" $(GO) build -tags $(BUILD_TAGS),debug,assert -gcflags=all="-N -l"  # see delve docs
 GOTEST = GOPRIVATE="$(GOPRIVATE)" GODEBUG=cgocheck=0 $(GO) test -tags $(BUILD_TAGS),debug,assert,test $(GO_FLAGS) ./... -p 2
 
-SC_COMMANDS = sync_committee sync_committee_cli proof_provider prover nil_block_generator
+SC_COMMANDS = sync_committee sync_committee_cli proof_provider prover nil_block_generator relayer
 COMMANDS += nild nil nil_load_generator exporter cometa faucet journald_forwarder relay $(SC_COMMANDS)
 
 all: $(COMMANDS)
--- a/nil/internal/db/badger.go
+++ b/nil/internal/db/badger.go
@@ -51,12 +51,17 @@ type BadgerIter struct {
 	toPrefix    []byte
 }
 
+type BadgerSequence struct {
+	seq *badger.Sequence
+}
+
 // interfaces
 var (
-	_ RoTx = new(BadgerRoTx)
-	_ RwTx = new(BadgerRwTx)
-	_ DB   = new(badgerDB)
-	_ Iter = new(BadgerIter)
+	_ RoTx     = new(BadgerRoTx)
+	_ RwTx     = new(BadgerRwTx)
+	_ DB       = new(badgerDB)
+	_ Iter     = new(BadgerIter)
+	_ Sequence = new(BadgerSequence)
 )
 
 func MakeTablePrefix(table TableName) string {
@@ -117,6 +122,14 @@ func (db *badgerDB) CreateRoTxAt(ctx context.Context, ts Timestamp) (RoTx, error
 
 func (db *badgerDB) CreateRoTx(ctx context.Context) (RoTx, error) {
 	return db.createRoTx(ctx, db.db.NewTransaction(false), false)
+}
+
+func (db *badgerDB) GetSequence(ctx context.Context, key []byte, bandwidth uint64) (Sequence, error) {
+	seq, err := db.db.GetSequence(key, bandwidth)
+	if err != nil {
+		return nil, err
+	}
+	return &BadgerSequence{seq: seq}, nil
 }
 
 func (db *badgerDB) createRwTx(_ context.Context, txn *badger.Txn) (RwTx, error) {
@@ -299,4 +312,8 @@ func (it *BadgerIter) Next() ([]byte, []byte, error) {
 
 func (it *BadgerIter) Close() {
 	it.iter.Close()
+}
+
+func (seq *BadgerSequence) Next() (uint64, error) {
+	return seq.seq.Next()
 }
--- a/nil/services/relayer/internal/l1/contract_wrapper.go
+++ b/nil/services/relayer/internal/l1/contract_wrapper.go
@@ -1,0 +1,77 @@
+package l1
+
+import (
+	"context"
+
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/event"
+)
+
+type L1Contract interface {
+	SubscribeToEvents(ctx context.Context, sink chan<- *L1MessageSent) (event.Subscription, error)
+	GetEventsFromBlockRange(ctx context.Context, from uint64, to *uint64) ([]*L1MessageSent, error)
+}
+
+type l1ContractWrapper struct {
+	impl   *L1
+	l2Addr common.Address
+}
+
+var _ L1Contract = (*l1ContractWrapper)(nil)
+
+func NewL1ContractWrapper(ethClient EthClient,
+	l1ContractAddr, l2ConractAddr string,
+) (*l1ContractWrapper, error) {
+	addr := common.HexToAddress(l1ContractAddr)
+	impl, err := NewL1(addr, ethClient)
+	if err != nil {
+		return nil, err
+	}
+
+	return &l1ContractWrapper{
+		impl:   impl,
+		l2Addr: common.HexToAddress(l2ConractAddr),
+	}, nil
+}
+
+func (w *l1ContractWrapper) SubscribeToEvents(
+	ctx context.Context,
+	sink chan<- *L1MessageSent,
+) (event.Subscription, error) {
+	return w.impl.WatchMessageSent(
+		&bind.WatchOpts{Context: ctx},
+		sink,
+		nil,                        // any sender (for now)
+		[]common.Address{w.l2Addr}, // destination is the contract this relayer is bound to
+		nil,                        // any nonce
+	)
+}
+
+func (w *l1ContractWrapper) GetEventsFromBlockRange(
+	ctx context.Context,
+	from uint64,
+	to *uint64,
+) ([]*L1MessageSent, error) {
+	iter, err := w.impl.FilterMessageSent(
+		&bind.FilterOpts{
+			Start: from,
+			End:   to,
+		},
+		nil,                        // any sender (for now)
+		[]common.Address{w.l2Addr}, // destination is the contract this relayer is bound to
+		nil,                        // any nonce
+	)
+	if err != nil {
+		return nil, err
+	}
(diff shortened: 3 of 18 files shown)


nil/services/synccommittee/internal/storage/common_storage.go
```diff
@@ -8,25 +8,25 @@ import (
 	"github.com/rs/zerolog"
 )
 
-type commonStorage struct {
-	database    db.DB
-	retryRunner common.RetryRunner
-	logger      zerolog.Logger
+type CommonStorage struct {
```
oclaw, line 11: I do not want to store my types inside common sync committee folder but I want to use this struct :)
It is anyway is unavailable outside the sync_committee project (due to internal folder) so fields should be public

nil/internal/db/badger.go
```diff
@@ -118,6 +123,14 @@ func (db *badgerDB) CreateRoTx(ctx context.Context) (RoTx, error) {
 	return db.createRoTx(ctx, db.db.NewTransaction(false), false)
 }
 
+func (db *badgerDB) GetSequence(ctx context.Context, key []byte, bandwidth uint64) (Sequence, error) {
+	seq, err := db.db.GetSequence(key, bandwidth)
```
oclaw, line 127: I have to keep insertion history due to avoid invalid order of events passed to L2 so Badger's sequencer seems to be a good tool to use here

nil/services/synccommittee/relayer/internal/l1/event_listener.go
```diff
@@ -0,0 +1,380 @@
...
+	if cfg.BatchSize == 0 {
+		return errors.New("empty batch size for fetching old events")
+	}
+	if cfg.PollInterval == 0 {
+		return errors.New("empty poll interval for fetching old events")
+	}
+	return nil
+}
+
+type EventListener struct {
```
oclaw, line 46: Main object with business logic in this PR. It has 3 subroutines inside (subscriber, fetcher, eventProcessor)

nil/services/synccommittee/relayer/internal/l1/contract_wrapper.go
```diff
@@ -0,0 +1,69 @@
...
+
+import (
+	"context"
+
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/event"
+)
+
+type L1Contract interface {
```
oclaw, line 11: Otherwise I will have to mock generated by `abigen` tool code which is pain

nil/services/synccommittee/relayer/internal/l1/eth_client.go
```diff
@@ -0,0 +1,11 @@
+package l1
+
+import (
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+)
+
+type EthClient interface {
```
oclaw, line 7: For mock generation

nil/services/synccommittee/relayer/internal/l1/finality_ensurer.go
```diff
@@ -0,0 +1,29 @@
+package l1
+
+import (
+	"context"
+	"errors"
+)
+
+type FinalityEnsurer struct {
```
oclaw, line 8: Just a placeholder for future

nil/services/synccommittee/relayer/internal/l1/l1_bridge_messenger_abi.json
oclaw, line 1: Not for review, copypasted generated code

nil/services/synccommittee/relayer/internal/l1/types.go
```diff
@@ -0,0 +1,46 @@
...
+	BlockNumber uint64      `json:"blkNum"`
+	BlockHash   common.Hash `json:"blkHash"`
+
+	// Used for proper ordering events while sending to L2
+	// Assigned locally (and sequentially for each fetched from the L1 event)
+	// Does not guarantee order for events collected by different relayer instances
+	SequenceNumber uint64 `json:"sequenceNumber"`
+
+	// Payload
+	Sender             common.Address `json:"sender"`
```
oclaw, line 23: Copied from L1BridgeMessenger contract event data

nil/services/synccommittee/relayer/internal/l2/transaction_sender.go
```diff
@@ -0,0 +1,21 @@
+package l2
+
+import (
+	"context"
+	"errors"
+)
+
+type TransactionSender struct {
```
oclaw, line 8: Just a placeholder for future

Makefile
```diff
@@ -8,7 +8,7 @@ GOBUILD = GOPRIVATE="$(GOPRIVATE)" $(GO) build $(GO_FLAGS)
 GO_DBG_BUILD = GOPRIVATE="$(GOPRIVATE)" $(GO) build -tags $(BUILD_TAGS),debug,assert -gcflags=all="-N -l"  # see delve docs
 GOTEST = GOPRIVATE="$(GOPRIVATE)" GODEBUG=cgocheck=0 $(GO) test -tags $(BUILD_TAGS),debug,assert,test $(GO_FLAGS) ./... -p 2
 
-SC_COMMANDS = sync_committee sync_committee_cli proof_provider prover nil_block_generator
+SC_COMMANDS = sync_committee sync_committee_cli proof_provider prover nil_block_generator sync_committee_relayer
```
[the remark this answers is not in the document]
oclaw: I wish I knew it before :D 
Will fix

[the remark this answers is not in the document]
oclaw: Done. But I have noticed that we already have a service named `relay` in nil/cmd

@dmtrskv Seems like we have a naming collision here. As far as I can see your service is a network-related infrastructural one. Is it possible to highlight it somehow in the binary name?

nil/services/relayer/internal/l1/event_listener.go
```diff
@@ -0,0 +1,381 @@
...
+			// TODO (oclaw) metrics
+		}
+		return err
+	})
+
+	// TODO(oclaw) metrics (this routine is not expected to run for too long, we should now if something is stuck here)
+}
+
+func (el *EventListener) fetchPastEvents(ctx context.Context, eventCh chan<- *L1MessageSent) error {
+	header, err := el.rawEthClient.HeaderByNumber(ctx, nil)
```
akokoshn, line 216: nice to have comment here, what actually return `el.rawEthClient.HeaderByNumber` with `nil`

nil/services/relayer/internal/l1/event_listener.go
```diff
@@ -0,0 +1,381 @@
...
+
+// Routine responsible for fetching historical blocks in range [lastProcessedBlock; latest)
+// until it is executed incoming event processing is shutdown
+func (el *EventListener) fetcher(ctx context.Context, eventCh chan<- *L1MessageSent) error {
+	defer close(eventCh) // no more events will be posted to the channel after routine finished its work
+
+	el.logger.Info().Msg("started fetcher")
+
+	// try fetching as long as possible, force exit after large enough attempt number
+	retrier := common.NewRetryRunner(
```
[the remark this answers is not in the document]
oclaw: IMO yes, it makes sense, because in case of failure here we are going to reinitialize whole listener (and call Unsubscribe eventually). So if failure here is a one-time glitch - it is better to try again in a fast way

nil/services/relayer/internal/l1/event_listener.go
oclaw: It is described in the go ethereum doc, but I agree, will add

oclaw: done

nil/services/relayer/internal/l1/contract_wrapper.go
```diff
@@ -0,0 +1,69 @@
...
+}
+
+func (w *l1ContractWrapper) SubscribeToEvents(ctx context.Context, sink chan<- *L1MessageSent) (event.Subscription, error) {
+	return w.impl.WatchMessageSent(
+		&bind.WatchOpts{Context: ctx},
+		sink,
+
+		// TODO(oclaw) do we need filters?
+		nil, // messageSender []common.Address,
+		nil, // messageTarget []common.Address,
```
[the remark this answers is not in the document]
oclaw: fixed

nil/services/relayer/internal/l1/event_listener.go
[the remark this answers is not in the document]
oclaw: About unifying these retries - I'd rather keep them separately for fine tuning purposes

nil/services/relayer/internal/l1/event_listener.go
```diff
@@ -0,0 +1,381 @@
...
+	if err != nil {
+		el.logger.Error().
+			Str("contract_addr", el.config.BridgeMessengerContractAddress).
+			Err(err).
+			Msg("failed to subscribe to updates from L1 contract")
+		return err
+
+		// TODO(oclaw) metrics
+	}
+	defer sub.Unsubscribe()
```
x-mass, line 161: The current subscription model does not handle L1 reorganization events. This can lead to a situation where a received block's parent hash does not match the hash of the previously received block. In such cases, we need to trace back through the blocks until we find the divergence point and then rollback any processed results from that point onward. 
Alternatively (imho, way easier), we could wait for block finality (no reorg could happen if finality is reached) before processing new blocks. From what I understand, subscriptions don't support this directly. However, we could consider switching to a long-polling model and periodically query the `rpc.FinalizedBlockNumber` to manage this. 
Finalization takes ~16 minutes in Ethereum.

oclaw: We actually do exactly what you suggesting in an another component - finality ensurer. 
IMO approach when we have to repeatedly re-fetch all events from block and mix up fetching and finalization check logic is less reliable as it results in bigger amount of syscalls that might fail due to external reasons. 
Also fetching these events earlier gives us an opportunity for adding metrics/logs/doing some preparations

oclaw: To sum up - this component is only responsible for ordered storage of all incoming events. Making decision if the event A is to be forwarded to the nil chain is not a part of the listener's functionality

akokoshn (approved):

nil/services/relayer/internal/l1/event_listener.go
```diff
@@ -0,0 +1,381 @@
...
+		el.setCurrentProcessingBlock(latestBlock, header.Hash())
+	}
+
+	el.logger.Info().
+		Uint64("latest_block_num", latestBlock).
+		Uint64("latest_processed_block_num", el.state.currentBlockNumber).
+		Msg("connected to Etherium")
+
+	if el.state.currentBlockNumber >= latestBlock {
+		el.logger.Info().Msg("no need to fetch old events")
```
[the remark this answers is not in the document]
oclaw: exactly, fixed, thx

PR #536, opened 2025-03-12
Sync Committee: Block Storage Refactoring

zadykian:
### Split block storage operations into dedicated components

* Introduced separate types (`batchOp`, `batchCountOp`, `stateRootOp`, etc.) to represent isolated sets of operations on different data structures: blocks, batches, secondary indices, etc;
* Reduced the size of `block_storage.go` file from ~1100 lines to ~600;
* No functional changes introduced

### DIFF
--- a/nil/services/synccommittee/internal/storage/block_storage.go
+++ b/nil/services/synccommittee/internal/storage/block_storage.go
@@ -2,13 +2,8 @@ package storage
 
 import (
 	"context"
-	"encoding/binary"
-	"encoding/hex"
-	"encoding/json"
 	"errors"
 	"fmt"
-	"iter"
-	"time"
 
 	"github.com/NilFoundation/nil/nil/common"
 	"github.com/NilFoundation/nil/nil/common/logging"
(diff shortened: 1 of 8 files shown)


nil/services/synccommittee/internal/storage/block_storage.go
```diff
@@ -153,20 +70,7 @@ func (bs *BlockStorage) TryGetProvedStateRoot(ctx context.Context) (*common.Hash
...
-	if errors.Is(err, db.ErrKeyNotFound) {
-		return nil, nil
-	}
-	if err != nil {
-		return nil, err
-	}
-
-	hash := common.BytesToHash(hashBytes)
-	return &hash, nil
+	return newStateRootOp().getProvedStateRoot(tx)
```
akokoshn, line 73: can we keep APIs in `BlockStorage` instance instead call `New...` each time?

zadykian: Thanks, good suggestion
Embedded all `*Op` structs directly into the storage:

https://github.com/NilFoundation/nil/pull/536/files#diff-c9bfa9dc9eda5ca8c9dae2789f503d9eefd2f96167dfe03c5eb967fc52d2aa90R43-R50

akokoshn (approved):

PR #544, opened 2025-03-12
Add a nixos container that can run full cluster

knazarov:
This PR when merged will allow you to run cluster setup almost exactly like it does in production, in a matter of seconds. The word "container" here may be a bit misleading, because even though we do start a cgroup, it will run systemd, nild, nginx, and other things like a normal operating system (and unlike Docker for example).

Note, that you need NixOS installed and not just Nix package manager, because the `nixos-container` tool is not available with just Nix.

```
sudo nixos-container create devcontainer --flake .
sudo nixos-container start devcontainer
```

If you made changes to nild or configs and want to update the container in-place, you can do:

```
sudo nixos-container update devcontainer --flake .
```

As a follow-up, I will add ClickHouse, exporter, and all frontend parts to run in the container as well.

### DIFF
--- a/explorer_backend/services/sqlite.ts
+++ b/explorer_backend/services/sqlite.ts
@@ -1,7 +1,7 @@
 import { createHash } from "node:crypto";
 import BetterSqlite3 from "better-sqlite3";
 
-const db: BetterSqlite3.Database = new BetterSqlite3("./database.db");
+const db: BetterSqlite3.Database = new BetterSqlite3(process.env.EXPLORER_DB || "./database.db");
 
 export { db };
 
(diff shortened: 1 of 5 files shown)


Gezort:
>Note, that you need NixOS installed

tbh I doubt that this setup will be extensively used then...

Gezort (approved):

avm:
> > Note, that you need NixOS installed
> 
> tbh I doubt that this setup will be extensively used then...

Apparently it's possible to build a standard OCI container from nix and run the container with Docker or Podman. I'll look into that.

knazarov:
> > > Note, that you need NixOS installed
> > 
> > 
> > tbh I doubt that this setup will be extensively used then...
> 
> Apparently it's possible to build a standard OCI container from nix and run the container with Docker or Podman. I'll look into that.

I'm pretty sure they aren't equivalent as OCI containers don't run systemd. But you can try.

PR #549, opened 2025-03-13
[jsonrpc] implement web3_clientVersion

### LINKED ISSUE #541: add API handle to return validator version
may be useful during debugging

@olegrok suggest to implement https://docs.getblock.io/api-reference/ethereum-eth/web3_clientversion-ethereum

olegrok:
After this patch it will be possible to retrieve a version of current validator node.

Closes #541

### DIFF
--- a/nil/client/client.go
+++ b/nil/client/client.go
@@ -30,6 +30,7 @@ type BatchRequest interface {
 type Client interface {
 	RawClient
 	DbClient
+	Web3Client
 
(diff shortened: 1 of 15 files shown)


Gezort:
maybe we'll need to add this as CLI command later. wdyt?

Gezort (approved):

olegrok:
> maybe we'll need to add this as CLI command later. wdyt?

We can do that. I suggest to file an issue to clijs

PR #550, opened 2025-03-13
chore(ci): add `help` target to display all targets documentation

polymaer:
This PR add a  new  target `help` and documentation of every targets.

Run `make help` will output all targets documentation, which will help new developers contribute to this project easily.

image


knazarov:
Hi @polymaer ,
We don't accept trivial changes (including grammar and spelling). If you want to get something like that merged - open an issue instead.
For external contribution to be merged, it needs to be reasonably non-trivial and be in something functional.

PR #551, opened 2025-03-13
docs: contributing

Ocheretovich:
added an icon to the workflow


knazarov:
Hi @Ocheretovich ,
We don't accept pull requests with minor changes. If you want to contribute - the contribution should be reasonably non-trivial.
Otherwise open an issue instead.

PR #552, opened 2025-03-13
Add per-shard-pair incremental tx ids.

avm:
For each pair of (source, destination) shards, we keep track of the number of internal transactions ever sent from source to destination, and use it to assign incremental tx ids. This allows us to prove that we have processed all the transactions from a specific shard, in order, without missing any.

In this PR, we store outbound transaction counts in the block, as a mapping: `dstShardId ShardId -> txCount TransactionIndex`. We reuse the outbound transaction trie for this purpose, without storing zeroes. The transaction trie now stores two kinds of keys:
* TransactionIndex (uint64) keys for transaction data;
* ShardId (uint16) keys for transaction counts.

We don't store zero values, so the zerostate does not have to be changed in any way.

### DIFF
--- a/nil/internal/config/params.go
+++ b/nil/internal/config/params.go
@@ -241,10 +241,10 @@ func SetParamL1Block(c ConfigAccessor, params *ParamL1BlockInfo) error {
 	return setParamImpl(c, params)
 }
 
-func GetParamNShards(c ConfigAccessor) (uint32, error) {
+func GetParamNShards(c ConfigAccessor) (types.ShardId, error) {
 	param, err := getParamImpl[ParamGasPrice](c)
 	if err != nil {
 		return 0, err
 	}
-	return uint32(len(param.Shards)), nil
+	return types.ShardId(len(param.Shards)), nil
 }
--- a/nil/internal/execution/state.go
+++ b/nil/internal/execution/state.go
@@ -57,6 +57,8 @@ type RollbackParams struct {
 	SearchDepth uint32
 }
 
+type ShardTxCounts map[types.ShardId]types.TransactionIndex
+
 type ExecutionState struct {
 	tx                 db.RwTx
 	ContractTree       *ContractTrie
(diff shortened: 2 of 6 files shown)


nil/internal/execution/state.go
```diff
@@ -261,6 +264,19 @@ type StateParams struct {
 	FeeCalculator  FeeCalculator
 }
 
+func fetchShardTxIds(tx db.RoTx, shardId, nShards types.ShardId, root common.Hash) ShardTxCounts {
```
olegrok, line 267: Don't forget to remove [CollatorState](https://github.com/NilFoundation/nil/blob/f885c13eb1cb5b0a37b6ddfaca799246da978285/nil/internal/types/collator.go)-related structures

nil/internal/config/params.go
```diff
@@ -241,10 +241,10 @@ func SetParamL1Block(c ConfigAccessor, params *ParamL1BlockInfo) error {
 	return setParamImpl(c, params)
 }
 
-func GetParamNShards(c ConfigAccessor) (uint32, error) {
```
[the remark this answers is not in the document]
avm: I don't see how writing `uint32` is better than `ShardId`. Would you prefer another type alias called `ShardCount`?

nil/internal/execution/state.go
avm: Definitely, when I deal with the receiver side.

nil/internal/config/params.go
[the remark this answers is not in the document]
olegrok: Yes. ShardCount would be better.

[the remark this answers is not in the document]
avm: Added, but now we'll have to cast between those two obviously equivalent types forever. This makes me sad.

nil/internal/types/shard.go
```diff
@@ -8,7 +8,10 @@ import (
 )
 
 // 32 bits are more than enough while avoiding problems with marshaling 64-bit values as numbers in JSON.
-type ShardId uint32
+type (
+	ShardId    uint32
+	ShardCount ShardId
```
dmtrskv, line 13: If you say `type ShardCount = ShardId` they will be completely equivalent and you won't have to convert.

nil/internal/config/config.go
```diff
@@ -78,19 +78,21 @@ func NewConfigReader(tx db.RoTx, mainShardHash *common.Hash) (ConfigAccessor, er
 	return &configReader{trie}, nil
 }
 
-type ConfigAccessorStub struct{}
+type ConfigAccessorStub struct {
+	data map[string][]byte
```
dmtrskv, line 82: is it actually used?

PR #555, opened 2025-03-13
rename chain to shard

olegrok:
Just for consistency in the code. We don't use chain in public interfaces. Let's rename it in the code.


dmtrskv (approved):
Thank you for the contribution.

PR #559, opened 2025-03-13
[academy] : lending protocol example frontend

gitshreevatsa:
This PR adds a frontend for our lending protocol and enables users to transact via the frontend and explain how to build frontend for applictions built on =nil; leveraging nil.js and our wallet SDK. The frontend is built using react.js and provides a detailed walkthrough to help developers grasp how to build frontends with =nil;


academy/lending-protocol/frontend/README.md
```diff
@@ -0,0 +1,292 @@
...
+Before running the frontend application, you must deploy the lending pool contract and obtain its address.
+
+<strong>Deploy the Lending Pool Contract</strong>
+
+If you haven’t already deployed the contract, navigate to the lending protocol directory and deploy it using Hardhat:
+
+```bash
+cd ../ # Move to the lending protocol directory
+npx hardhat compile
+npx hardhat run-lending-protocol
```
idea404, line 69: this requires: 

- installed `node_modules` for this lending protocol dir package.json
- a globally installed `hardhat`
- defined `.env` in this lending protocol dir

idea404 (approved):
LGTM. A couple ideas: 

- show on dashboard:
  - total tokens available for lending
  - total tokens lent
  - total paid back interest
  - total loans processed
- Maybe also good to extract the nil-specific portions of code into components, that helps find where in code you are using nil, could put them in a dir `components/nil/CompA.tsx`

academy/lending-protocol/frontend/package-lock.json
ukorvl, line 1: Don't add lockfile here. use this as a workspace of monorepo.

ukorvl (changes requested):

academy/lending-protocol/frontend/package-lock.json
gitshreevatsa: This project works as a template, such that anyone can just `cp -r /this-repo /desired-location` and build their application logic with the wallet integration already existing. If lockfile is removed then it would no longer serve as a template for quick development I believe.

ukorvl: Why it can not serve as a template without `package-lock`? Moreover removal of `package-lock` will only improve the template because you don't know which package manager our potential developer uses.

gitshreevatsa: Agreed, previously the lending protocol hardhat project had mismatched dependencies for the local package.json and the root package.json especially with niljs versions and the lockfile actually evaded this issue and hence retained in this project. But I agree to your point that any package manager could be used.

PR #562, opened 2025-03-14
[GH-561] Connection manager

### LINKED ISSUE #561: Implement connection manager with scoring mechanism
The goal is to keep "bad" peers from poisoning our lives. Of course, we [check][1] the signature of a block that comes to us and discard it if it is incorrect, but we are not protecting ourselves from spam. It is suggested to refer to [peerset][2] from `gossamer` as a reference.

In `gossamer` the corresponding mechanism is a rather complex entity, involving a full implementation from scratch of the `connmgr.ConnManager` interface, various connection pools, message queues, etc. etc. etc. 

Nevertheless, they started with quite [primitive mechanisms][3], which can serve as a good argument for us also not to overcomplicate ahead of time. Essentially, we just need to inject ourselves into the `network.Notifiee.Connected` processing and, depending on the reputation of the peer, decide whether to drop the connection with it or not.

The current state of `gossamer` was significantly approached in the task https://github.com/ChainSafe/gossamer/issues/1399. According to it, a [document][4] was written describing how such a mechanism is implemented in `substrate`. The implementation was added as part of https://github.com/ChainSafe/gossamer/pull/1791.

Separately, I was interested in the issue of choosing the values of numerous constants in `substrate`/`gossamer`. Investigation showed that they [started][5] with [arbitrarily chosen values][6] and then iteratively [tweaked and added new ones][7]. So I think we should do the same and add/tune constants as the subsystem evolves.

[1]: https://github.com/NilFoundation/nil/blob/072cfe9136645b9014fd5a17db4bbf37ed690279/nil/internal/collate/validator.go#L340
[2]: https://github.com/ChainSafe/gossamer/tree/ff33dc50f902b71bb7940a66269ac2bf194a59c7/dot/peerset
[3]: https://github.com/ChainSafe/gossamer/pull/1034/files
[4]: https://hackmd.io/@FyilpQ7sQ265a9xyqSxyLg/B1EIbIIfY
[5]: https://github.com/paritytech/substrate/pull/2110/commits/43ca14e752eb1d7606b963fbb8a907a4bf09dad7#diff-796f8563b2fdf3d032751227d48d9ee8a6e1190078e018d8e3d79293b670c0f0R314
[6]: https://github.com/paritytech/substrate/pull/2110/commits/904a521297f42cb9653e514588596deccf4164e3#r269980744
[7]: https://github.com/paritytech/substrate/pull/2229/files#diff-82213a47899307431bdf8ab0d4dddae30eadb5e8d1048083a26e9d18ccb90098

Un1oR:
Resolves #561

### DIFF
--- a/nil/internal/network/connection_manager/config.go
+++ b/nil/internal/network/connection_manager/config.go
@@ -1,0 +1,53 @@
+package connection_manager
+
+import (
+	"time"
+
+	"github.com/jonboulle/clockwork"
+)
+
+// NOTE: This is a variation of an enum. The type is declared as private,
+// while the constants of this type are public. Therefore, module users
+// can only pass a strictly defined set of values of the given type,
+// which is what we aim to achieve.
+type reputationChangeReason string
+
+const (
+	ReputationChangeInvalidBlockSignature = reputationChangeReason("invalid block signature")
+)
+
+type ReputationChangeSettings = map[reputationChangeReason]Reputation
+
+func DefaultReputationChangeSettings() ReputationChangeSettings {
+	return ReputationChangeSettings{
+		ReputationChangeInvalidBlockSignature: -100,
+	}
+}
+
+type Config struct {
+	// DecayReputationPerSecondPercent is the percentage of reputation that is lost per second.
+	DecayReputationPerSecondPercent uint `yaml:"decayReputationPerSecondPercent,omitempty"`
+	// RecalculateReputationsTimeout is the amount of time between recalculating the reputations of all peers.
+	RecalculateReputationsTimeout time.Duration `yaml:"recalculateReputationsTimeout,omitempty"`
+	// ForgetAfterTime amount of time between the moment we disconnect
+	// from a peer and the moment we remove it from the list.
+	ForgetAfterTime time.Duration `yaml:"forgetAfterTime,omitempty"`
+	// ReputationBanThreshold is the reputation threshold below which a peer is banned.
+	ReputationBanThreshold Reputation `yaml:"reputationBanThreshold,omitempty"`
+	// ReputationChangeSettings is a map of reasons for reputation changes and the
+	// reputation change that should be applied.
+	ReputationChangeSettings ReputationChangeSettings `yaml:"reputationChangeSettings,omitempty"`
+
+	clock clockwork.Clock `yaml:"-"`
+}
+
+func NewDefaultConfig() *Config {
+	return &Config{
+		DecayReputationPerSecondPercent: 2, // A bit low, then 35 seconds to reduce reputation by half (0.98^35 = 0.49)
+		RecalculateReputationsTimeout:   1 * time.Second,
+		ForgetAfterTime:                 2 * time.Hour,
+		ReputationBanThreshold:          -200,
+		ReputationChangeSettings:        DefaultReputationChangeSettings(),
+		clock:                           clockwork.NewRealClock(),
+	}
+}
--- a/nil/internal/network/connection_manager/connection_manager.go
+++ b/nil/internal/network/connection_manager/connection_manager.go
@@ -1,0 +1,41 @@
+package connection_manager
+
+import (
+	"context"
+
+	libp2pconnmgr "github.com/libp2p/go-libp2p/core/connmgr"
+	"github.com/libp2p/go-libp2p/core/network"
+	"github.com/libp2p/go-libp2p/p2p/net/connmgr"
+	"github.com/rs/zerolog"
+)
+
+type withCustomNotifeeDecorator struct {
+	libp2pconnmgr.ConnManager
+
+	notifee network.Notifiee
+}
+
+func (cm *withCustomNotifeeDecorator) Notifee() network.Notifiee {
+	return cm.notifee
+}
+
+var _ libp2pconnmgr.ConnManager = (*withCustomNotifeeDecorator)(nil)
+
+func NewConnectionManagerWithPeerReputationTracking(
+	ctx context.Context,
+	conf *Config,
+	logger zerolog.Logger,
+	low, hi int,
+	opts ...connmgr.Option,
+) (libp2pconnmgr.ConnManager, error) {
+	baseConnectionManager, err := connmgr.NewConnManager(low, hi, opts...)
+	if err != nil {
+		return nil, err
+	}
+	notifee := newNotifiee(baseConnectionManager.Notifee(), conf, logger)
+	notifee.start(ctx)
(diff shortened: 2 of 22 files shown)


Un1oR:
@dmtrskv @olegrok I will finalize the TODO edits in the code and the issue and PR descriptions a bit later, but I think it's already possible to evaluate the general points so that we don't waste time.

nil/internal/network/host.go
```diff
@@ -15,46 +17,52 @@ import (
...
 // newHost creates a new libp2p host. It must be closed after use.
-func newHost(ctx context.Context, conf *Config) (Host, error) {
+func newHost(ctx context.Context, conf *Config) (Host, zerolog.Logger, error) {
 	addr := conf.IPV4Address
 	if addr == "" {
 		addr = "0.0.0.0"
 	}
 
-	options, err := getCommonOptions(ctx, conf.PrivateKey)
+	options, logger, err := getCommonOptions(ctx, conf)
```
olegrok, line 63: Maybe we need to call `getCommonOptions` outside `newHost`. I don't like that newHost returns logger now.

nil/internal/network/host.go
```diff
@@ -102,14 +114,18 @@ func newClient(ctx context.Context, conf *Config) (Host, error) {
...
-			return nil, err
+			return nil, zerolog.Nop(), err
 		}
 	}
 
-	options, err := getCommonOptions(ctx, privateKey)
+	options, logger, err := getCommonOptions(ctx, &Config{PrivateKey: privateKey})
 	if err != nil {
-		return nil, err
+		return nil, zerolog.Logger{}, err
```
olegrok, line 123: It's a bit inconsistent with return above: `zerolog.Logger{}` vs `zerolog.Nop()`

nil/internal/network/connection_manager/notifiee.go
```diff
@@ -0,0 +1,180 @@
...
+			n.logger.Error().Err(err).Msgf("Failed to close peer %s", peer)
+		}
+	}
+	var pi *peerInfo
+	var ok bool
+	if pi, ok = n.peerReputations[peer]; !ok {
+		pi = newPeerInfo(peer, 0, closeFunc)
+		n.peerReputations[peer] = pi
+	} else if pi.closeFunc == nil {
+		pi.closeFunc = closeFunc
```
olegrok, line 78: I guess you can create `closeFunc` under this if

nil/internal/network/connection_manager/peer_info.go
```diff
@@ -0,0 +1,40 @@
...
+// and will be able to use it if necessary.
+type peerInfo struct {
+	id         peer.ID
+	reputation Reputation
+	closeFunc  func()
+}
+
+func (pi *peerInfo) closePeer(logger zerolog.Logger) {
+	if pi.closeFunc != nil {
+		logger.Debug().Stringer("peerId", pi.id).Msg("Disconnecting banned peer")
```
olegrok, line 27: You can log this right inside `pi.closeFunc` and don't pass logger inside `closePeer`

nil/internal/network/connection_manager/testaide.ge.go
```diff
@@ -0,0 +1,31 @@
+//go:build test
```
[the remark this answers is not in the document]
Un1oR: A typo)

nil/internal/network/host.go
Un1oR: I don't have a good opinion on that. I can do that if you think it's best.

Un1oR: Frankly, this would affect `newClient`/`NewClientManager`, and thus the private key work that happens there. I'm not sure how good and timely this is. If not critical, we can make some changes here as a separate PR.

nil/internal/network/connection_manager/config.go
```diff
@@ -0,0 +1,49 @@
...
+	// ReputationChangeSettings is a map of reasons for reputation changes and the
+	// reputation change that should be applied.
+	ReputationChangeSettings ReputationChangeSettings `yaml:"reputationChangeSettings,omitempty"`
+
+	clock clockwork.Clock `yaml:"-"`
+}
+
+func NewDefaultConfig() *Config {
+	return &Config{
+		DecayReputationPerSecondPercent: 2, // A bit low, then 35 seconds to reduce reputation by half
```
olegrok, line 42: I'd add a formula to calculate it. Connection between 32 and 2 is not so obvious.

nil/internal/network/connection_manager/connection_manager.go
```diff
@@ -0,0 +1,41 @@
...
+	return cm.notifee
+}
+
+var _ libp2pconnmgr.ConnManager = (*withCustomNotifeeDecorator)(nil)
+
+func NewConnectionManagerWithPeerReputationTracking(
+	ctx context.Context,
+	conf *Config,
+	logger zerolog.Logger,
+	low, hi int,
```
olegrok, line 28: I suggest to change the order of arguments here. Usually we pass logger as the last argument.
Here it's impossible to make it the latest but let's move it a bit closer to the end.

nil/internal/network/host.go
```diff
@@ -15,46 +17,52 @@ import (
...
+		return nil, zerolog.Nop(), err
 	}
 
-	pid, err := peer.IDFromPublicKey(privateKey.GetPublic())
+	logger := internal.Logger.With().
+		Stringer(logging.FieldP2PIdentity, pid).
+		Logger()
+
+	cm, err := connection_manager.NewConnectionManagerWithPeerReputationTracking(
+		ctx, conf.ConnectionManagerConfig, logger, 100, 400, defaultGracePeriod)
```
olegrok, line 38: What do low and hight mean here?
Why did you choose 100 and 400 here?

Un1oR: This is the transposed code, I didn't change the values

nil/internal/network/connection_manager/connection_manager.go
Un1oR: I don't really want to break up the argument list that is passed to `connmgr.NewConnManager` as is.

nil/internal/network/host.go
olegrok: It's worth to understand that they mean

Un1oR: > // lo and hi are watermarks governing the number of connections that'll be maintained.
// When the peer count exceeds the 'high watermark', as many peers will be pruned (and
// their connections terminated) until 'low watermark' peers remain.

olegrok (approved):

PR #571, opened 2025-03-17
[niljs] errors improvements

ukorvl:
This diff brings several niljs errors improvements.
- Adds proper `name` parameter
- Adds `docsBaseUrl` to ease usage of docs urls in errors
- Adds docs urls to "block not found" errors
- Throws an error if block was not found

Also it fixes integration test local running by fixing reading of env variables which are used to connect to cluster and services.


KlonD90 (approved):
lgmt gread docs

PR #573, opened 2025-03-17
[network] check that FindPeers channel is not closed

olegrok:
Otherwise we can get into infinite loop. Right now we observe high CPU usage of this function and let's add this check to the function maybe it will help.

### DIFF
--- a/nil/internal/network/dht.go
+++ b/nil/internal/network/dht.go
@@ -139,16 +139,21 @@ func findPeers(ctx context.Context, rd *routing.RoutingDiscovery, h host.Host, p
 	defer cancel()
 	peerCh, err := rd.FindPeers(ctx, string(protocol))
 	if err != nil {
-		logger.Warn().Err(err).Msgf("failed to begin finding peers via DHT")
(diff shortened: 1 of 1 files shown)


nil/internal/network/dht.go
```diff
@@ -139,16 +139,21 @@ func findPeers(ctx context.Context, rd *routing.RoutingDiscovery, h host.Host, p
...
 	for {
 		select {
 		case <-ctx.Done():
-			logger.Debug().Msgf("findPeers: timer expired")
+			logger.Debug().Msg("findPeers: timer expired")
 			return
-		case peer := <-peerCh:
+		case peer, ok := <-peerCh:
+			if !ok {
+				logger.Trace().Msg("findPeers: peer channel closed")
```
shermike, line 153: Maybe change log level to `Warning` as closed channel is unusual behavior afaiu?

olegrok: Usually producer closes its channel if there is no more information for consumer. So I think it's normal behaviour that channel is closed.

PR #576, opened 2025-03-17
[relay] Generate config

dmtrskv:
Generating config for relay from relay cli.
Additionally, some improvements of cli commands

### DIFF
--- a/nil/cmd/relay/main.go
+++ b/nil/cmd/relay/main.go
@@ -2,10 +2,14 @@ package main
 
 import (
 	"context"
+	"errors"
+	"fmt"
+	"os"
 	"os/signal"
 	"syscall"
+	"time"
 
-	"github.com/NilFoundation/nil/nil/common/check"
+	"github.com/NilFoundation/nil/nil/common/concurrent"
 	"github.com/NilFoundation/nil/nil/common/logging"
 	"github.com/NilFoundation/nil/nil/internal/cobrax"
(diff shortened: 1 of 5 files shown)


nil/internal/network/config.go
```diff
@@ -24,7 +24,7 @@ type Config struct {
 
 	DHTEnabled        bool          `yaml:"dhtEnabled,omitempty"`
 	DHTBootstrapPeers AddrInfoSlice `yaml:"dhtBootstrapPeers,omitempty"`
-	DHTMode           dht.ModeOpt   `yaml:"-,omitempty"`
+	DHTMode           dht.ModeOpt   `yaml:"-"`
```
dmtrskv, line 27: adding omitempty appears to be incorrect! in this case yaml considers `-` a field name.

nil/cmd/relay/main.go
```diff
@@ -39,34 +60,93 @@ func parseArgs() *config {
...
 	nm, err := network.NewManager(ctx, cfg.Network, nil)
-	check.PanicIfErr(err)
+	if err != nil {
+		return err
+	}
 	defer nm.Close()
 
-	<-ctx.Done()
+	concurrent.RunTickerLoop(ctx, reportPeriod, func(ctx context.Context) {
+		logger.Info().Msg("I am still alive")
```
dmtrskv, line 139: added some logs, so that the relay doesn't stay completely silent

Un1oR (approved):

PR #578, opened 2025-03-17
[logging] Introduce typed logger with clickhouse support

isergeyam:
This patch rewrites the logging library, introducing a wrapper over `zerolog.Logger`, that retains the typing information. This information is used to introduce columns in clickhouse.


nil/common/logging/logger.go
```diff
@@ -33,18 +34,38 @@ func (w ComponentFilterWriter) Write(p []byte) (n int, err error) {
...
+			switch k {
+			case zerolog.CallerFieldName,
+				zerolog.TimestampFieldName,
+				zerolog.MessageFieldName,
+				zerolog.LevelFieldName,
+				FieldComponent:
+				logWithoutType[k] = v
+				continue
+			}
+			logWithoutType[k[:len(k)-2]] = v
```
olegrok, line 60: Why `-2`?

isergeyam: That's the length of the type [abbreviation](https://github.com/NilFoundation/nil/blob/b7668859d6b1a8af4383a254211711bbf40852de/nil/common/logging/typed_logger.go#L15)

olegrok: please move to a constant and add assert in init that all abbreviations have 2 symbols length.

isergeyam:
Tested on nil-dev/data-analytics-dev, the thing generally works

nil/common/logging/logger.go
isergeyam: done

nil/common/logging/fields.go
```diff
@@ -55,4 +55,10 @@ const (
 	FieldClientType    = "clientType"
 	FieldClientVersion = "clientVersion"
 	FieldUid           = "uid"
+
+	FieldStoreToClickhouse      = "storeToClickhouse"
+	FieldStoreToClickhouseTyped = "storeToClickhouseB1"
```
Zerg1996, line 60: I think we also need to call abbr function here. Because it's not obvious what is B1 mean.

Zerg1996 (approved):
lgtm, but small note...

nil/common/logging/fields.go
isergeyam: Done

PR #579, opened 2025-03-17
[exporter] Support `AsyncRequestInfo` in clickhouse driver

shermike:


nil/cmd/exporter/internal/clickhouse/clickhouse_test.go
```diff
@@ -0,0 +1,76 @@
...
+	clickhouse *exec.Cmd
+	driver     *ClickhouseDriver
+}
+
+func (s *SuiteClickhouse) SetupSuite() {
+	dir := s.T().TempDir()
+	clickhouse := exec.Command( //nolint:gosec
+		"clickhouse", "server", "--",
+		"--listen_host=127.0.0.1",
+		"--tcp_port=9002",
```
olegrok, line 23: cometa test already uses this port. It can fail in CI during parallel test

olegrok (approved):

PR #582, opened 2025-03-18
improved clickhouse columns

### LINKED ISSUE #469: put expolorer's ClickHouse columns in order
two minor improvements to our ClickHouse storage which is used by block explorer:

1. column with num of out transactions is duplicated twice: [here](https://github.com/NilFoundation/nil/blob/31e0ffba9e9c9bf687b6b73aefd97c444f733464/nil/cmd/exporter/internal/clickhouse/clickhouse.go#L33) and [here](https://github.com/NilFoundation/nil/blob/31e0ffba9e9c9bf687b6b73aefd97c444f733464/nil/internal/types/block.go#L42)
2. most of columns are in snake case, but [some](https://github.com/NilFoundation/nil/blob/fc0c868ca12c14c21894c944b4c1f438a6726730/nil/internal/types/transaction.go#L164) are in camel case. let's consistently keep only the first one

varun-doshi:
Closes #469 

Changes:
- Changed clickhouse column name to `out_txn_num` in `BlockData.OutTransactionsNum` in file `internal/types/block.go` to match with `BlockWithBinary.OutTxnNum` in `clickhouse.go`
- converted from camelcase ch names to snake case in `internal/types/transaction.go`

### DIFF
--- a/nil/internal/types/block.go
+++ b/nil/internal/types/block.go
@@ -41,7 +41,7 @@ type BlockData struct {
 	// It can be considered as an array, where each segment is referred by corresponding receipt.
 	OutTransactionsRoot common.Hash `json:"outTransactionsRoot" ch:"out_transactions_root"`
 	// We cache the size of out transactions, otherwise we should iterate all the tree to get its size
-	OutTransactionsNum  TransactionIndex `json:"outTransactionsNum" ch:"out_transaction_num"`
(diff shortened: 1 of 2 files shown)


knazarov:
I've absorbed your pull request in #589. We can't directly merge your PR because commits have to be signed by maintainers. But the absorbed PR will still show you as author, so that should be OK.

olegrok:
We pushed your commit as a part of https://github.com/NilFoundation/nil/pull/589

Thanks for your contribution!

PR #583, opened 2025-03-18
[rpc] add metrics to rpc request handler

olegrok:
This patch adds duration/count metrics to JSON RPC request handler.


shermike:
Can we also add a counter for failed rpc requests? I guess it would be useful to know how many requests were failed.

olegrok:
> Can we also add a counter for failed rpc requests? I guess it would be useful to know how many requests were failed.

Yes. Added a counter for failed requests.

shermike (approved):
Thanks!

Issue #584, opened 2025-03-18
Explorer fails to deploy a valid contract


khannanov-nil:
# Problem statement

Explorer fails to deploy a valid contract. The contract address is empty but Explorer displays a successful deployment message in the logs. Cometa registration fails, however.

# Steps to reproduce

1. Use these contracts:

```solidity
// SPDX-License-Identifier: MIT

pragma solidity ^0.8.21;

import "@nilfoundation/smart-contracts/contracts/Nil.sol";
import "@nilfoundation/smart-contracts/contracts/NilTokenBase.sol";

/**
 * @title Receiver
 * @author =nil; Foundation
 * @notice A simple contract for storing the NFT.
 */
contract Receiver {
    function deposit() public payable {}
}

/**
 * @title NFT
 * @author =nil; Foundation
 * @notice A contract representing a non-fungible token.
 */
contract NFT is NilTokenBase {
    bool private hasBeenSent = false;
    address private owner;

    constructor() public payable {
        owner = msg.sender;
    }

    function deposit() public payable {}

    modifier onlyOwner() {
        require(msg.sender == owner, "Only owner can call this function");
        _;
    }

    /**
     * The function for minting the NFT.
     * It must call mintTokenInternal().
     */
    function mintNFT() public onlyOwner {
        require(totalSupply == 0, "NFT has already been minted");
        require(!hasBeenSent, "NFT has already been sent");
        mintTokenInternal(1);
    }

    /**
     * The function for sending the NFT.
     * It must call sendTokenInternal().
     * @param to The address where the NFT should be sent.
     */
    function sendNFT(address to) public onlyOwner {
      require(!hasBeenSent, "NFT has already been sent");
        Nil.Token[] memory nft = new Nil.Token[](1);
        nft[0].id = getTokenId();
        nft[0].amount = 1;
        Nil.asyncCallWithTokens(
            to,
            msg.sender,
            msg.sender,
            0,
            Nil.FORWARD_REMAINING,
            0,
            nft,
            ""
        );
        hasBeenSent = true;
    }
}
```

2. Compile and deploy both contracts (Receiver on Shard 1, NFT on Shard 2)

# Actual behavior

Receiver is deployed. NFT fails to deploy and the logs show this:

![Image](https://github.com/user-attachments/assets/6fb25f8e-373f-4ca5-9088-e928ff3e6f0c)

# Expected behavior

Both contracts are deployed normally.

khannanov-nil:
[...]

PR #585, opened 2025-03-18
[explorer] added two more interactive tutorials

khannanov-nil:
This diff adds two more tutorials and their checks to interactive tutorials. It also adjusts the hardcoded `feeCredit` used for deploying smart contracts. The previous `feeCredit` was too low to deploy some contracts, most notably the `NFT` contract from Tutorial 5. In the future, a better solution for handling `feeCredit` here is needed.


explorer_frontend/src/features/contracts/models/base.ts
```diff
@@ -112,7 +112,7 @@ export const deploySmartContractFx = createEffect<
     args,
     salt,
     shardId,
-    feeCredit: convertEthToWei(0.00001),
+    feeCredit: convertEthToWei(0.0002),
```
KlonD90, line 115: Why use feeCredit here?

khannanov-nil: Because of https://github.com/NilFoundation/nil/issues/584

PR #588, opened 2025-03-18
[GH-#] Limit the length of lines to 120 characters

Un1oR:
In some places our code looks pretty awful, going over the horizon with its lines. In the W4-5'25 retrospective @shermike suggested adding a 120-character line length linter, which was done in this PR.
The whole code was reviewed and partially corrected by hand subjectively in the most sympathetic way in each case.


nil/internal/params/protocol_params.go
```diff
@@ -64,21 +71,26 @@ const (
...
 	CreateGas             uint64 = 32000 // Once per CREATE operation & contract-creation transaction.
 	Create2Gas            uint64 = 32000 // Once per CREATE2 operation
 	SelfdestructRefundGas uint64 = 24000 // Refunded following a selfdestruct operation.
-	MemoryGas             uint64 = 3     // Times the address of the (highest referenced byte in memory + 1). NOTE: referencing happens on read, write and in instructions such as RETURN and CALL.
-
-	TxDataNonZeroGasFrontier  uint64 = 68   // Per byte of data attached to a transaction that is not equal to zero. NOTE: Not payable on data of calls between transactions.
-	TxDataNonZeroGasEIP2028   uint64 = 16   // Per byte of non-zero data attached to a transaction after EIP 2028 (part in Istanbul)
+	MemoryGas             uint64 = 3     // Times the address of the (highest referenced byte in memory + 1).
+	//                                      NOTE: referencing happens on read,
+	//                                      write and in instructions such as RETURN and CALL.
```
olegrok, line 88: Let's put the comments above variables

Un1oR: image
What confuses me is that the IDE then tells me that the comments should be styled differently. This is not a mechanical change that can be done without thinking (and without knowing the code). Our linters don't check such things, so we can forget about it. What do you think is the best thing to do? Should I still make all comments over variables?

zadykian (approved):

knazarov:
Please always add descriptions to PRs. Submitting a PR without a description sets a bad precedent overall. It should at least contain motivation for the change.

nil/internal/params/protocol_params.go
olegrok: Yes. Because right now this part is formatted in a quite ugly way.

Un1oR: I agree, I thought long and hard about what to do here too. I did it the way you suggested.

PR #595, opened 2025-03-18
relayer: finality ensurer impl

oclaw:
Initial implementation of Relayer `FinalityEnsurer` component impl

It is responsible for:
- keeping actual finalized block number from L1
- polling the L1 event storage for event from blocks whose number less or equal than the latest finalized block
- filtering orphaned events (events from blocks that were not included into the final chain)
- forwarding events to the L2 storage for further relaying to L2


nil/services/relayer/internal/l1/finality_ensurer.go
```diff
@@ -1,29 +1,377 @@
...
+	// limited size storage to fetch events with min sequence number
+	eventBySeqNo := heap.NewBoundedMaxHeap(1000, func(a, b *Event) int {
+		return cmp.Compare(a.SequenceNumber, b.SequenceNumber)
+	})
+
+	checkedEvents := 0
+	if err := fe.l1Storage.IterateEventsByBatch(ctx, 100, func(eventBatch []*Event) error {
+		checkedEvents += len(eventBatch)
+		for _, evt := range eventBatch {
+			if evt.BlockNumber <= finalizedBlock.BlockNumber {
```
akokoshn, line 215: if heap ordered, can we break once `evt.BlockNumber > finalizedBlock.BlockNumber`

nil/services/relayer/internal/l1/finality_ensurer.go
```diff
@@ -1,29 +1,377 @@
...
+				Uint64("orphaned_block_num", blk.BlockNumber).
+				Msg("found orphaned block")
+
+			orphaned = append(orphaned, blk)
+		} else {
+			finalized = append(finalized, blk)
+		}
+	}
+
+	return finalized, orphaned, nil
```
akokoshn, line 327: events from `orphaned` blocks will be removed from `l1Storage` and not stored to `l2Storage` - relayer just skip them, right?

nil/services/relayer/internal/l2/storage.go
```diff
@@ -0,0 +1,133 @@
...
+			if err := tx.Delete(pendingEventsTable, hash.Bytes()); err != nil && !errors.Is(err, db.ErrKeyNotFound) {
+				return err
+			}
+		}
+
+		return es.Commit(tx)
+	})
+}
+
+func (*EventStorage) Commit(tx db.RwTx) error {
```
akokoshn, line 128: seems already defined in `BaseStorage`

nil/services/relayer/internal/l1/finality_ensurer.go
oclaw: Yes, this case is covered in tests

nil/services/relayer/internal/l1/finality_ensurer.go
```diff
@@ -1,29 +1,377 @@
...
+	fe.logger.Info().Msg("started l1 pending event processor")
+
+	ticker := fe.clock.NewTicker(fe.config.DbPollInterval)
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-ticker.Chan():
+			fe.logger.Debug().Msg("wake up by timer")
+		case <-fe.eventProvider.EventReceived():
```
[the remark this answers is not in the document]
oclaw: Not sure about it. We have external event provider (listener) which writes events to the l1 storage and I would like to use it as a notifier anyway

Let's imagine the case: relayer was down for some time and after geth and relayer are back and running we have fetched some historical events. And it is better for us to start processing them immediately as soon as they have arrived

We can also add new condition to the poller - `blockFetched` you've mentioned. But replacing event provider with it seems doubtful for me

nil/services/relayer/internal/l1/finality_ensurer.go
oclaw: Heap is ordered, but the storage is not (actually is is, but only by the key, which is message hash here)
We can avoid full scan of the collection with introducing additional index collection which will be ordered by sequence number of the event, but it will complicate the logic and will require more testing. As soon as we agreed that for the beginning querying the whole collection is ok for us - this way is the best imho

nil/services/relayer/internal/l1/finality_ensurer.go
```diff
@@ -1,29 +1,377 @@
...
+			if evt.BlockNumber <= finalizedBlock.BlockNumber {
+				eventBySeqNo.Add(evt)
+			}
+		}
+		return nil
+	}); err != nil {
+		return err
+	}
+
+	events := eventBySeqNo.PopAllSorted()
```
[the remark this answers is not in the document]
oclaw: They are pointers, copying is not involved

nil/services/relayer/internal/l1/finality_ensurer.go
```diff
@@ -1,29 +1,377 @@
...
+
+func (fe *FinalityEnsurer) forwardFinalizedEvents(ctx context.Context) error {
+	finalizedBlock, hasFinalizedBlock := fe.getLatestFinalizedBlock()
+	if !hasFinalizedBlock {
+		fe.logger.Debug().Msg("no finalized block number received from L1")
+		return nil
+	}
+
+	// limited size storage to fetch events with min sequence number
+	eventBySeqNo := heap.NewBoundedMaxHeap(1000, func(a, b *Event) int {
```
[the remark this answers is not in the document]
oclaw: done

nil/services/relayer/internal/l1/finality_ensurer.go
```diff
@@ -1,29 +1,377 @@
...
+					log = log.Int("time_since_last_update_sec", int(diff.Seconds()))
+				}
+				if fe.finalizedBlock != nil {
+					log = log.Uint64("local_finalized_block_number", fe.finalizedBlock.BlockNumber)
+				}
+				log.Msg("failed to fetch last finalized block number from Etherium")
+				continue
+			}
+
+			fe.finalizedBlockLock.Lock()
```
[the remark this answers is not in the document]
oclaw: done

nil/services/relayer/internal/l2/storage.go
oclaw: yep, removed

nil/services/relayer/internal/l1/finality_ensurer.go
akokoshn: ok, let's keep as is

akokoshn (approved):

PR #596, opened 2025-03-18
implement basic geth test

AndreyMlashkin:


nix/rollup-bridge-contracts.nix
```diff
@@ -45,6 +48,39 @@ stdenv.mkDerivation rec {
     npx hardhat clean && npx hardhat compile
   '';
 
+  doCheck = enableTesting;
+  checkPhase = ''
+    source .env
+    echo "Starting go-ethereum in background..."
```
akokoshn, line 54: Don't run separate `geth` process in CI.
Only unit tests

akokoshn (changes requested):

PR #600, opened 2025-03-19
Sync Committee: Initial Batch Construction

zadykian:
### Sync Committee: Initial Batch Construction

To align with state root initialization on the `Proposer` side, the latest proved state root (or genesis block) is used as a starting point in the Aggregator as well.

* Implemented the `getStartingBlockRef()` method, which retrieves the starting point for the next fetching iteration, prioritizing the latest fetched block if available;
* If neither `LatestFetched` nor `LatestProvedStateRoot` is defined, the method returns `ErrStateRootNotInitialized`;
* Added a check to verify whether the `LatestFetched` block exists on L2, triggering a reset if it does not (`requireBlockExistsInChain`);

In short, `LatestProvedStateRoot` must now be initialized to be used by `Aggregator` as the starting point for block processing

### DIFF
--- a/nil/services/synccommittee/core/aggregator.go
+++ b/nil/services/synccommittee/core/aggregator.go
@@ -143,24 +143,24 @@ func (agg *aggregator) Resume(ctx context.Context) error {
 }
 
 func (agg *aggregator) runIteration(ctx context.Context) {
-	err := agg.processNewBlocks(ctx)
+	err := agg.processBlocksAndHandleErr(ctx)
 	if err != nil {
-		agg.logger.Error().Err(err).Msg("error during processing new blocks")
 		agg.metrics.RecordError(ctx, agg.Name())
 	}
 }
 
-// processNewBlocks fetches and processes new blocks for all shards.
+// processBlocksAndHandleErr fetches and processes new blocks for all shards.
 // It handles the overall flow of block synchronization and proof creation.
-func (agg *aggregator) processNewBlocks(ctx context.Context) error {
-	latestBlock, err := agg.fetchLatestBlockRef(ctx)
-	if err != nil {
-		return err
-	}
+func (agg *aggregator) processBlocksAndHandleErr(ctx context.Context) error {
+	err := agg.processBlockRange(ctx)
+	return agg.handleProcessingErr(ctx, err)
+}
 
-	err = agg.processShardBlocks(ctx, *latestBlock)
-
+func (agg *aggregator) handleProcessingErr(ctx context.Context, err error) error {
 	switch {
+	case err == nil:
+		return nil
+
 	case errors.Is(err, types.ErrBatchNotReady):
 		agg.logger.Warn().Err(err).Msg("received unready block batch, skipping")
 		return nil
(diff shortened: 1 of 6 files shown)


nil/services/synccommittee/core/aggregator.go
```diff
@@ -144,24 +144,24 @@ func (agg *aggregator) Resume(ctx context.Context) error {
 }
 
 func (agg *aggregator) runIteration(ctx context.Context) {
-	err := agg.processNewBlocks(ctx)
+	err := agg.processAndHandleErr(ctx)
```
akokoshn, line 147: `processBlocksAndHandleErr`

nil/services/synccommittee/core/aggregator.go
```diff
@@ -173,69 +173,84 @@ func (agg *aggregator) processNewBlocks(ctx context.Context) error {
...
 // If `latestFetched` value is not defined, method uses `latestProvedStateRoot`.
-func (agg *aggregator) getLatestHandledBlockRef(ctx context.Context) (*types.MainBlockRef, error) {
+// If neither of the two values is defined, method returns an error.
+func (agg *aggregator) getStartingBlockRef(ctx context.Context) (*types.MainBlockRef, error) {
 	latestFetched, err := agg.blockStorage.TryGetLatestFetched(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("error reading latest fetched block for the main shard: %w", err)
 	}
 	if latestFetched != nil {
+		if err := agg.requireBlockExistsInChain(ctx, *latestFetched); err != nil {
```
akokoshn, line 250: seems this check no needed, if the `lastFetched` not exist we should detect it when fetch from `lastFetched` to `latest`

nil/services/synccommittee/core/aggregator.go
```diff
@@ -246,22 +261,39 @@ func (agg *aggregator) getLatestHandledBlockRef(ctx context.Context) (*types.Mai
...
 	if latestProvedRoot == nil {
-		agg.logger.Debug().Msg("Latest proved state root is not defined, waiting for proposer initialization")
-		return nil, errors.New("latest proved state root is not defined")
+		return nil, storage.ErrStateRootNotInitialized
 	}
 
 	rpcBlock, err := agg.rpcClient.GetBlock(ctx, coreTypes.MainShardId, *latestProvedRoot, false)
 	if err != nil {
-		return nil, fmt.Errorf("error fetching main block by with hash=%s: %w", *latestProvedRoot, err)
+		return nil, fmt.Errorf("error fetching main block with hash=%s: %w", *latestProvedRoot, err)
```
akokoshn, line 269: good place for use `requireBlockExistsInChain`

nil/services/synccommittee/core/aggregator.go
```diff
@@ -246,22 +261,39 @@ func (agg *aggregator) getLatestHandledBlockRef(ctx context.Context) (*types.Mai
...
 	}
 	if rpcBlock == nil {
-		agg.logger.Warn().Msgf("Main block with hash=%s not found", *latestProvedRoot)
-		return nil, nil
+		return nil, fmt.Errorf(
+			"%w: block with hash=%s (latest proved state root) not found in chain",
+			types.ErrBlockMismatch, *latestProvedRoot,
```
akokoshn, line 274: may be introduce `ErrBlockNotExists`

nil/services/synccommittee/core/aggregator.go
zadykian: Not exactly
On each iteration we fetch starting from the block with number `latestFetched + 1`

nil/services/synccommittee/core/aggregator.go
zadykian: `ErrBlockMismatch` is used to trigger a reset
Here, the mismatch is characterized as "Block does not exists on L2, but exists in Sync Committee"

akokoshn (approved):

PR #608, opened 2025-03-19
Introduce Stresser service

shermike:
Stresser service is a service which spams transactions to a target cluster. Type of transactions and their intensity can be customized in configuration.
At the moment, the following workloads are supported:
- External transactions with adjustable gas consumption.
- Request/response transactions (sendRequest)
- `awaitCall` transactions
- `getBlockRange` RPC requests
- blockchain metric collector - auxiliary workload that fetches blocks from the blockchain and emits metrics about it

Stresser can run in three modes:
- `single-instance` it runs one `nild` binary with all validators and rpc node inside
- `devnet` run separate nodes for validators and rpc node (similar to devnet server).
- `external` runs nothing and uses a specified endpoint as a target rpc node

### DIFF
--- a/nil/common/util.go
+++ b/nil/common/util.go
@@ -1,0 +1,56 @@
+package common
+
+import (
+	"context"
+	"time"
+)
+
+// WaitFor repeatedly calls the given function until it returns true or an error.
+func WaitFor(ctx context.Context, timeout, tick time.Duration, f func(ctx context.Context) bool) error {
+	ctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+
+	ticker := time.NewTicker(tick)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-ticker.C:
+			if f(ctx) {
+				return nil
+			}
+		}
+	}
+}
+
+// WaitForValue repeatedly calls the given function until it returns true or an error.
+// In case function return some data not equal to nil, it returns true.
+func WaitForValue[T any](
+	ctx context.Context,
+	timeout,
+	tick time.Duration,
+	f func(ctx context.Context) (*T, error),
+) (*T, error) {
+	ctx, cancel := context.WithTimeout(ctx, timeout)
+	defer cancel()
+
+	ticker := time.NewTicker(tick)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil, ctx.Err()
+		case <-ticker.C:
+			res, err := f(ctx)
+			if err != nil {
+				return res, err
+			}
+			if res != nil {
+				return res, nil
+			}
+		}
+	}
+}
--- a/nil/internal/execution/state.go
+++ b/nil/internal/execution/state.go
@@ -26,13 +26,18 @@ import (
 	"github.com/NilFoundation/nil/nil/internal/vm"
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/holiman/uint256"
+	zerolog "github.com/rs/zerolog"
 )
 
-var logger = logging.NewLogger("execution")
+const (
+	TraceBlocksEnabled                    = false
+	ExternalTransactionVerificationMaxGas = types.Gas(100_000)
 
-const TraceBlocksEnabled = false
-
-const ExternalTransactionVerificationMaxGas = types.Gas(100_000)
+	ModeReadOnly = "read-only"
+	ModeProposal = "proposal"
+	ModeReplay   = "replay"
+	ModeVerify   = "verify"
+)
 
 var blocksTracer *BlocksTracer
 
@@ -129,6 +134,8 @@ type ExecutionState struct {
 
 	// filled in if a rollback was requested by a transaction
 	rollback *RollbackParams
+
+	logger logging.Logger
 }
 
 type ExecutionResult struct {
@@ -218,6 +225,16 @@ func (e *ExecutionResult) GetError() error {
 	return nil
 }
 
+func (e *ExecutionResult) String() string {
+	if e.Error != nil {
+		return fmt.Errorf("error: %w", e.Error).Error()
+	}
+	if e.FatalError != nil {
+		return fmt.Errorf("fatal: %w", e.FatalError).Error()
+	}
+	return "success"
+}
+
 type revision struct {
 	id           int
 	journalIndex int
@@ -260,6 +277,7 @@ type StateParams struct {
 	Block          *types.Block
 	ConfigAccessor config.ConfigAccessor
 	FeeCalculator  FeeCalculator
+	Mode           string
 }
 
 func NewExecutionState(tx any, shardId types.ShardId, params StateParams) (*ExecutionState, error) {
@@ -274,6 +292,11 @@ func NewExecutionState(tx any, shardId types.ShardId, params StateParams) (*Exec
 		return nil, errors.New("invalid tx type")
 	}
 
+	logger := logging.NewLogger("execution")
+	if params.Mode != "" {
+		logger = logger.With().Str("mode", params.Mode).Logger()
+	}
+
 	feeCalculator := params.FeeCalculator
 	if feeCalculator == nil {
 		feeCalculator = &MainFeeCalculator{}
@@ -283,6 +306,12 @@ func NewExecutionState(tx any, shardId types.ShardId, params StateParams) (*Exec
 	var prevBlockHash common.Hash
 	if params.Block != nil {
 		baseFeePerGas = feeCalculator.CalculateBaseFee(params.Block)
+		if baseFeePerGas.Cmp(params.Block.BaseFee) != 0 {
+			logger.Debug().
+				Stringer("Old", params.Block.BaseFee).
+				Stringer("New", baseFeePerGas).
+				Msg("BaseFee changed")
+		}
 		prevBlockHash = params.Block.Hash(shardId)
 	}
 
@@ -309,6 +338,8 @@ func NewExecutionState(tx any, shardId types.ShardId, params StateParams) (*Exec
 		isReadOnly: isReadOnly,
 
 		FeeCalculator: feeCalculator,
+
+		logger: logger,
 	}
 
 	return res, res.initTries()
@@ -364,10 +395,11 @@ func (es *ExecutionState) GetAccount(addr types.Address) (*AccountState, error) 
 		return nil, fmt.Errorf("GetAccount failed: %w", err)
 	}
 
-	acc, err = NewAccountState(es, addr, data)
+	acc, err = NewAccountState(es, addr, data, es.logger)
 	if err != nil {
 		return nil, fmt.Errorf("NewAccountState failed: %w", err)
 	}
+
 	es.Accounts[addr] = acc
 	return acc, nil
 }
@@ -562,9 +594,9 @@ func (es *ExecutionState) EnableVmTracing() {
 			pc uint64, op byte, gas, cost uint64, scope tracing.OpContext, rData []byte, depth int, err error,
 		) {
 			for i, item := range scope.StackData() {
-				logger.Debug().Msgf("     %d: %s", i, item.String())
+				es.logger.Debug().Msgf("     %d: %s", i, item.String())
 			}
-			logger.Debug().Msgf("%04x: %s", pc, vm.OpCode(op).String())
+			es.logger.Debug().Msgf("%04x: %s", pc, vm.OpCode(op).String())
 		},
 	}
 }
@@ -715,7 +747,7 @@ func (es *ExecutionState) createAccount(addr types.Address) (*AccountState, erro
 
 	es.journal.append(createObjectChange{account: &addr})
 
-	accountState, err := NewAccountState(es, addr, nil)
+	accountState, err := NewAccountState(es, addr, nil, es.logger)
 	if err != nil {
 		return nil, err
 	}
@@ -793,7 +825,7 @@ func (es *ExecutionState) updateGasPrice(txn *types.Transaction) error {
 	if !es.isReadOnly && es.GasPrice.Cmp(txn.MaxFeePerGas) > 0 {
 		if es.BaseFee.Cmp(txn.MaxFeePerGas) > 0 {
 			es.GasPrice = types.Value0
-			logger.Error().
+			es.logger.Error().
 				Stringer("MaxFeePerGas", txn.MaxFeePerGas).
 				Stringer("BaseFee", es.BaseFee).
 				Msg("MaxFeePerGas is less than BaseFee")
@@ -911,7 +943,7 @@ func (es *ExecutionState) AddOutTransaction(
 		}
 	}
 
-	logger.Trace().
+	es.logger.Trace().
 		Stringer(logging.FieldTransactionHash, txnHash).
 		Stringer(logging.FieldTransactionFrom, txn.From).
 		Stringer(logging.FieldTransactionTo, txn.To).
@@ -933,7 +965,7 @@ func (es *ExecutionState) sendBounceTransaction(txn *types.Transaction, execResu
 		return false, nil
 	}
 	if txn.BounceTo == types.EmptyAddress {
-		logger.Debug().Msg("Bounce transaction not sent, no bounce address")
+		es.logger.Debug().Msg("Bounce transaction not sent, no bounce address")
 		return false, nil
 	}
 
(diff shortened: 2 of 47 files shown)


nil/services/stresser/instance.go
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	n.workingDir = rootDir
+	if err := os.MkdirAll(n.workingDir, 0o755); err != nil {
+		return fmt.Errorf("failed to create working directory: %w", err)
+	}
+
+	n.cmd = exec.Command(nildPath, "run", "--nshards", strconv.Itoa(numShards), "--http-port", //nolint:gosec
+		strconv.Itoa(rpcPort), "-l", "debug", "--log-filter", "-consensus")
+	n.cmd.Dir = n.workingDir
```
olegrok, line 62: It's simpler just to run nilservice instead of exec.Command

nil/services/stresser/jsonrpc.go
```diff
@@ -0,0 +1,12 @@
...
+type StresserAPI interface {
+	New(name, config string) error
+	Remove(name string) error
+	ListStressers() ([]string, error)
+	GetStatus(name string) (string, error)
+	Stop(name string) error
+	Resume(name string) error
+}
+
+// TODO
```
olegrok, line 12: What should be done to resolve this TODO?

nil/services/stresser/stresser.go
```diff
@@ -0,0 +1,471 @@
...
+	return NewStresser(string(data), filepath.Dir(configFile))
+}
+
+func NewStresser(configYaml string, configPath string) (*Stresser, error) {
+	var cfg Config
+	if err := yaml.Unmarshal([]byte(configYaml), &cfg); err != nil {
+		return nil, fmt.Errorf("can't parse config: %w", err)
+	}
+
+	if cfg.NumShards <= 1 {
```
olegrok, line 101: If you pass it into `InitSingle` that should be at least two shards - main + base

nil/services/stresser/stresser.go
```diff
@@ -0,0 +1,471 @@
...
+	case modeSingleInstance:
+		node := &NildInstance{}
+		if err := node.InitSingle(cfg.NildPath, cfg.WorkingDir, cfg.RpcPort, cfg.NumShards); err != nil {
+			return nil, fmt.Errorf("failed to init node: %w", err)
+		}
+		s.nodes = append(s.nodes, node)
+		rpc_endpoint = fmt.Sprintf("http://127.0.0.1:%d", cfg.RpcPort)
+	case modeDevnet:
+		devnetFile := resolveFile(cfg.DevnetFile, configPath)
+		cmd := exec.Command(cfg.NildPath, "gen-configs", devnetFile, "--basedir", cfg.WorkingDir) //nolint:gosec
```
olegrok, line 135: Since it's a part of our source code you can do that without `exec.Command`.

nil/services/stresser/workload/blockchain_metrics.go
```diff
@@ -0,0 +1,101 @@
...
+	batchSize = 20
+)
+
+var (
+	meter               = telemetry.NewMeter("stresser")
+	blockTxsNum         = telemetry.Int64Gauge(meter, "block_txs_num")
+	blockExternalTxsNum = telemetry.Int64Gauge(meter, "block_external_txs_num")
+	blockInternalTxsNum = telemetry.Int64Gauge(meter, "block_internal_txs_num")
+	blockResponseTxsNum = telemetry.Int64Gauge(meter, "block_response_txs_num")
+	blockRefundTxsNum   = telemetry.Int64Gauge(meter, "block_refund_txs_num")
```
olegrok, line 23: They look like counters

nil/services/stresser/core/helper.go
```diff
@@ -0,0 +1,160 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to send external transaction: %w", err)
+	}
+
+	txn := NewTransaction(tx)
+
+	return txn, nil
+}
+
+func (h *Helper) CallAndWait(contract *Contract, method string, options *TxOptions, args ...any) (*Transaction, error) {
```
olegrok, line 131: Seems to be unused

nil/services/stresser/workload/send_requests.go
```diff
@@ -0,0 +1,41 @@
...
+	options := &core.TxOptions{FeePack: types.NewFeePackFromGas(100_000_000)}
+	for _, contract := range w.params.Contracts {
+		n := getNumForGasConsumer(w.GasRange.RandomValue())
+		tx, err := w.client.Call(contract, "sendRequests", options, w.addresses, big.NewInt(int64(n)))
+		if err != nil {
+			w.logger.Error().Err(err).Msg("failed to call sendRequests")
+		} else {
+			w.AddTx(tx)
+		}
+	}
```
olegrok, line 39: That should be done in separate gorounines

nil/services/stresser/workload/external_tx.go
```diff
@@ -0,0 +1,52 @@
...
+	for sentTxNum < w.MinTxsPerIteration {
+		for _, contract := range w.params.Contracts {
+			n := getNumForGasConsumer(w.GasRange.RandomValue())
+			if tx, err := w.client.Call(contract, "gasConsumer", options, big.NewInt(int64(n))); err != nil {
+				w.logger.Error().Err(err).Msg("failed to call contract")
+			} else {
+				w.AddTx(tx)
+			}
+			sentTxNum++
+		}
```
olegrok, line 49: That should be done in separate gorounines

nil/services/stresser/workload/workload.go
```diff
@@ -0,0 +1,123 @@
...
+	Init(ctx context.Context, client *core.Helper, args *WorkloadParams) error
+	PreRun(ctx context.Context, args *RunParams)
+	Run(ctx context.Context, args *RunParams) ([]*core.Transaction, error)
+	TotalTxsNum() int
+	CheckIsReady() bool
+}
+
+type WorkloadBase struct {
+	Interval       time.Duration `yaml:"interval"`
+	WaitTxs        bool          `yaml:"waitTxs"`
```
olegrok, line 27: I think you can replace it with `WaitTxsTimeout > 0`

nil/services/stresser/stresser.go
```diff
@@ -0,0 +1,471 @@
...
+
+	workloadArgs := &workload.WorkloadParams{Contracts: contracts, NumShards: s.cfg.NumShards}
+
+	for _, w := range s.workload {
+		if err := w.Workload.Init(ctx, s.client, workloadArgs); err != nil {
+			return fmt.Errorf("failed to init workload: %w", err)
+		}
+		if err := w.Run(ctx); err != nil {
+			return fmt.Errorf("failed to run workload runner: %w", err)
+		}
```
olegrok, line 283: Hmmm... Is it possible to run all workloads at the same time?

nil/services/stresser/stresser.go
```diff
@@ -0,0 +1,471 @@
...
+				_, err := s.failedTxsFile.WriteString(tx.Dump(false))
+				check.PanicIfErr(err)
+			}
+			delete(s.pendingTxs, tx.Hash)
+		}
+	}
+	metrics.PendingTxNum.Record(ctx, int64(len(s.pendingTxs)))
+	metrics.TotalTxNum.Record(ctx, int64(totalTxsNum))
+	metrics.SuccessTxNum.Record(ctx, int64(s.successTxsNum))
+	metrics.FailedTxNum.Record(ctx, int64(s.failedTxsNum))
```
olegrok, line 358: That should be also counters

nil/services/stresser/stresser.go
```diff
@@ -0,0 +1,471 @@
...
+		if !ok {
+			return errors.New("can't parse node")
+		}
+		shardIds, ok := n["shards"].([]any)
+		if !ok {
+			return errors.New("can't parse shards")
+		}
+		if len(shardIds) != 2 {
+			return errors.New("shards element must contain only one shard")
+		}
```
olegrok, line 414: A bit incorrect. `len(shardIds) >= 2` and `at least two shards`

nil/services/stresser/stresser.go
```diff
@@ -0,0 +1,471 @@
...
+		if err := node.Run(ctx, wg); err != nil {
+			return fmt.Errorf("failed to run node %s", node.Name)
+		}
+	}
+
+	logger.Info().Msgf("%d nodes started", len(s.nodes))
+
+	go func() {
+		err := s.runWorkload(ctx)
+		if err != nil {
```
olegrok, line 232:
```suggestion
		if err := s.runWorkload(ctx); err != nil {
```

nil/services/stresser/stresser.go
```diff
@@ -0,0 +1,471 @@
...
+			logger.Error().Err(err).Msg("workload returns error")
+		}
+		close(s.exitChan)
```
olegrok, line 235: You could create it right before `go func()` and remove it from Stresser struct

nil/services/stresser/core/utils.go
```diff
@@ -0,0 +1,16 @@
+package core
+
+import (
+	"crypto/rand"
+
+	"github.com/NilFoundation/nil/nil/common"
+)
+
+func RandomHash() common.Hash {
```
olegrok, line 9: It's a copy-paste from SC testaide.

Let's move it into the common and use in all places.

E.g. we have `GenerateRandomAddress` in `types`. So we can add `RandomHash` in `common`.

olegrok (commented):
Thanks for your patch.

Ideally I'd suggest to move all bugfixes (e.g. shard id out of range check) into separate patch just to reduce size of this one.

nil/services/stresser/stresser.go
shermike: I'm not sure I got your point

nil/services/stresser/stresser.go
shermike: Yeah, I thought about it, but In my mind it is not good to import something from `cmd` folder. Also, it will require some refactoring in `devnet.go`. So I decided it would be better to do it later separately from this patch.

nil/services/stresser/workload/send_requests.go
shermike: You mean each iteration of cycle?

nil/services/stresser/stresser.go
shermike: It doesn't run workload, it just starts their goroutines. Will rename it to `StartMainLoop`

nil/services/stresser/stresser.go
shermike: But afaik we need to pass an addend to the `Add` method, not an absolute value?

nil/services/stresser/stresser.go
olegrok: Oh, sorry. Everything is correct, my bad.

nil/internal/execution/state.go
```diff
@@ -1220,9 +1286,13 @@ func (es *ExecutionState) handleExecutionTransaction(
...
 	check.PanicIfNot(transaction.IsExecution())
 	addr := transaction.To
-	logger.Debug().
+	es.logger.Debug().
+		Stringer(logging.FieldTransactionFrom, transaction.From).
 		Stringer(logging.FieldTransactionTo, addr).
 		Stringer(logging.FieldTransactionFlags, transaction.Flags).
+		Stringer(logging.FieldTransactionHash, transaction.Hash()).
+		Stringer("value", transaction.Value).
+		Stringer("feeCredit", transaction.FeeCredit).
```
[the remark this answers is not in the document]
olegrok: I guess for consistency and to avoid different typos.

nil/internal/types/bitflags.go
```diff
@@ -55,3 +68,11 @@ func (b *BitFlags[T]) Clear() {
 func (b *BitFlags[T]) BitsNum() int {
 	return int(unsafe.Sizeof(b.Bits) * 8)
 }
+
+func (b *BitFlags[T]) Uint64() uint64 {
+	return uint64(b.Bits)
+}
+
+func (b *BitFlags[T]) None() bool {
```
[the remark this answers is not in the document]
olegrok: Ok, let's keep as is.

nil/services/stresser/stresser.go
olegrok: Ok. I won't insist here. We can do that later

nil/services/stresser/workload/send_requests.go
olegrok: Yes. We can perform requests to different contracts in parallel

nil/services/stresser/stresser.go
```diff
@@ -0,0 +1,471 @@
...
+
+func (s *Stresser) checkTransactions(ctx context.Context) {
+	totalTxsNum := 0
+	for _, w := range s.workload {
+		totalTxsNum += w.Workload.TotalTxsNum()
+	}
+	logger.Info().Msgf("Transactions: total=%d, pending=%d, success=%d, failed=%d", totalTxsNum,
+		len(s.pendingTxs), s.successTxsNum, s.failedTxsNum)
+	if s.suspendWorkload.Load() {
+		logger.Info().Msg("Suspended")
```
olegrok, line 336: Is `return` missed?

nil/services/stresser/stresser.go
olegrok: Yes. You need to consider only newly added transactions and add their diff.

nil/services/stresser/workload/send_requests.go
shermike: But each iteration is quite cheap, it doesn't wait anything, it just sends transaction. I'm not sure that paralleling it will give us big benefits.

olegrok: Yes, I understand that but it can test parallel requests to the server. E.g. it can discover some race conditions in txpool logic.

nil/tests/economy/economy_test.go
```diff
@@ -80,6 +81,7 @@ func (s *SuiteEconomy) SetupSuite() {
 		ZeroState:            zeroState,
 		CollatorTickPeriodMs: 200,
 		RunMode:              nilservice.CollatorsOnlyRunMode,
+		DisableConsensus:     true,
```
olegrok, line 84: Once I catch a bug in consensus because different instances calculated gas prices differently. Please return consensus back.

nil/common/timing.go
```diff
@@ -0,0 +1,25 @@
+package common
+
+import (
+	"context"
+	"time"
+)
+
+func WaitFor(ctx context.Context, timeout, tick time.Duration, f func(ctx context.Context) bool) error {
```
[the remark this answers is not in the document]
shermike: In `concurrent.WaitFor` I need to specify return type, and it looks very ugly when I do not need in return value - I must make some random type for it and return it from my function. Btw we can use closures instead of returning type.

nil/services/stresser/stresser.go
shermike: I made `SuccessTxNum` and `FailedTxNum` counters. But `TotalTxNum` is not so easy to make a counter since this metric is gathered from all workloads where it is stored as total tx num. `PendingTxNum` is not a counter, it can decrease.

nil/tests/economy/economy_test.go
shermike: Forgot to remove. I disable consensus while debugging.

nil/services/stresser/stresser.go
olegrok: > `PendingTxNum` is not a counter, it can decrease.

Is it possible to make it `UpDownCounter`?

nil/services/stresser/workload/send_requests.go
shermike: Let's do it in next patche? It is not so easy because we need to add synchronization for array of sent transactions.
And actually I think intervals between these requests won't be much different, regardless of whether we send them from different goroutines or just sequentially.

nil/services/stresser/workload/blockchain_metrics.go
shermike: No, it is different for each block.

nil/services/stresser/stresser.go
shermike: Done

nil/services/stresser/instance.go
shermike: I agree, but let's do it in the next iteration, because it needs some work on configuring the service directly.

nil/common/util.go
```diff
@@ -0,0 +1,25 @@
+package common
+
+import (
+	"context"
+	"time"
+)
+
+func WaitFor(ctx context.Context, timeout, tick time.Duration, f func(ctx context.Context) bool) error {
```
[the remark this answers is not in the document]
olegrok: > I'm not sure that it is somehow relating to concurrent. Tbh I'd prefer to move `concurrent.WaitFor` to `common.WaitForValue`.

That's also fine because it keeps this similar things in one place

olegrok (approved):

PR #609, opened 2025-03-19
Add missing Hardhat dependecies

AndreyMlashkin:
- add missing dependecies
- Reduce the bundle size

### DIFF
--- a/nix/rollup-bridge-contracts.nix
+++ b/nix/rollup-bridge-contracts.nix
@@ -47,7 +47,10 @@ stdenv.mkDerivation rec {
 
   installPhase = ''
     mkdir -p $out
+    cp ../package-lock.json $out/
     cp -r * $out/
     cp .env $out/
+    rm -rf $out/node_modules
+    rm -rf $out/cache
   '';
 }
--- a/nix/npmdeps.nix
+++ b/nix/npmdeps.nix
@@ -16,7 +16,8 @@ in
       ../explorer_backend/package.json
       ../explorer_frontend/package.json
       ../uniswap/package.json
+      ../rollup-bridge-contracts/package.json
     ];
   };
-  hash = "sha256-2WEEt2F8UCbviZg0OApUGbuS0UiOQnDrHFddFIt39i8=";
+  hash = "sha256-pRxleAeCV5iiYixf1f4nyi4St5LNGf9rPk/ul4vPCpU=";
 })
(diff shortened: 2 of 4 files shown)


nix/rollup-bridge-contracts.nix
```diff
@@ -1,6 +1,5 @@
 { lib
 , stdenv
-, biome
```
akokoshn, line 3: why remove `biome`

flake.nix
```diff
@@ -142,6 +142,11 @@
                   ./scripts/binary_patch_version.sh
                 } ./usr/bin/cometa ${versionFull}
                 ${pkgs.fpm}/bin/fpm -s dir -t deb --name ${pkg.pname} -v ${version} --deb-use-file-permissions usr
+
```
akokoshn, line 145: may be move script to `rollup-bridge-contracts/scropts`

akokoshn (commented):
And let's remove `l1-contracts` if it no used anymore

flake.nix
AndreyMlashkin: postrm is a script, that is executed, when a deb-package is removed.
It is something common for the whole bundle.
Right now, there is indeed a code only related to the rollup-bridge-contracts...
I would still keep it here

nix/rollup-bridge-contracts.nix
AndreyMlashkin: because we are not using it :)

flake.nix
akokoshn: if we'll not use this script for clean up another components, better move to the separate folder.

defistar (approved):

akokoshn (changes requested):
let's wait for approve from DBMS team

nix/rollup-bridge-contracts.nix
AndreyMlashkin: Can we skip the linting topic for this PR?
The linting job is failing for me now and it feels not really related to the PR

AndreyMlashkin: Can we skip the linting topic for this PR?
The linting job is failing for me now and it feels not really related to the PR

rollup-bridge-contracts/scripts/postrm.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+set -e
+
+case "$1" in
+remove | purge)
```
knazarov, line 5: I'm not sure I understand why this is needed

AndreyMlashkin: Removed

akokoshn (approved):

PR #610, opened 2025-03-19
[GH-#] Expose BasicHost in `network.Manager`

Un1oR:
A small preparatory edit. I didn't find another way to access `BasicHost`, and internally libp2p uses [the same one](https://github.com/libp2p/go-libp2p/blob/a743f9f88ab0f7249d83789f34234d86b6be5b3b/config/config.go#L567-L568).
The `newtork.Host` alias wasn't used outside the module, so I removed it and left the use of external types as an implementation detail.
We need access to BasicHost, for example, to access [identify.IDService](https://github.com/libp2p/go-libp2p/blob/a743f9f88ab0f7249d83789f34234d86b6be5b3b/p2p/host/basic/basic_host.go#L529).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -61,6 +61,7 @@ require (
 	github.com/shopspring/decimal v1.4.0
 	github.com/spf13/viper v1.20.0
 	go.dedis.ch/kyber/v3 v3.1.0
+	go.uber.org/fx v1.23.0
 	golang.org/x/term v0.30.0
 	golang.org/x/text v0.23.0
 )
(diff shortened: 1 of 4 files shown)


Un1oR:
@olegrok yes, indeed, I was in a hurry. I saw that the service already [does everything](https://github.com/libp2p/go-libp2p/blob/a743f9f88ab0f7249d83789f34234d86b6be5b3b/p2p/protocol/identify/id.go#L849) under the hood.

PR #612, opened 2025-03-20
[explorer] added tests for interactive tutorials

khannanov-nil:
This diff adds tests for interactive tutorials checks. Tests are run using Vitest. Each tutorial is provided with its own solution, and checks have been refactored to accept the `CheckProps` arg. These props provide different values and events depending on whether a check is run from Explorer or from Node.js (i.e., in a test). This allows for running checks themselves as tests without any additional modifications to them or boilerplate code. The new tests are added to CI via the change to Nix derivation and adding the `run_tutorials_tests.sh` script.


explorer_frontend/tests/tutorials/tutorials.test.mts
```diff
@@ -0,0 +1,117 @@
...
+}
+
+
+describe("Tutorial One tests", async () => {
+  test("Tutorial One passes with the given solution", async () => {
+    const code = await import(path.resolve(__dirname, "../../src/features/tutorial/assets/tutorialOne/tutorialOneSolution.sol?raw"));
+    const codeRes = code.default;
+
+    const contracts = await createContracts(codeRes);
+    TEST_PROPS.contracts = contracts;
```
KlonD90, line 53: if you want test propos. create new object with extension it would be more safe way. `const testProps = {...testPropos, contracts };`

KlonD90 (approved):
lgtm one moment fix and gogogo

PR #616, opened 2025-03-20
Support solc standard json in Cometa task

### LINKED ISSUE #598: Make Cometa input json file compatible with Solidity json file
Cometa's input json file is slightly different from Solidity json file(`--standard-json` option) and it has fewer options.
We need to make it fully compatible with a Solidity json file.
It makes it easier for users to interact with Cometa.

shermike:
Now it is possible to specify json input, which is passed to `solc` compiler via `--standard-json` argument, directly to Cometa json task.
To do this the `solcStandardJson` should be added to cometa task json:
```json
{
  "contractName": "Test.sol:Foo",
  "compilerVersion": "0.8.28",
  "solcStandardJson": {
    "language": "Solidity",
    "sources": {
      "Test.sol": {
        "content": "// SPDX-License-Identifier: MIT\npragma solidity >= 0.8.17;\ncontract Foo { constructor() payable {}\n function bar() public pure returns (uint) { return 42; } }\n"
      },
      "TestLib.sol": {
        "content": "// SPDX-License-Identifier: MIT\npragma solidity >= 0.8.17;\ncontract Bar { constructor() {}\n function foo() public pure returns (uint) { return 43; } }\n"
      }
    },
    "settings": {
      "optimizer": {
        "enabled": true
      },
      "debug": {
        "debugInfo": ["location", "snippet"]
      },
      "evmVersion": "cancun",
      "metadata": {
        "appendCBOR": true,
        "bytecodeHash": "ipfs"
      },
      "outputSelection": {
        "*": {
          "*": [
            "abi",
            "metadata",
            "evm.bytecode.object",
            "evm.deployedBytecode.object",
            "evm.deployedBytecode.sourceMap",
            "evm.deployedBytecode.generatedSources",
            "evm.deployedBytecode.functionDebugData",
            "evm.methodIdentifiers"
          ]
        }
      }
    }
  }
}
```

Closes #598


nil/services/cometa/compilation.go
```diff
@@ -136,24 +141,43 @@ func Compile(input *CompilerTask) (*ContractData, error) {
...
+		}
+		if contractDescr != nil {
 			break
 		}
 	}
 	if contractDescr == nil {
 		return nil, errors.New("contract not found in compilation output")
 	}
+	if contractDescr.Validate() != nil {
+		return nil, fmt.Errorf("compilation output validation failed: %w", contractDescr.Validate())
```
olegrok, line 179:
```suggestion
	if err := contractDescr.Validate(); err != nil {
		return nil, fmt.Errorf("compilation output validation failed: %w", err)
```

nil/services/cometa/types.go
```diff
@@ -142,6 +142,10 @@ type CompilerTask struct {
 	BasePath        string             `json:"basePath,omitempty"`
 	Sources         map[string]*Source `json:"sources,omitempty"`
 	Settings        Settings           `json:"settings"`
+	// SolcStandardJson is used to pass raw input to the compiler. If it is specified, it will be directly passed to
+	// solc compiler via `--standard-json` argument.
+	SolcStandardJson *CompilerJsonInput `json:"solcStandardJson,omitempty"`
```
olegrok, line 147: I guess we need to add some sandboxing. I'm not sure but is it possible that contract import some sensitive data and save it into contract state during compilation?

Shouldn't we specify `--allow-paths` for protection?

olegrok (approved):

nil/services/cometa/types.go
shermike: I'm not sure if I got your point correctly. But by default, contracts are very confined, they can import files only from their directory. To add another directories for imports, you need explicitly pass it through command line argument `--allow-paths`. But Cometa doesn't use this parameter and `SolcStandardJson` doesn't contain such option. So I don't see the problem.

nil/services/cometa/compilation.go
shermike: Copilot is so copilot :)

PR #617, opened 2025-03-20
relayer: transaction sender impl

oclaw:
Implementation of relayer `TransactionSender` component

This component is responsible for:
- Fetching event from L2 storage
- Ordered forwarding of the events to the L2
- Dropping events from L2 storage

PR also contains L2 contract wrapper which provides an interface to interact with `L2BridgeMessenger` contract and some initial config for it (nil RPC endpoint, contract/account address, key). Current version only generates the key (if it is not exists). Other initialization stuff is going to be implemented in the following PRs

### DIFF
--- a/nil/services/relayer/internal/l2/contract_wrapper.go
+++ b/nil/services/relayer/internal/l2/contract_wrapper.go
@@ -1,0 +1,131 @@
+package l2
+
+import (
+	"context"
+	"crypto/ecdsa"
+	"errors"
+	"fmt"
+	"os"
+
+	"github.com/NilFoundation/nil/nil/client"
+	"github.com/NilFoundation/nil/nil/common"
+	"github.com/NilFoundation/nil/nil/internal/abi"
+	"github.com/NilFoundation/nil/nil/internal/types"
+	"github.com/ethereum/go-ethereum/crypto"
+)
+
+type ContractConfig struct {
+	Endpoint            string
+	SmartAccountAddress string
+	ContractAddress     string
+	PrivateKeyPath      string
+	ContractABIPath     string
+}
+
+func DefaultContractConfig() *ContractConfig {
+	return &ContractConfig{
+		PrivateKeyPath:  "relayer_key.ecdsa",
+		ContractABIPath: "l2_bridge_messenger_abi.json",
+	}
+}
+
+func (cfg *ContractConfig) Validate() error {
+	if len(cfg.Endpoint) == 0 {
+		return errors.New("empty L2 endpoint")
+	}
+	if len(cfg.SmartAccountAddress) == 0 {
+		return errors.New("empty relayer smart account address")
+	}
+	if len(cfg.PrivateKeyPath) == 0 {
+		return errors.New("empty relayer private key file path")
+	}
+	if len(cfg.ContractAddress) == 0 {
+		return errors.New("empty L2BridgeMessenger contract address")
+	}
+	if len(cfg.ContractABIPath) == 0 {
+		return errors.New("empty L2BridgeMessenger contract ABI path")
+	}
+	return nil
+}
+
+type L2Contract interface {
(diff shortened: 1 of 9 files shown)


nil/services/relayer/internal/l2/contract_wrapper.go
```diff
@@ -0,0 +1,131 @@
...
+	if err != nil {
+		return common.EmptyHash, err
+	}
+
+	return client.SendTransactionViaSmartAccount(
+		ctx,
+		w.nilClient,
+		w.smartAccountAddr,
+		calldata,
+		evt.FeePack, // TODO(oclaw) check me
```
oclaw, line 124: @defistar please take a look if I am using this field (and below) correctly

nil/services/relayer/internal/l2/transaction_sender.go
```diff
@@ -1,21 +1,159 @@
...
+			Int("event_count", len(droppingEvents)).
+			Msg("dropping events from L2 storage")
+
+		if err := ts.storage.DeleteEvents(ctx, droppingEvents); err != nil {
+			ts.logger.Warn().Err(err).Msg("failed to drop events from L2 storage")
+		}
+	}()
+
+	for i, evt := range events {
+		if _, err := ts.contractBinding.RelayMessage(ctx, evt); err != nil {
```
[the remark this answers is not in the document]
oclaw: We will, transaction sender anyway will drop all successfully sent events and try again from the point of failure during next iteration. Have a unit test for this :)

nil/services/relayer/internal/l2/transaction_sender.go
```diff
@@ -1,21 +1,159 @@
...
+	events := eventBySeqNumber.PopAllSorted()
+
+	if len(events) == 0 {
+		ts.logger.Debug().Msg("no ready events to be relayed to L2")
+		return nil
+	}
+
+	ts.logger.Info().
+		Int("fetched_events_count", len(events)).
+		Int("checked_events_count", eventsIterated).
```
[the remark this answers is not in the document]
oclaw: They are but only if total event count <= heap size. I will reflect this in metrics, but it might be useful to track this value in logs too (in case if we face some unexpected growth of the collection)

nil/services/relayer/service.go
```diff
@@ -75,27 +87,97 @@ func New(
...
+			return nil, err
+		} else {
+			if err := crypto.SaveECDSA(config.PrivateKeyPath, key); err != nil {
+				return nil, err
+			}
+		}
+		rs.Logger.Info().Msg("key generated")
+	}
+
+	// TODO(oclaw) here needs to be put initial L2 routine
```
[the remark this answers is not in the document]
oclaw: I have adjusted the comment. Deployment of the contract is not going to be mandatory relayer init part, but smart account & private key creation seem to belong here

nil/services/relayer/service.go
```diff
@@ -75,27 +87,95 @@ func New(
...
+			return nil, err
+		} else {
+			if err := crypto.SaveECDSA(config.PrivateKeyPath, key); err != nil {
+				return nil, err
+			}
+		}
+		rs.Logger.Info().Msg("key generated")
+	}
+
+	// TODO deploy relayer's L2 smart account (if it is not exist)
```
akokoshn, line 176: I guess here relayer shout check and fails if smart account not exists.
All deployments are CLI tool responsibility

oclaw: We did not decide on CLI tool actually, just to make it a common code component :)
Anyway I would like to split smart account related logic from this PR

akokoshn (approved):

nil/services/relayer/internal/l2/contract_wrapper.go
defistar: @oclaw 
this is correct sequence of fields.

PR #623, opened 2025-03-21
[stresser] Add ability to run single-instance as a service within Str…

shermike:
…esser's process


nil/services/stresser/stresser.go
```diff
@@ -221,9 +226,33 @@ func (s *Stresser) Run(ctx context.Context) error {
...
+			ExportMetrics: true,
+		}
+		nilConfig.RPCPort = 8529
+
+		go func() {
+			exitCode := nilservice.Run(ctx, nilConfig, badger, nil)
+			if exitCode != 0 {
+				logger.Error().Int("exit_code", exitCode).Msg("nilservice exited with error")
+			}
+			close(exitChan)
```
olegrok, line 250: You close this channel twice. Here and below. It causes panic.

olegrok (approved):

Issue #599, opened 2025-03-19
[explorer] Transaction page UI enhancements


mdaria17:
- [ ] Remove the back icon in the Explorer (will be using browser navigation)
- [ ] Remove 0 from log tab when there are no details – use [empty state](https://www.figma.com/design/bfoVui5t00ph56kr6qIJNq/UI-Bugs?node-id=53-972&t=8a8KNlQXcisBrvo8-4)  
- [ ] Don't show the outgoing transactions tab when there aren't any 

<img width="1650" alt="Image" src="https://github.com/user-attachments/assets/aa87d99e-808f-4edb-903a-b07b2adbeb32" />

<img width="887" alt="Image" src="https://github.com/user-attachments/assets/25edf635-ad3d-462f-b43d-0df887cf2012" />

JayWebtech:
Hello, my name is Jethro. I am a full-stack/blockchain developer with 6 years of experience building solution-oriented products.

I am also an expert in building dApps

I can confidently handle this issue.
Here is my portfolio and projects I have handled:

https://jaykosai.vercel.app
https://escrowlink.vercel.app
https://ecohotels.com
https://blockheader-wallet.vercel.app/
https://nijostam.org

0xAleksaOpacic:
Hey @JayWebtech , nice to meet you! 

I will assign this issue to you! Thank you

salazarsebas:
## My background

I'm a software engineer and passionate about solving problems, I've been contributing to open source software for a few months and it's an experience that has taught me a lot and I really like to contribute to other projects. I am a member of the Dojo Coding community⛩️. I will dedicate all my effort to complete this issue in the best way.

wheval:
Hello, I am Wheval, an experienced Frontend Developer with experience building with React, Typescript and Javascript.
I have made several contributions on Onlydust.

I can handle this issue properly.
Please assign.

idea404:
Hey @abdegenius thanks for indicating your interest in this issue. Happy to assign this issue to you. If you have any questions feel free to tag me, @0xAleksaOpacic or @gitshreevatsa in case of questions or reach us directly through our developer Telegram group which you can [find here](https://t.me/NilDevBot?start=ref_odhack)

PR #628, opened 2025-03-21
fix: Transaction page UI enhancements

abdegenius:
To close: #599 
The go back button is removed, Outgoing transactions tab is not visible if value is 0 and Logs is now rendering the right design when logs = 0


0xAleksaOpacic (changes requested):
Thanks for fixing the changes. Although i still see few issues regarding the padding
Screenshot 2025-03-26 at 15 02 38

Transaction page should have left and right padding 48px, also the Transaction payload should not overflow the page 

One more thing is that i see commits are not tidy, there are some conflicts with latest main (you can se 22 commints in commits tab) And there we should only have the once which contains your changes (not even merge commits)

PR #629, opened 2025-03-21
[mpt] Use tighter SSZ sizes, add test for path.Consume.

avm:


nil/internal/mpt/path.go
```diff
@@ -28,8 +27,10 @@ type PathAccessor interface {
 
 func newPath(data []byte, shifted bool) *Path {
 	if shifted {
-		check.PanicIfNot(len(data) > 0)
-		return &Path{Data: data[1:], Shifted: true, FirstNibble: data[0] & 0x0F} //nolint:gosec
+		if len(data) == 0 {
+			panic("shifted path must have at least one nibble")
+		}
```
olegrok, line 32: I'm not sure about this change. We usually use "check" module for that. But if you want to make this module independent of our internals it maybe ok.
```suggestion
		check.PanicIfNot(len(data) > 0, "shifted path must have at least one nibble")
```

olegrok (approved):

PR #631, opened 2025-03-21
[global refactoring] rename `cometaService`  to  `cometaClient`

strmfos:
global refactoring referencing the previous PR #574 to ensure a consistent rename from `cometaService` to `cometaClient`


knazarov:
The cometaClientEndpoint doesn't make sense. It does in fact refer to cometaService. The original rename to cometaClient was in relation to a JS class which was a client for this service. But the configuration and variables don't need a name change.

Issue #458, opened 2025-03-04
Cli.js: receipt command


Rolaman:
Implement a command to get transaction receipt. 

Reference: https://github.com/NilFoundation/nil/blob/main/nil/cmd/nil/internal/receipt/receipt.go

manoahLinks:
Could I be assigned to this?

Abeeujah:
Hey @Rolaman ,

I'd love to take on the implementation of the "fetch transaction receipt by hash" command for `clijs`.

I've reviewed the issue description and the reference Go implementation in `nil/cmd/nil/internal/receipt/receipt.go`, and I'm confident I can translate that logic into JavaScript within the `clijs` tool.

I'm eager to contribute to the project and believe this feature would be a valuable addition for users interacting with the =nil; through the CLI.

Please let me know if this is something I can proceed with.

Thanks!

saimeunt:
Could I grab this task?

manoahLinks:
i just raised a draft pr @0xAleksaOpacic @gitshreevatsa , please have a look at my implementation. thanks

gitshreevatsa:
@manoahLinks could you please link the PR to this issue, it will be easier to track and view.

Gezort:
#855

PR #636, opened 2025-03-23
Feat: [cli.js] receipt command

manoahLinks:
# Draft: Implement `getTransactionReceipt` Command in CLI

## Summary  
This PR introduces the `getTransactionReceipt` command to the CLI, allowing users to fetch a transaction receipt using its hash.

## Implementation Details  
- **Command Name**: `receipt [hash]`  
- **Retrieves Transaction Receipt**:  
  - Takes a **transaction hash** as input.  
  - Calls the relevant RPC function to fetch the receipt details.  
  - Displays the receipt in JSON format.  
- **Structure**:  
  - Uses **`oclif`** for command-line argument handling.  
  - Integrates with **`niljs`** to interact with the blockchain.  
  - Handles errors properly, ensuring invalid or empty hashes return appropriate messages.
  
  
Closes https://github.com/NilFoundation/nil/issues/458


Un1oR (changes requested):
All new commands should be added explicitly in [sea.ts](https://github.com/NilFoundation/nil/blob/2a07f70e29c2d59b1f8447bfa20be0771affe8db/clijs/src/sea.ts).
It's also worth adding tests.

gitshreevatsa:
closes https://github.com/NilFoundation/nil/issues/458

Rolaman:
Need at least 1 test

Rolaman (changes requested):
Tests

manoahLinks:
> Tests

i'm on it ser

clijs/src/sea.ts
```diff
@@ -53,6 +54,7 @@ export const COMMANDS: Record<string, Command.Class> = {
   "system:chain-id": ChainId,
   "system:gas-price": GasPrice,
   "system:shards": Shards,
+  "smart-account:get-transactin-reciept": SmartAccountGetTransactionReceipt
```
Un1oR, line 57: We're keeping the alphabetical order here

manoahLinks:
update made ser, command is added in alphabetical order

Un1oR (changes requested):
Wait, why did we even add this as a subcommand to `smart-account`? What does receipt have to do with smart account? The reference Go implementation mentioned in #458 has a top-level `receipt` command
```
$ nil receipt --help
Retrieve a receipt from the cluster

Usage:
  nil receipt [hash] [flags]

Flags:
  -h, --help   help for receipt

Global Flags:
  -c, --config string      The path to the config file (default "/home/un1or/.config/nil/config.ini")
  -l, --log-level string   Log level: trace|debug|info|warn|error|fatal|panic (default "info")
  -q, --quiet              Quiet mode (print only the result and exit)
  -v, --verbose            Verbose mode (print logs)
```

manoahLinks:
done, as stated above, reciepts is now its seperate command

Un1oR:
@manoahLinks see above I wrote the syntax for calling the command as it is now in Go: `nil receipt --help`. It should be the same in js. Is it the same now? Why do we need some kind of `get-transaction-receipt`?
Next, getting a receipt should have nothing to do with using a smart account. Right now it isn't, and you're calling `smartAccount.getTransactionReceipt`. This is not the way it should be. I guess you need to use `client.getTransactionReceiptByHash`.

idea404:
Closing as inactive

PR #638, opened 2025-03-23
[txpool] Fix data race in txpool

shermike:
Data race was in accessing `baseFee` field in the `TxnPool`.

### DIFF
--- a/nil/services/txnpool/txnpool.go
+++ b/nil/services/txnpool/txnpool.go
@@ -92,7 +92,8 @@ func (p *TxnPool) listen(ctx context.Context, sub *network.Subscription) {
 			continue
 		}
 
-		mm := newMetaTxn(txn, p.baseFee)
+		mm := newMetaTxn(txn, p.GetBaseFee())
+
 		reasons, err := p.add(mm)
 		if err != nil {
 			p.logger.Error().Err(err).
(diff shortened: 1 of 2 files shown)


nil/services/txnpool/txnpool.go
```diff
@@ -110,10 +113,16 @@ func (p *TxnPool) listen(ctx context.Context, sub *network.Subscription) {
...
+	if len(txns) == 0 {
+		return nil, nil
+	}
 	mms := make([]*metaTxn, len(txns))
+
+	p.lock.Lock()
 	for i, txn := range txns {
 		mms[i] = newMetaTxn(txn, p.baseFee)
 	}
+	p.lock.Unlock()
```
olegrok, line 125: Is it possible to add protected getter for this field?

```suggestion

func (p *TxnPool) GetBaseFee() (baseFee types.Value) {
    return p.baseFee
}

func (p *TxnPool) Add(ctx context.Context, txns ...*types.Transaction) ([]DiscardReason, error) {
	if len(txns) == 0 {
		return nil, nil
	}
	mms := make([]*metaTxn, len(txns))

	baseFee := p.GetBaseFee()
	for i, txn := range txns {
		mms[i] = newMetaTxn(txn, baseFee)
	}
```

olegrok (approved):

PR #640, opened 2025-03-24
refactor(sc): split contract CommitBatch and UpdateState calls

x-mass:
- Refactored CommitBatch and UpdateState methods:
  - CommitBatch now called in the aggregator service just before creating proof tasks. Transaction creation is skipped if a batch with the same ID has already been submitted.
  - UpdateState is still called in the proposer service after it mentions there is a new state to be proposed.
  - UpdateState now retrieves the CommitBatch transaction for the specified batch ID from L1 and extracts blob data.
  - Instead of returning etherium transaction methods wait for transaction for being included into block.

- Improved error handling and configuration:
  - Proposer now handles `ErrBatchAlreadyFinalized` by refreshing the last proved state root from L1 if local tasks state is out of sync.
  - Introduced a dummy structure mode for the EthClient when the DisableL1 config option is set, which logs calls without returning errors.

- Optimized contract and client interactions:
  - Shared the contract wrapper between Aggregator and Proposer to streamline interactions.
  - Encapsulated retry logic within the EthClient, which now acts as a wrapper around go-ethereum/ethclient.

- Extended testing and utilities:
  - Migrated contract mocker to the testaide package for broader reuse in tests.
  - Added new tests for both the proposer and the wrapper to ensure robustness.


nil/services/synccommittee/internal/testaide/contract.go
```diff
@@ -0,0 +1,74 @@
...
+	"context"
+	"fmt"
+	"math/big"
+
+	ethereum "github.com/ethereum/go-ethereum"
+	"github.com/ethereum/go-ethereum/accounts/abi"
+)
+
+type CallContractMock struct {
+	methodsReturnValue map[string][][]interface{}
```
oclaw, line 15: any

nil/services/synccommittee/core/service.go
```diff
@@ -46,24 +46,31 @@ func New(cfg *Config, database db.DB, ethClient rollupcontract.EthClient) (*Sync
...
 		taskStorage,
 		stateResetter,
+		rollupContractWrapper,
 		clock,
 		logger,
 		metricsHandler,
 		cfg.AggregatorConfig,
 	)
 
-	ctx := context.Background()
```
oclaw, line 60: let's keep this variable, otherwise there is a danger to mix different ctx`es inside the function (if we need one somewhere)

nil/services/synccommittee/internal/rollupcontract/eth_client.go
```diff
@@ -2,15 +2,180 @@ package rollupcontract
...
+	ctx context.Context, endpoint string, requestsTimeout time.Duration, logger logging.Logger,
+) (EthClient, error) {
+	ethClient, err := ethclient.DialContext(ctx, endpoint)
+	if err != nil {
+		return nil, fmt.Errorf("connecting to ETH RPC node: %w", err)
+	}
+
+	retryRunner := common.NewRetryRunner(
+		common.RetryConfig{
+			ShouldRetry: common.LimitRetries(5),
```
oclaw, line 35: Is there any way to understand if error from ethclient is temporary or not? Adjusting retry policy would be nice here

nil/services/synccommittee/internal/rollupcontract/wrapper.go
```diff
@@ -16,121 +15,159 @@ import (
...
-type Wrapper struct {
+type Wrapper interface {
+	UpdateState(
+		ctx context.Context,
+		batchIndex string,
+		oldStateRoot, newStateRoot common.Hash,
+		validityProof []byte,
+		publicDataInputs INilRollupPublicDataInfo,
+	) error
+	FinalizedStateRoot(ctx context.Context, finalizedBatchIndex string) ([32]byte, error)
```
oclaw, line 29: Don't we have some alias for the state root instead of [32]byte? If not - makes sense to add one

nil/services/synccommittee/internal/rollupcontract/update_state.go
```diff
@@ -0,0 +1,197 @@
...
+	batchState := &batchState{}
+
+	// Check if batch is finalized
+	isFinalized, err := r.rollupContract.IsBatchFinalized(r.getEthCallOpts(ctx), batchIndex)
+	if err != nil {
+		return nil, err
+	}
+	batchState.IsFinalized = isFinalized
+
+	if batchState.IsFinalized {
```
akokoshn, line 147: i guess we should not return error here, just check `IsFinalized` in caller

nil/services/synccommittee/core/proposer.go
```diff
@@ -207,9 +146,17 @@ func (p *proposer) proposeNextBatch(ctx context.Context) error {
 		return nil
 	}
 
-	err = p.sendProof(ctx, data)
+	err = p.updateState(ctx, data)
 	if err != nil {
-		return fmt.Errorf("failed to send proof to L1 for batch with id=%s: %w", data.BatchId, err)
+		if errors.Is(err, rollupcontract.ErrBatchAlreadyFinalized) {
```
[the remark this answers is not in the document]
zadykian: Sync Committee can resubmit state update for the same batch in some cases (network issues / process termination / etc)

For example:
```
Proposer iteration is started
    -> Proposer calls rollupcontract.UpdateState()
    -> State on the contract side is updated
    -> Sync Committee process is terminated right before `storage.SetBatchAsProposed()` is called
```

Also `WaitForReceipt()` can fail due to timeout

nil/services/synccommittee/core/proposer.go
```diff
@@ -112,15 +84,15 @@ func (*proposer) Name() string {
 }
 
 func (p *proposer) Run(ctx context.Context, started chan<- struct{}) error {
-	if err := p.initializeProvedStateRoot(ctx); err != nil {
-		return err
-	}
-
 	close(started)
```
zadykian, line 87: `started` channel should be closed after `updateStoredStateRootFromL1` is completed successfully, since it's part of proposer's initialization

nil/services/synccommittee/core/proposer.go
```diff
@@ -131,30 +103,14 @@ func (p *proposer) Run(ctx context.Context, started chan<- struct{}) error {
 	return nil
 }
 
-func (p *proposer) initRollupContractWrapper(ctx context.Context) error {
-	if p.rollupContractWrapper != nil || p.params.DisableL1 {
-		return nil
-	}
+func (p *proposer) updateStoredStateRootFromL1(ctx context.Context) error {
```
zadykian, line 106: This naming a little bit misleading, stored state can be fetched both from L1 and L2
I'd suggest just `updateStoredStateRoot`

nil/services/synccommittee/core/proposer.go
```diff
@@ -207,9 +146,17 @@ func (p *proposer) proposeNextBatch(ctx context.Context) error {
 		return nil
 	}
 
-	err = p.sendProof(ctx, data)
+	err = p.updateState(ctx, data)
 	if err != nil {
-		return fmt.Errorf("failed to send proof to L1 for batch with id=%s: %w", data.BatchId, err)
+		if errors.Is(err, rollupcontract.ErrBatchAlreadyFinalized) {
+			// someone has already updated, we need to refetch state from contract
+			p.logger.Warn().Msg("batch is already committed, skipping UpdateState tx")
```
zadykian, line 153: Message should be changed here

nil/services/synccommittee/internal/rollupcontract/wrapper.go
```diff
@@ -16,121 +15,159 @@ import (
...
+	transactOpts.Context = ctx
+	return transactOpts, nil
 }
 
-func (r *Wrapper) StateRoots(ctx context.Context, finalizedBatchIndex string) ([32]byte, error) {
-	callOpts, cancel := r.getEthCallOpts(ctx)
-	defer cancel()
-	return r.rollupContract.FinalizedStateRoots(callOpts, finalizedBatchIndex)
-}
+func (r *wrapperImpl) transactWithTimeout(ctx context.Context, transactFunc contractTransactFunc) error {
```
zadykian, line 154: Maybe rename this one to `transactWithCtx`? No timeout is explicitly configured here, just attaching the current `ctx`

nil/services/synccommittee/core/proposer.go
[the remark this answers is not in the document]
akokoshn: In this case we have check if already submitted `state root `for particular `batchId` the same as proposer want to send

nil/services/synccommittee/core/service.go
x-mass: Didn't get you. Make this a `_ := context.Background()` in case we will need this in future?

nil/services/synccommittee/internal/rollupcontract/eth_client.go
x-mass: Besides network errors, JSON marshalling and unmarshalling errors are possible. These errors often occur when an RPC node returns malformed data. For instance, receiving a '429 Too Many Requests' message can typically be resolved with a retry.
I think, better to keep it like this and adjust if we encounter non-retryable errors later, since it's not obvious which of them are really so.

nil/services/synccommittee/core/proposer.go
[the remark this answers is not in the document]
x-mass: It could be finalized with the same root, just by another actor. For example, we accidentally launch two instances ofr proposer.
I'm not sure it could be done with different state root. As I understand, verification should fail in this case?

nil/services/synccommittee/core/proposer.go
x-mass: Made it `updateStoredStateRootFromContract` insted, so we couild distinguish this update and update during new state root proposal

nil/services/synccommittee/core/proposer.go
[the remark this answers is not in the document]
akokoshn: Yes, data proof verification should fails, but in this case contract reject transaction before.

nil/services/synccommittee/core/service.go
zadykian: Yes, top-level `ctx := context.Background()` can be declared [here](https://github.com/NilFoundation/nil/blob/main/nil/cmd/sync_committee/main.go#L124) and passed both to `core.New(...)` and to `service.Run(...)`

zadykian (approved):

PR #643, opened 2025-03-24
[stresser] fix waiting for receipt on contract deploy

olegrok:
`WaitTx` was implemented in a way when sometimes it could fail with DestinationContractDoesNotExist error. That happened because we need to be sure that all messages in receipt is completed (not only top-level exteral message).

### DIFF
--- a/nil/services/stresser/core/helper.go
+++ b/nil/services/stresser/core/helper.go
@@ -83,9 +83,9 @@ func (h *Helper) DeployContract(name string, shardId types.ShardId) (*Contract, 
 
 	tx, addr, err := h.Client.DeployExternal(h.ctx, shardId, payload, types.NewFeePackFromGas(100_000_000))
 	if err != nil {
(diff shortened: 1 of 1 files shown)


nil/services/stresser/core/helper.go
```diff
@@ -135,7 +135,7 @@ func (h *Helper) TopUp(addr types.Address, value types.Value) error {
 		return fmt.Errorf("failed to top up via faucet: %w", err)
 	} else {
 		if receipt, err := h.WaitTx(tx); err != nil {
-			return fmt.Errorf("failed to get receipt during top up: %w", err)
+			return fmt.Errorf("failed to get receipt %s during top up: %w", tx, err)
 		} else if !receipt.Success {
```
shermike, line 139: Can you also add this change then?
```suggestion
		} else if !receipt.AllSuccess() {
```

PR #647, opened 2025-03-24
Limit number of tasks for proof provider

makxenov:
Currently proof provider fetches new batch tasks regardless of number of prover tasks in the queue. This commit introduces a limit for the queue.

### DIFF
--- a/nil/cmd/proof_provider/main.go
+++ b/nil/cmd/proof_provider/main.go
@@ -76,6 +76,12 @@ func addFlags(cmd *cobra.Command, cfg *cmdConfig) {
 		cfg.SkipRate,
 		"rate of skip tasks, will skip N from 10, where N is value of option (0 means no skip)."+
 			" Possible values: [0,10]")
+	cmd.Flags().Uint32Var(
+		&cfg.MaxConcurrentBatches,
+		"max-concurrent-batches",
+		cfg.MaxConcurrentBatches,
+		"maximum value of batches that proof provider can handle concurrently",
+	)
 	logLevel := cmd.Flags().String(
 		"log-level",
 		"info",
--- a/nil/services/synccommittee/internal/executor/task_executor_test.go
+++ b/nil/services/synccommittee/internal/executor/task_executor_test.go
@@ -25,7 +25,9 @@ type TestSuite struct {
 func (s *TestSuite) SetupTest() {
 	s.context, s.cancellation = context.WithCancel(context.Background())
 	s.requestHandler = newTaskRequestHandlerMock()
-	s.taskHandler = &api.TaskHandlerMock{}
+	s.taskHandler = &api.TaskHandlerMock{
+		IsReadyToHandleFunc: func(ctx context.Context) (bool, error) { return true, nil },
+	}
 
 	config := Config{
 		TaskPollingInterval: 10 * time.Millisecond,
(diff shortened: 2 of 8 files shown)


nil/services/synccommittee/proofprovider/service.go
```diff
@@ -28,6 +29,7 @@ func NewDefaultConfig() *Config {
 		SyncCommitteeRpcEndpoint: "tcp://127.0.0.1:8530",
 		TaskListenerRpcEndpoint:  "tcp://127.0.0.1:8531",
 		SkipRate:                 0,
+		MaxActiveTasks:           100,
```
akokoshn, line 32: I guess 1 active task should be enough for test setup

nil/services/synccommittee/proofprovider/task_handler.go
```diff
@@ -13,38 +15,60 @@ import (
...
+		logger:         logger,
 	}
 }
 
+func (h *taskHandler) IsReadyToHandle(ctx context.Context) (bool, error) {
+	active, err := h.taskStorage.GetActiveTaskCount(ctx)
+	if err != nil {
+		return false, err
+	}
+	return active <= h.maxActiveTasks, nil
```
akokoshn, line 55: may be `<`, otherwise when `active == h.maxActiveTasks` will be handled max + 1 tasks

nil/services/synccommittee/proofprovider/task_handler.go
```diff
@@ -13,38 +15,60 @@ import (
...
+}
+
 func (h *taskHandler) Handle(ctx context.Context, executorId types.TaskExecutorId, task *types.Task) error {
 	if task.TaskType != types.ProofBatch {
 		return types.NewTaskErrNotSupportedType(task.TaskType)
 	}
+
+	// Check if max limit of tasks is not exceeded
+	ready, err := h.IsReadyToHandle(ctx)
```
akokoshn, line 64: When limit archived service should not fetch new tasks, otherwise the tasks will be skipped.
Need set this condition here: https://github.com/NilFoundation/nil/blob/46acd00066b9994b13175d48602368f329b2ebcf/nil/services/synccommittee/internal/executor/task_executor.go#L89

nil/cmd/proof_provider/main.go
```diff
@@ -76,6 +76,12 @@ func addFlags(cmd *cobra.Command, cfg *cmdConfig) {
 		cfg.SkipRate,
 		"rate of skip tasks, will skip N from 10, where N is value of option (0 means no skip)."+
 			" Possible values: [0,10]")
+	cmd.Flags().IntVar(
+		&cfg.MaxActiveTasks,
+		"max-active-tasks",
```
zadykian, line 81: I suggest adjusting the approach: replace the `max-active-tasks` parameter with `max-concurrent-batches` as  concurrency limit

This value can be determined by the number of active `ProofBatch` tasks that have already been requested from the `Sync Committee`

For that we'd need to store `ProofBatch` tasks in `ProofProvider`'s storage along with child tasks for `Prover` [here](https://github.com/NilFoundation/nil/blob/limit-proofprovider-tasks/nil/services/synccommittee/proofprovider/task_handler.go#L100). Corresponding `TaskEntry` should have an `Active` status and be assigned to the `ProofProvider` itself

Otherwise, managing concurrency becomes difficult because each new request for `ProofBatch` tasks leads to creation of N new tasks for `Prover` nodes

For example, if there are 99 active tasks and the limit is 100, `IsReadyToHandle()` would return `true`, a new `BatchProof` task would be requested, and we'd end up with ~110 tasks in storage

nil/services/synccommittee/internal/executor/task_executor_test.go
```diff
@@ -98,3 +100,23 @@ func (s *TestSuite) Test_TaskExecutor_Executes_Tasks() {
...
+	defer cancelFunc()
+	started := make(chan struct{})
+	go func() {
+		_ = s.taskExecutor.Run(ctx, started)
+	}()
+	s.Require().Never(
+		func() bool {
+			return len(s.taskHandler.HandleCalls()) != 0
+		},
+		5*time.Second,
```
zadykian, line 119: Let's reduce this interval to let tests pass faster
One second should be enough

nil/services/synccommittee/internal/storage/task_storage.go
```diff
@@ -148,6 +148,30 @@ func (st *TaskStorage) TryGetTaskEntry(ctx context.Context, id types.TaskId) (*t
 	return entry, err
 }
 
+// GetActiveTaskCount
+func (st *TaskStorage) GetActiveTaskCount(ctx context.Context) (int, error) {
```
zadykian, line 152: This method could accept `TaskType` param to align with the approach mentioned above

zadykian: Now `TaskStorage` has [GetStats](https://github.com/NilFoundation/nil/blob/main/nil/services/synccommittee/internal/storage/task_storage.go#L66) method, you can use it instead :)

nil/cmd/proof_provider/main.go
makxenov: Reworked to check count of `MergeProof` tasks

PR #649, opened 2025-03-24
[exporter, execution] Small improve in exporter, enable `assert` tags by default

shermike:
- Remove unnecessary receipts map because the receipt is placed in the same index as the corresponding transaction.
- Set tag `assert` by default for all binaries. ~~Add an additional check that the transaction is always unique in the database.~~

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ test: generated
 %.cmd: generated
 	@# Note: $* is replaced by the command name
 	@echo "Building $*"
-	@cd ./nil/cmd/$* && $(GOBUILD) -o $(GOBIN)/$*
+	@cd ./nil/cmd/$* && $(GOBUILD) -o $(GOBIN)/$* -tags assert
 	@echo "Run \"$(GOBIN)/$*\" to launch $*."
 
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -23,7 +23,7 @@ test: generated
 %.cmd: generated
 	@# Note: $* is replaced by the command name
 	@echo "Building $*"
-	@cd ./nil/cmd/$* && $(GOBUILD) -o $(GOBIN)/$*
+	@cd ./nil/cmd/$* && $(GOBUILD) -o $(GOBIN)/$* -tags assert
```
olegrok, line 26: It affects only local build. Asserts will be still disabled on nil-dev/prod.

shermike: You're right, thanks. Added option for nix build.

shermike:
@olegrok is the duplicated transaction bug already fixed in the main? Otherwise this patch will fail.

olegrok:
> @olegrok is the duplicated transaction bug already fixed in the main? Otherwise this patch will fail.

Not yet

PR #652, opened 2025-03-25
[collate] workaround doubled tx

olegrok:
During the replay block, we do not update the collator state, which may cause the proposer to include transactions that have already been executed.

This patch adds a check to detect already executed transactions — such messages will now be skipped.

### DIFF
--- a/nil/internal/collate/proposer.go
+++ b/nil/internal/collate/proposer.go
@@ -439,9 +439,6 @@ func (p *proposer) handleTransactionsFromNeighbors(tx db.RoTx) error {
 
 				if txn.To.ShardId() == p.params.ShardId {
 					// TODO: Temporary workaround to prevent transaction duplication
-					// This check also breaks refunds from not initialized logic
-					//nolint: lll
-					// See https://github.com/NilFoundation/nil/blob/06885370c1efc4c5990f52869f2b03d14d102d85/nil/internal/execution/state.go#L901-L905
 					isProcessed, err := p.isTxProcessed(tx, txn.Hash())
 					if err != nil {
 						return err
--- a/nil/internal/collate/proposer.go
+++ b/nil/internal/collate/proposer.go
@@ -351,6 +351,10 @@ func (p *proposer) handleTransactionsFromPool() error {
 	return nil
 }
 
+func (p *proposer) isTxProcessed(dbTx db.RoTx, txHash common.Hash) (bool, error) {
+	return dbTx.ExistsInShard(p.params.ShardId, db.BlockHashAndInTransactionIndexByTransactionHash, txHash.Bytes())
+}
+
 func (p *proposer) handleTransactionsFromNeighbors(tx db.RoTx) error {
 	state, err := db.ReadCollatorState(tx, p.params.ShardId)
 	if err != nil && !errors.Is(err, db.ErrKeyNotFound) {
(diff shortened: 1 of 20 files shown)


shermike:
Could you please also add this check to ensure we don't have duplicate transactions in the future?
```diff
diff --git a/nil/internal/execution/postprocess.go b/nil/internal/execution/postprocess.go
index b38cbd91..b0025418 100644
--- a/nil/internal/execution/postprocess.go
+++ b/nil/internal/execution/postprocess.go
@@ -56,6 +56,12 @@ func (pp *blockPostprocessor) fillBlockHashAndTransactionIndexByTransactionHash(
                                return err
                        }
 
+                       if assert.Enable {
+                               if _, err := pp.tx.Get(db.TableName(table), hash.Bytes()); err == nil {
+                                       return fmt.Errorf("fatal: duplicate transaction %x", hash)
+                               }
+                       }
+
                        if err := pp.tx.PutToShard(pp.shardId, table, hash.Bytes(), value); err != nil {
                                return err
                        }
```

olegrok:
> Could you please also add this check to ensure we don't have duplicate transactions in the future?
> 
> ```diff
> diff --git a/nil/internal/execution/postprocess.go b/nil/internal/execution/postprocess.go
> index b38cbd91..b0025418 100644
> --- a/nil/internal/execution/postprocess.go
> +++ b/nil/internal/execution/postprocess.go
> @@ -56,6 +56,12 @@ func (pp *blockPostprocessor) fillBlockHashAndTransactionIndexByTransactionHash(
>                                 return err
>                         }
>  
> +                       if assert.Enable {
> +                               if _, err := pp.tx.Get(db.TableName(table), hash.Bytes()); err == nil {
> +                                       return fmt.Errorf("fatal: duplicate transaction %x", hash)
> +                               }
> +                       }
> +
>                         if err := pp.tx.PutToShard(pp.shardId, table, hash.Bytes(), value); err != nil {
>                                 return err
>                         }
> ```

Your patch is incorrect a bit. It should be pp.tx.GetInShard. Otherwise this condition will be always true.
I tried to do that but it break a lot. It break collator tests because we apply the same transaction. And it breaks validator/replay modes. I can try to solve it a bit later.

shermike:
> Also we have refunds from not-intiialized accounts -

I think you can remove this condition and add a check that the account is already created:
```diff
diff --git a/nil/internal/execution/state.go b/nil/internal/execution/state.go
index 10ae4f8d..66ec217c 100644
--- a/nil/internal/execution/state.go
+++ b/nil/internal/execution/state.go
@@ -898,20 +898,19 @@ func (es *ExecutionState) AddOutTransaction(
                return nil, err
        }
 
-       // TODO:        This happens when we send refunds from uninitialized accounts when transferring money to them.
-       //                      For now we will write all such refunds with identical zero seqno, because we can't change seqno
-       //                      of uninitialized accounts.
-       //                      In future we should add transfer that is free on the reciepient's side, so that these transfers
-       //                      won't require refunds.
-       if seqno != 0 {
-               if seqno+1 < seqno {
-                       return nil, vm.ErrNonceUintOverflow
-               }
-               if err := es.SetSeqno(caller, seqno+1); err != nil {
-                       return nil, err
+       if assert.Enable {
+               if a, err := es.GetAccount(caller); err != nil || a == nil {
+                       return nil, fmt.Errorf("account does not exist: %s", caller.Hex())
                }
        }
 
+       if seqno+1 < seqno {
+               return nil, vm.ErrNonceUintOverflow
+       }
+       if err := es.SetSeqno(caller, seqno+1); err != nil {
+               return nil, err
+       }
+
        txn := payload.ToTransaction(caller, seqno)
 
        // Propagate fee settings from the inbound message
```

shermike (approved):
Thanks!

PR #655, opened 2025-03-25
[journald-forwarder] Check logs write success

Zerg1996:
In case logs can't be handle successfully return error to otel contrib collector, so in this can he can try to send them later.

### DIFF
--- a/nil/services/journald_forwarder/main.go
+++ b/nil/services/journald_forwarder/main.go
@@ -3,6 +3,7 @@ package journald_forwarder
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net"
 	"slices"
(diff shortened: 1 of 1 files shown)


nil/services/journald_forwarder/main.go
```diff
@@ -148,74 +146,75 @@ func storeData(ctx context.Context, logger logging.Logger, connect driver.Conn,
 		strValue, ok := value.(string)
 		if !ok {
 			logger.Error().Msg("Timestamp is not a string")
```
olegrok, line 148: You can write actual value. And it will help during issue debugging.

olegrok (commented):
To be honest I don't understand initial idea.
Don't drop logs in case of schema mismatch, am I right? But in this case who is responsible to fix this schema?
Is OOM possible in this case (we can't keep buffer infinitely). Do we limit amount of retries somehow?

nil/services/journald_forwarder/main.go
Zerg1996: Oh, thanks! Will fix it

>initial idea
Don't miss logs if they wasn't handle for example when clickhouse for some reason wasn't available.

>we limit amount of retries somehow
yes, it's limiting inside otelcol-contrib config. We can setup time for send repeating and batch amount of logs that we can loose when the time expire. Now it's set to unlimited.

> who is responsible to fix this schema
We are :) it's better to understand that we have a problem and fix it than start loosing some logs.

olegrok (approved):

PR #661, opened 2025-03-25
feat: make transaction methods awaitable

### LINKED ISSUE #173: [nil.js] Make sendMessage in nil.js awaitable
Now we can receive recipe of transaction like this:

```ts
const hash = await wallet.sendMessage({
  to: anotherAddress,
  value: 10_000_000n,
  gas: 100_000n,
});

await waitTillCompleted(client, 1, hash);
```

There is a proposal to change api accordingly with the following example

```ts
const tx = await wallet.sendMessage({...});
await tx.waitTillCompleted();
```

`sendMessage` can return object what can be awaited. We need some testing and usability testing to prove this idea. Probably there are cases when `await waitTillCompleted(client, 1, hash)` is more convenient.

manassehO:
# Make Transaction Methods Awaitable

## Changes
- Added `Transaction` class to handle transaction completion
- Updated all transaction methods to return `Transaction` objects instead of raw hashes
- Added `waitTillCompleted()` method to `Transaction` class
- Updated tests to handle new return types

## Usage
```typescript
// Old
const hash = await wallet.sendTransaction({...});
await waitTillCompleted(client, 1, hash);

// New
const tx = await wallet.sendTransaction({...});
await tx.waitTillCompleted();
```

Closes #173


manassehO:
Ok, so most of these test are failing because the update was a breaking changes that returns `Transaction` type rather than  `0xstring`

niljs/src/utils/transaction.ts
```diff
@@ -0,0 +1,70 @@
...
+
+/**
+ * Represents a transaction that can be awaited for completion
+ */
+export class TransactionResponse {
+  public readonly hash: Hex;
+  constructor(
+    private readonly tx: Uint8Array,
+    private readonly client: PublicClient,
+    private readonly shardId: number,
```
KlonD90, line 15: Yo don't need shardId here. Shard id also part of tx. Also it should be still be awaitable by waitTillCompleted (and you should update this method also to work with Hex and Transaction) and it should be upgraded.

niljs/src/utils/transaction.ts
```diff
@@ -0,0 +1,70 @@
...
+  async waitTillCompleted(options?: { waitTillMainShard?: boolean }): Promise<ProcessedReceipt[]> {
+    return await waitTillCompleted(this.client, this.hash, options);
+  }
+
+  /**
+   * Waits until the transaction is completed
+   *
+   * @notice same logic as waitTillCompleted, more compatible with ethers.js libraries
+   */
+  async wait(options?: { waitTillMainShard?: boolean }): Promise<ProcessedReceipt[]> {
```
KlonD90, line 39: It's better to extract types from waitTillComplted

niljs/src/utils/transaction.ts
```diff
@@ -0,0 +1,70 @@
+import type { PublicClient } from "../clients/PublicClient.js";
+import { bytesToHex } from "../encoding/fromBytes.js";
+import type { Hex } from "../types/Hex.js";
+import type { ProcessedReceipt } from "../types/IReceipt.js";
+import { waitTillCompleted } from "./receipt.js";
+
+/**
+ * Represents a transaction that can be awaited for completion
+ */
+export class TransactionResponse {
```
KlonD90, line 10: It's better to be named Transaction

niljs/src/utils/transaction.ts
```diff
@@ -0,0 +1,70 @@
...
+   * Returns the string representation of the transaction hash
+   */
+  toString(): Hex {
+    return this.hash;
+  }
+
+  /**
+   * Converts the transaction to a hex string for backward compatibility
+   */
+  toHexString(): Hex {
```
KlonD90, line 53: We don't need two the same method. remove this. extended APi make it harder to support

niljs/src/utils/transaction.ts
```diff
@@ -0,0 +1,70 @@
...
+    return await waitTillCompleted(this.client, this.hash, options);
+  }
+
+  /**
+   * Waits until the transaction is completed
+   *
+   * @notice same logic as waitTillCompleted, more compatible with ethers.js libraries
+   */
+  async wait(options?: { waitTillMainShard?: boolean }): Promise<ProcessedReceipt[]> {
+    return await waitTillCompleted(this.client, this.hash, options);
```
KlonD90, line 40: you don't need await here

niljs/src/utils/transaction.ts
```diff
@@ -0,0 +1,70 @@
...
+   * Returns the transaction hash
+   */
+  getHash(): Hex {
+    return this.hash;
+  }
+
+  /**
+   * Waits until the transaction is completed
+   */
+  async waitTillCompleted(options?: { waitTillMainShard?: boolean }): Promise<ProcessedReceipt[]> {
```
KlonD90, line 30: remove this one to avoid method duplication.

niljs/src/utils/transaction.ts
```diff
@@ -0,0 +1,70 @@
...
+import type { ProcessedReceipt } from "../types/IReceipt.js";
+import { waitTillCompleted } from "./receipt.js";
+
+/**
+ * Represents a transaction that can be awaited for completion
+ */
+export class TransactionResponse {
+  public readonly hash: Hex;
+  constructor(
+    private readonly tx: Uint8Array,
```
KlonD90, line 13: Also this is not a tx. it's just a hash in this case. So it's better to be named txHash

niljs/src/utils/transaction.ts
```diff
@@ -0,0 +1,70 @@
...
+    private readonly shardId: number,
+  ) {
+    this.hash = bytesToHex(this.tx);
+  }
+
+  /**
+   * Returns the transaction hash
+   */
+  getHash(): Hex {
+    return this.hash;
```
KlonD90, line 24: let's skip this method because we already expose it through hash

niljs/src/utils/transaction.ts
KlonD90: Or maybe better TransactionWaiter. It's better to have more infromation but leave it for future things

ukorvl:
Maybe `waitTillCompleted` -> `waitTillCompletion`?

manassehO:
> Maybe `tx.waitTillCompleted()` -> `tx.waitTillCompletion()`?

i do not completely understand this. a question?

niljs/src/utils/transaction.ts
manassehO: removing this in favour of wait.

usage `tx.wait()`

niljs/src/utils/transaction.ts
```diff
@@ -0,0 +1,41 @@
...
+   * await transaction.wait();
+   */
+  async wait(options?: { waitTillMainShard?: boolean, interval?: number }): Promise<ProcessedReceipt[]> {
+    return waitTillCompleted(this.client, this.hash, options);
+  }
+
+  /**
+   * Returns the transaction as a Uint8Array
+   */
+  toUint8Array(): Uint8Array {
```
KlonD90, line 38: let's delete this one method. It could mislead so it's better to avoid exposing mislead function

niljs/src/utils/transaction.ts
```diff
@@ -0,0 +1,41 @@
...
+   *
+   * @async
+   * @param {Object} options - The options for the wait operation.
+   * @param {boolean} options.waitTillMainShard - Whether to wait until the transaction is processed on the main shard.
+   * @param {number} options.interval - The interval to check the transaction status.
+   * @returns {Promise<ProcessedReceipt[]>} A promise that resolves to an array of processed receipts.
+   * @example
+   * await transaction.wait();
+   */
+  async wait(options?: { waitTillMainShard?: boolean, interval?: number }): Promise<ProcessedReceipt[]> {
```
KlonD90, line 31: You should infer options from waitTillCompleted

idea404:
Hey @manassehO if you have any questions and would like direct answer just reach out to us in the TG dev group you can find through the [TG RPC Bot](https://t.me/NilDevBot?start=ref_odhack)

KlonD90 (approved):
lgtm

manassehO:
Run Lint to possibly fix some errors

@KlonD90

ukorvl:
@manassehO i see that some workflows fail because of incorrect code format. To fix that is should be enough to run `pnpm run lint:fix` in the root.

idea404:
A couple additions to the above: 

- You still have to change the old `getHash()` syntax to the new one in `./niljs/src/smart-accounts/SmartAccountV1/SmartAccountV1.test.ts`
- You might run into an error when doing `pnpm run lint:fix` with missing config in `./rollup-bridge-contracts`. You can safely ignore the dir you can add in package.json as filter, or just run: `pnpm --filter '!./rollup-bridge-contracts' -r run lint:fix`

manassehO:
> * getHash()

Done. Thank you for pointing out...

ukorvl:
Hey @manassehO, i signed your commit and gonna merge. Thanks for your contribution.

PR #665, opened 2025-03-26
Sync Committee: Observable Counters in TaskStorage

zadykian:
### Sync Committee: Observable Counters in TaskStorage

Currently, `Current Active Tasks` and `Current Pending Tasks` metric values can go below zero in SigNoz because they are both based on the OpenTelemetry `UpDownCounter`, which is reset to zero on each process restart

This commit fixes the issue by:

* Extending the TaskStorage API with a `GetStats()`` method, which returns the total number of active and pending tasks, grouped by type and executor;
* Switching to `ObservableUpDownCounter` with a callback that calls `GetStats()`;


nil/services/synccommittee/internal/metrics/task_storage_metrics.go
```diff
@@ -66,18 +84,55 @@ func (h *taskStorageMetricsHandler) init(attributes metric.MeasurementOption, me
...
-	h.totalTasksCreated.Add(ctx, 1, h.attributes, taskAttributes)
-	h.currentPendingTasks.Add(ctx, 1, h.attributes, taskAttributes)
+func (h *taskStorageMetricsHandler) registerStatsCallback(meter telemetry.Meter) error {
+	_, err := meter.RegisterCallback(
+		func(ctx context.Context, observer metric.Observer) error {
+			var provider types.TaskStatsProvider
+			h.initMutex.RLock()
+			provider = h.provider
+			h.initMutex.RUnlock()
+			if provider == nil {
```
akokoshn, line 94: seems we cam check `h.provider` before lock mutex

nil/services/synccommittee/internal/metrics/task_storage_metrics.go
```diff
@@ -66,18 +84,55 @@ func (h *taskStorageMetricsHandler) init(attributes metric.MeasurementOption, me
...
+				return nil
+			}
+
+			stats, err := provider.GetTaskStats(ctx)
```
akokoshn, line 98: not exactly understand why need to protect copy of interface `TaskStatsProvider` instead call `h.provider.GetTaskStats`

zadykian: Synchronization is needed here because metrics callback, which uses `provider`, and method `SetStatsProvider(...)`, which sets it, are executed by two different goroutines

nil/services/synccommittee/internal/metrics/task_storage_metrics.go
zadykian: In that case we'd get a race + interface assignment is not atomic in Go

nil/services/synccommittee/internal/metrics/task_storage_metrics.go
zadykian: Replaced `provider`, `initMutex` and `initOnce` fields with single `atomic.Value` field for simplicity

akokoshn (approved):

PR #669, opened 2025-03-26
Bundle explorer backend into a single .js file

knazarov:
There was a long-standing problem with how we deliver explorer backend to prod machines. That's because it used to be delivered in the source form and we executed "npm" at the spot. This introduces flakiness and generally makes distributing a single deb package difficult.

This diff uses "rollup" JS module to bundle the explorer backend into a single file which can be just executed with `node explorer.js`.


explorer_backend/package.json
```diff
@@ -48,7 +40,8 @@
 		"viem": "^2.17.3",
 		"whatwg-fetch": "^3.6.2",
 		"yaml": "^2.3.1",
-		"zod": "^3.22.2"
+		"zod": "^3.22.2",
+    "@rollup/plugin-json": "^6.1.0"
```
ukorvl, line 44: can this be a dev dependency?

explorer_backend/tsconfig.json
```diff
@@ -30,6 +30,6 @@
   "ts-node": {
     "esm": true
   },
-  "exclude": ["node_modules"],
-  "include": ["./**/*.ts", "package.json"],
+  "exclude": ["node_modules", "dist"],
```
ukorvl, line 33: outDir should be excluded by default by typescript

explorer_backend/tsconfig.json
```diff
@@ -30,6 +30,6 @@
...
+  "include": ["index.ts", "start.ts", "./**/*.ts", "./start.ts"],
```
ukorvl, line 34: Isn't it redundant? `"./**/*.ts"` should cover the `"./start.ts"`

explorer_backend/tsconfig.json
knazarov: I've copied it from niljs :)
But fine, I'll remove it.

knazarov: Wait, I can't remove it. Because I don't use outDir, I use outFile

ukorvl (approved):

PR #670, opened 2025-03-26
[rpc, stresser] Introduce dev-api with `DoPanicOnShard` method

shermike:
`DEVAPI` is an api for utilities useful for developers. Currently, it contains only one method `DoPanicOnShard` which forces validator, which is currently being used for the given shard, to crash.

Also, there are some improvements in Stresser:
- Add `do_panic` workload.
- Add retries for top-up via Faucet. It is necessary because of a known issue with data race of Faucet contract's seqno.
- Improve the procedure of waiting for cluster readiness.

### DIFF
--- a/nil/services/rpc/jsonrpc/dev_api.go
+++ b/nil/services/rpc/jsonrpc/dev_api.go
@@ -1,0 +1,26 @@
+package jsonrpc
+
+import (
+	"context"
+
+	"github.com/NilFoundation/nil/nil/internal/types"
+	"github.com/NilFoundation/nil/nil/services/rpc/rawapi"
+)
+
+type DevAPI interface {
+	DoPanicOnShard(ctx context.Context, shardId types.ShardId) (uint64, error)
+}
+
+type DevAPIImpl struct {
+	rawApi rawapi.NodeApi
+}
+
+func NewDevAPI(rawApi rawapi.NodeApi) DevAPI {
+	return &DevAPIImpl{
+		rawApi: rawApi,
+	}
+}
+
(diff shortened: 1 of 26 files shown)


nil/services/rpc/jsonrpc/dev_api.go
```diff
@@ -0,0 +1,26 @@
...
+	rawApi rawapi.NodeApi
+}
+
+func NewDevAPI(rawApi rawapi.NodeApi) DevAPI {
+	return &DevAPIImpl{
+		rawApi: rawApi,
+	}
+}
+
+func (d *DevAPIImpl) DoPanicOnShard(ctx context.Context, shardId types.ShardId) (string, error) {
```
olegrok, line 24: Do we really need to return string here? Why not just error?

shermike: Initially, I was thinking about returning some identifier of the validator which was crashed. It would probably be a libp2p id or something similar. But it seems it is not easily feasible, so maybe it indeed worth changing to bool.

nil/services/stresser/stresser.go
```diff
@@ -239,6 +239,7 @@ func (s *Stresser) Run(ctx context.Context) error {
 			ExportMetrics: true,
 		}
 		nilConfig.RPCPort = 8529
+		nilConfig.EnableDevApi = true
```
olegrok, line 242: That's a bit strange to enable it for the single instance mode.

Our purpose is to test how consensus works if some participant is killed. But here we can kill all shards and all participants via single call.

Issue #470, opened 2025-03-05
Introduce `txpool` RPC API for txpool inspecting


shermike:
Implement a new namespace in JSON-RPC with prefix `txpool_`. It should provide an API to inspect the content of the transaction pool, similar to Geth's API: https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-txpool
At the moment we can implement only the following methods:
- `txpool_status`
- `txpool_content`

PR #671, opened 2025-03-26
[txpool, api] Introduce txpool RPC API for txpool inspecting  

Zerg1996:
This PR implemented #470

### DIFF
--- a/nil/client/direct_client.go
+++ b/nil/client/direct_client.go
@@ -17,13 +17,14 @@ import (
 	"github.com/NilFoundation/nil/nil/services/rpc/transport"
 )
 
-// DirectClient is a client that interacts with the	 end api directly, without using the rpc server.
+// DirectClient is a client that interacts with the end api directly, without using the rpc server.
 type DirectClient struct {
-	ethApi   jsonrpc.EthAPI
-	debugApi jsonrpc.DebugAPI
-	dbApi    jsonrpc.DbAPI
-	web3Api  jsonrpc.Web3API
-	devApi   jsonrpc.DevAPI
+	ethApi    jsonrpc.EthAPI
+	debugApi  jsonrpc.DebugAPI
+	dbApi     jsonrpc.DbAPI
+	web3Api   jsonrpc.Web3API
+	devApi    jsonrpc.DevAPI
+	txPoolApi jsonrpc.TxPoolAPI
 }
 
 var _ Client = (*DirectClient)(nil)
@@ -39,13 +40,15 @@ func NewEthClient(
 	dbApi := jsonrpc.NewDbAPI(db, logger)
 	web3Api := jsonrpc.NewWeb3API(localApi)
 	devApi := jsonrpc.NewDevAPI(localApi)
+	txPoolApi := jsonrpc.NewTxPoolAPI(localApi, logger)
 
 	return &DirectClient{
-		ethApi:   ethApi,
-		debugApi: debugApi,
-		dbApi:    dbApi,
-		web3Api:  web3Api,
-		devApi:   devApi,
+		ethApi:    ethApi,
+		debugApi:  debugApi,
+		dbApi:     dbApi,
+		web3Api:   web3Api,
+		devApi:    devApi,
+		txPoolApi: txPoolApi,
 	}, nil
 }
 
@@ -396,6 +399,10 @@ func (c *DirectClient) DoPanicOnShard(ctx context.Context, shardId types.ShardId
 	return c.devApi.DoPanicOnShard(ctx, shardId)
 }
 
-func (c *DirectClient) GetTxpoolStatus(ctx context.Context, shardId types.ShardId) (uint64, error) { // Zerg
-	panic("Not supported")
+func (c *DirectClient) GetTxpoolStatus(ctx context.Context, shardId types.ShardId) (jsonrpc.TxPoolStatus, error) {
+	return c.txPoolApi.GetTxpoolStatus(ctx, shardId)
+}
+
+func (c *DirectClient) GetTxpoolContent(ctx context.Context, shardId types.ShardId) (jsonrpc.TxPoolContent, error) {
+	return c.txPoolApi.GetTxpoolContent(ctx, shardId)
 }
--- a/nil/client/direct_client.go
+++ b/nil/client/direct_client.go
@@ -17,7 +17,7 @@ import (
 	"github.com/NilFoundation/nil/nil/services/rpc/transport"
 )
 
-// DirectClient is a client that interacts with the end api directly, without using the rpc server.
+// DirectClient is a client that interacts with the	 end api directly, without using the rpc server.
 type DirectClient struct {
 	ethApi   jsonrpc.EthAPI
 	debugApi jsonrpc.DebugAPI
@@ -394,4 +394,8 @@ func (c *DirectClient) ClientVersion(ctx context.Context) (string, error) {
 
 func (c *DirectClient) DoPanicOnShard(ctx context.Context, shardId types.ShardId) (uint64, error) {
 	return c.devApi.DoPanicOnShard(ctx, shardId)
+}
+
+func (c *DirectClient) GetTxpoolStatus(ctx context.Context, shardId types.ShardId) (uint64, error) { // Zerg
+	panic("Not supported")
 }
--- a/nil/client/rpc/client.go
+++ b/nil/client/rpc/client.go
@@ -81,6 +81,8 @@ const (
 	Debug_getContract                    = "debug_getContract"
 	Web3_clientVersion                   = "web3_clientVersion"
 	Dev_doPanicOnShard                   = "dev_doPanicOnShard"
+	Txpool_getTxpoolStatus               = "txpool_GetTxpoolStatus"
+	Txpool_getTxpoolContent              = "txpool_GetTxpoolContent"
 )
 
 const (
@@ -968,4 +970,19 @@ func simpleCallUint64[ReturnType ~uint64](
 		return 0, err
 	}
 	return ReturnType(result), err
+}
+
+func (c *Client) GetTxpoolStatus(ctx context.Context, shardId types.ShardId) (uint64, error) {
+	request := c.newRequest(Txpool_getTxpoolStatus, shardId)
+	res, err := c.performRequest(ctx, request)
+	if err != nil {
+		return 0, err
+	}
+
+	var resValue uint64
+	if err := json.Unmarshal(res, &resValue); err != nil {
+		return 0, err
+	}
+
+	return resValue, err
 }
--- a/nil/client/rpc/client.go
+++ b/nil/client/rpc/client.go
@@ -81,8 +81,8 @@ const (
 	Debug_getContract                    = "debug_getContract"
 	Web3_clientVersion                   = "web3_clientVersion"
 	Dev_doPanicOnShard                   = "dev_doPanicOnShard"
-	Txpool_getTxpoolStatus               = "txpool_GetTxpoolStatus"
-	Txpool_getTxpoolContent              = "txpool_GetTxpoolContent"
+	Txpool_getTxpoolStatus               = "txpool_getTxpoolStatus"
+	Txpool_getTxpoolContent              = "txpool_getTxpoolContent"
 )
 
 const (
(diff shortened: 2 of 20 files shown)


nil/services/rpc/rawapi/server.go
```diff
@@ -37,6 +37,9 @@ type NetworkTransportProtocolRo interface {
 	GetNumShards() pb.Uint64Response
 
 	ClientVersion() pb.StringResponse
+
+	GetTxpoolStatus() pb.Uint64Response
+	GetTxpoolContent() pb.Uint64Response
```
Un1oR, line 42: I don't understand what is going on here
In `LocalShardApi` the return type is `[]*types.TxnWithHash`, and here it is `pb.Uint64Response`. It looks like an obvious inconsistency (besides I don't see converters added for `[]*types.TxnWithHas`).

nil/services/rpc/jsonrpc/txpool_api_test.go
```diff
@@ -0,0 +1,106 @@
...
+	}
+
+	shardApis := make(map[types.ShardId]rawapi.ShardApi)
+	localShardApi := rawapi.NewLocalShardApi(types.MainShardId, database, suite.pool)
+
+	shardApis[types.MainShardId], err = rawapi.NewLocalRawApiAccessor(types.MainShardId, localShardApi)
+
+	localApi := rawapi.NewNodeApiOverShardApis(localShardApis)
+	suite.txnpoolApi = NewTxPoolAPI(localApi, logging.NewLogger("Test"))
+	suite.api = rawapi.NewNodeApiOverShardApis(shardApis)
```
Un1oR, line 60: I don't understand why we need `localShardApis` and `shardApis` at the same time. Why don't we want to do just that?
```go
	mainShardApi,err := rawapi.NewLocalRawApiAccessor(
		types.MainShardId,
		rawapi.NewLocalShardApi(types.MainShardId, database, suite.pool))
	suite.Require().NoError(err)
	localShardApis := map[types.ShardId]rawapi.ShardApi{types.MainShardId: mainShardApi}

	suite.txnpoolApi = NewTxPoolAPI(localShardApis, logging.NewLogger("Test"))
	suite.api = rawapi.NewNodeApiOverShardApis(localShardApis)
```

nil/services/rpc/rawapi/local_txnpool.go
```diff
@@ -25,3 +25,11 @@ func (api *LocalShardApi) SendTransaction(ctx context.Context, encoded []byte) (
 	}
 	return reasons[0], nil
 }
+
+func (api *LocalShardApi) GetTxpoolStatus(ctx context.Context) (uint64, error) {
```
Un1oR, line 29: I don't really like the name of the method. Why not call it `GetTxpoolQueueueueLen` for example? Seeing the word "status" and some number, I rather expect it to be a value from some enumeration of possible statuses or something.

nil/services/rpc/jsonrpc/txpool_api_test.go
```diff
@@ -0,0 +1,106 @@
...
+	transactionAmount := uint64(10)
+
+	for i := range transactionAmount {
+		addr := types.ShardAndHexToAddress(0, "deadbeef"+fmt.Sprintf("%02d", i))
+		txn := newTransaction(addr, 0, 123, types.Code{byte(i)})
+		_, err := suite.pool.Add(ctx, txn)
+		suite.Require().NoError(err)
+	}
+
+	suite.Run("NodeApi", func() {
```
Un1oR, line 79: Am I correct in understanding that this is a separate test on RawAPI so that our checks are more granular than when checking only the JSON-RPC API?

nil/internal/types/transaction.go
```diff
@@ -588,20 +588,20 @@ func (m TransactionFlags) IsResponse() bool {
...
+		Transaction: *txn,
 		hash:        txn.Hash(),
 	}
 }
 
 func (m *TxnWithHash) Hash() common.Hash {
 	return m.hash
 }
+
+//go:generate go run github.com/NilFoundation/fastssz/sszgen --path transaction.go -include ../../common/hexutil/bytes.go,../../common/length.go,address.go,gas.go,value.go,code.go,shard.go,bloom.go,log.go,../../common/hash.go,signature.go,account.go,bitflags.go --objs TxnWithHash,Transaction,ExternalTransaction,InternalTransactionPayload,TransactionDigest,TransactionFlags,EvmState,AsyncContext,AsyncResponsePayload
```
olegrok, line 607: You don't need TxnWithHash... You need to transfer only transaction ssz. Hash is trivially calculated on the client.

nil/services/rpc/jsonrpc/txpool_api.go
```diff
@@ -0,0 +1,36 @@
...
+	"context"
+
+	"github.com/NilFoundation/nil/nil/common/logging"
+	"github.com/NilFoundation/nil/nil/internal/types"
+	"github.com/NilFoundation/nil/nil/services/rpc/rawapi"
+)
+
+type TxPoolAPI interface {
+	GetTxpoolStatus(ctx context.Context, shardId types.ShardId) (uint64, error)
+	GetTxpoolContent(ctx context.Context, shardId types.ShardId) ([]*types.TxnWithHash, error)
```
olegrok, line 13: Let's implement something similar to eth: https://geth.ethereum.org/docs/interacting-with-geth/rpc/ns-txpool

```
txpool.Content() (map[string]map[string]map[string]*RPCTransaction)
```

Right now we have only "pending" field. But it can be changed in future.

nil/services/rpc/jsonrpc/txpool_api_test.go
```diff
@@ -0,0 +1,106 @@
...
+			MaxFeePerGas:         types.NewValueFromUint64(defaultMaxFee),
+			Data:                 code,
+		},
+	}
+}
+
+func (suite *SuiteTxnPoolApi) SetupSuite() {
+	suite.SuiteAccountsBase.SetupSuite()
+	var err error
+	ctx := context.Background()
```
olegrok, line 40: s.T().Context()

nil/internal/types/transaction.go
```diff
@@ -588,20 +588,20 @@ func (m TransactionFlags) IsResponse() bool {
 	return m.GetBit(TransactionFlagResponse)
 }
 
-//go:generate go run github.com/NilFoundation/fastssz/sszgen --path transaction.go -include ../../common/hexutil/bytes.go,../../common/length.go,address.go,gas.go,value.go,code.go,shard.go,bloom.go,log.go,../../common/hash.go,signature.go,account.go,bitflags.go --objs Transaction,ExternalTransaction,InternalTransactionPayload,TransactionDigest,TransactionFlags,EvmState,AsyncContext,AsyncResponsePayload
-
 type TxnWithHash struct {
-	*Transaction
+	Transaction
```
shermike, line 592: Why did you change pointer to value?

Un1oR: `fastssz` doesn't chew otherwise. But I [suggested](https://github.com/NilFoundation/nil/pull/671#discussion_r2017522712) above not to make that change after all.

nil/internal/types/transaction.go
dmtrskv: Also, the client MUST calculate the hash itself, so it should not be returned by the api.

nil/services/rpc/jsonrpc/txpool_api_test.go
Zerg1996: Yes! Now we also have different output format for them.

nil/internal/types/transaction.go
Zerg1996: Reverted!

nil/internal/types/transaction.go
Zerg1996: Fixed!

Un1oR (approved):

nil/client/rpc/client.go
```diff
@@ -969,3 +971,11 @@ func simpleCallUint64[ReturnType ~uint64](
 	}
 	return ReturnType(result), err
 }
+
+func (c *Client) GetTxpoolStatus(ctx context.Context, shardId types.ShardId) (uint64, error) {
+	return simpleCallUint64[uint64](ctx, c, Txpool_getTxpoolStatus, shardId)
+}
+
+func (c *Client) GetTxpoolContent(ctx context.Context, shardId types.ShardId) ([]*types.Transaction, error) {
```
olegrok, line 979:
```suggestion
func (c *Client) GetTxpoolContent(ctx context.Context, shardId types.ShardId) (jsonrpc.TxPoolContent, error) {
```

olegrok: Seems this function is untested...

nil/client/rpc/client.go
```diff
@@ -969,3 +971,11 @@ func simpleCallUint64[ReturnType ~uint64](
 	}
 	return ReturnType(result), err
 }
+
+func (c *Client) GetTxpoolStatus(ctx context.Context, shardId types.ShardId) (uint64, error) {
```
olegrok, line 975:
```suggestion
func (c *Client) GetTxpoolStatus(ctx context.Context, shardId types.ShardId) (TxPoolStatus, error) {
```

nil/client/direct_client.go
```diff
@@ -395,3 +395,11 @@ func (c *DirectClient) ClientVersion(ctx context.Context) (string, error) {
...
 	return c.devApi.DoPanicOnShard(ctx, shardId)
 }
+
+func (c *DirectClient) GetTxpoolStatus(ctx context.Context, shardId types.ShardId) (uint64, error) {
+	panic("Not supported")
+}
+
+func (c *DirectClient) GetTxpoolContent(ctx context.Context, shardId types.ShardId) ([]*types.Transaction, error) {
+	panic("Not supported")
+}
```
olegrok, line 405: You can implement it just to check that this methods have correct types. Anyway implementation is direct call of jsonrpc.TxPoolApi

nil/services/rpc/rawapi/pb/conversion.go
```diff
@@ -1370,3 +1370,37 @@ func (r *SendTransactionRequest) PackProtoMessage(transactionSSZ []byte) error {
 func (r *SendTransactionRequest) UnpackProtoMessage() ([]byte, error) {
 	return r.TransactionSSZ, nil
 }
+
+func (txn *RawTxnsResponse) PackProtoMessage(txnWithHash []*types.Transaction, err error) error {
```
olegrok, line 1374:
```suggestion
func (txn *RawTxnsResponse) PackProtoMessage(txn []*types.Transaction, err error) error {
```

This type doesn't include hash

nil/client/rpc/client.go
Zerg1996: I add simple test just for call. Now I don't really understand how to stop pool to handle transaction. Maybe it's a good todo for next PR.

nil/services/rpc/jsonrpc/txpool_api.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/NilFoundation/nil/nil/services/rpc/rawapi"
+)
+
+type TxPoolStatus struct {
+	Pending uint64 `json:"pending"`
+	Queued  uint64 `json:"queued"`
+}
+
+type TxPoolContent struct {
+	Pending map[string]map[string]*types.Transaction `json:"pending"`
```
olegrok, line 17: Let's use jsonrpc.RPCTransaction type instead of types.Transaction. It will be consistent with other functions.

Erigon also uses RPCtransaction type in their implementation https://github.com/erigontech/erigon/blob/df85a9a43f899a69a3c3ce7b894364c0891dd662/turbo/jsonrpc/eth_api.go#L418

nil/tests/cli/cli_test.go
```diff
@@ -626,6 +626,14 @@ func (s *SuiteCliExec) TestCliCometa() {
 	})
 }
 
+func (s *SuiteCliExec) TestCallCliTxPool() {
```
olegrok, line 629: Move this test into rpc test suite. There is no connection with CLI here

olegrok: https://github.com/NilFoundation/nil/blob/main/nil/tests/rpc_service/rpc_service_test.go

nil/services/rpc/jsonrpc/types.go
```diff
@@ -53,13 +49,39 @@ type RPCInTransaction struct {
...
+// @componentprop Index index string true "The transaction index."
+// @componentprop Seqno seqno string true "The sequence number of the transaction."
+// @componentprop Signature signature string true "The transaction signature."
+// @componentprop Success success boolean true "The flag that shows whether the transaction was successful."
+// @componentprop Flags flags string true "The array of transaction flags."
+// @componentprop To to string true "The address where the transaction was sent."
+// @componentprop Value value string true "The transaction value."
+// @componentprop Token value array true "Token values."
+type RPCInTransaction struct {
+	Transaction *Transaction      `json:"transaction"`
```
olegrok, line 77: Just embed it

```suggestion
	Transaction
```

nil/services/rpc/jsonrpc/types.go
```diff
@@ -303,38 +325,44 @@ func (re *RPCReceipt) IsCommitted() bool {
...
+	transaction *types.Transaction, receipt *types.Receipt, index types.TransactionIndex,
+	blockHash common.Hash, blockId types.BlockNumber,
+) (*RPCInTransaction, error) {
+	hash := transaction.Hash()
+	if receipt == nil || hash != receipt.TxnHash {
+		return nil, errors.New("txn and receipt are not compatible")
+	}
+
+	result := &RPCInTransaction{
+		Transaction: NewTransaction(transaction),
```
olegrok, line 359: You can calculate hash only once
```suggestion
	txn := NewTransaction(transaction)
	if receipt == nil || txn.Hash != receipt.TxnHash {
		return nil, errors.New("txn and receipt are not compatible")
	}

	result := &RPCInTransaction{
		Transaction: txn,
```

olegrok (approved):

PR #677, opened 2025-03-27
[playground] add Playground version to playground logo

ukorvl:
This diff adds `Playground v123` text to playground logo if version is specified.
The version is injected as `env` during the build because it is not a subject to be changed after the app is built.

### DIFF
--- a/nix/nilexplorer.nix
+++ b/nix/nilexplorer.nix
@@ -24,6 +24,7 @@ stdenv.mkDerivation rec {
     "biome.json"
     "^explorer_frontend(/.*)?$"
     "^explorer_backend(/.*)?$"
+    "VERSION"
   ];
 
   pnpmDeps = (callPackage ./npmdeps.nix { });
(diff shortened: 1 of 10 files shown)


nix/nilexplorer.nix
```diff
@@ -54,7 +56,11 @@ stdenv.mkDerivation rec {
     (cd smart-contracts; pnpm run build)
     (cd niljs; pnpm run build)
 
-    (cd explorer_frontend; pnpm run build)
+    cd explorer_frontend
+    export VITE_PLAYGROUND_VERSION=${revCount}
```
knazarov, line 60: I've commented in the chat, but I'll put it here as well. If you do it like this, the explorer will be re-built and re-tested every time new comits are made. This is bad for the purposes of caching. Ideally, the derivation/package should only depend on the source files and not the revision or commit number.

What you can do instead is read the version information from a file, and this file can be written to the deb package at the time deb packages is created.

ukorvl: Added `VERSION` file in the root

KlonD90 (commented):
Do we still need env?

ukorvl:
> Do we still need env?
Yep, at least for this PR

KlonD90 (approved):

PR #681, opened 2025-03-27
relayer metrics

oclaw:
This PR adds metrics needed for monitoring state of the nil relayer service components

Added the following metrics:
- Common DB metrics (collection size, inserts, deletes)
- Event listener metrics (fetcher status, amount of incoming events, errors)
- Finality ensurer metrics (lag of updating finalized block value, amount of finalized & orphaned events)
- Transaction sender (relay/error count)

### DIFF
--- a/nil/services/relayer/internal/l1/storage.go
+++ b/nil/services/relayer/internal/l1/storage.go
@@ -28,13 +28,8 @@ const (
 	lastProcessedBlockKey   = "last_processed_block_key"
 )
 
-type EventStorageMetrics interface {
-	// TODO(oclaw)
-}
-
 type EventStorage struct {
 	*storage.BaseStorage
-	metrics         EventStorageMetrics
 	eventsSequencer db.Sequence
 }
 
@@ -42,12 +37,11 @@ func NewEventStorage(
 	ctx context.Context,
 	database db.DB,
 	clock clockwork.Clock,
-	metrics EventStorageMetrics,
+	metrics storage.TableMetrics,
 	logger logging.Logger,
 ) (*EventStorage, error) {
 	es := &EventStorage{
-		BaseStorage: storage.NewBaseStorage(ctx, database, clock, logger),
-		metrics:     metrics,
+		BaseStorage: storage.NewBaseStorage(ctx, database, clock, logger, metrics),
 	}
 	var err error
 	es.eventsSequencer, err = database.GetSequence(ctx, []byte(pendingEventsSequencer), 100)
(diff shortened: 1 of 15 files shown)


nil/services/relayer/internal/l1/storage.go
```diff
@@ -135,7 +133,9 @@ func (es *EventStorage) DeleteEvents(ctx context.Context, hashes []ethcommon.Has
 			}
 		}
 
-		return es.Commit(tx)
+		return es.Commit(tx, func() {
+			es.Metrics.RecordDeletes(ctx, pendingEventsTable, len(hashes))
```
akokoshn, line 137: in case of error from `tx.Delete` number of deleted events may be less than `len(hashes)`

nil/services/relayer/internal/l2/storage.go
```diff
@@ -121,6 +123,8 @@ func (es *EventStorage) DeleteEvents(ctx context.Context, hashes []ethcommon.Has
 			}
 		}
 
-		return es.Commit(tx)
+		return es.Commit(tx, func() {
+			es.Metrics.RecordDeletes(ctx, pendingEventsTable, len(hashes))
```
akokoshn, line 127: incorrect value in case of delete error

nil/services/relayer/internal/storage/json_writer.go
```diff
@@ -78,7 +84,9 @@ func (jdwr *jsonDbWriter[T]) PutManyTx(ctx context.Context, reqs []InsertRequest
 		}
 	}
 
-	return jdwr.storage.Commit(tx)
+	return jdwr.storage.Commit(tx, func() {
+		jdwr.storage.Metrics.RecordInserts(ctx, jdwr.table, len(reqs))
```
akokoshn, line 88: same problem with error exit

nil/services/relayer/service.go
```diff
@@ -29,6 +32,9 @@ func DefaultRelayerConfig() *RelayerConfig {
 		FinalityEnsurerConfig:   l1.DefaultFinalityEnsurerConfig(),
 		TransactionSenderConfig: l2.DefaultTransactionSenderConfig(),
 		L2ContractConfig:        l2.DefaultContractConfig(),
+		TelemetryConfig: &telemetry.Config{
```
akokoshn, line 35: missed space?

nil/services/relayer/internal/l1/storage.go
oclaw: Actually now metrics only in case if transaction successfully commited, so this code is OK. But maybe it makes sense to rework this approach, will research a bit

oclaw: I thought about it again and this approach seems the most reasonable for our needs. On Commit() error the whole transaction will be rolled back so no need to track any metric. And the whole storage operation error will be reachable via high level metrics so we won't miss it

nil/services/relayer/service.go
oclaw: it is a `make lint` formatting, not gonna argue with it :)

akokoshn (approved):

PR #688, opened 2025-03-27
[GH-#] Code deduplication in the client

Un1oR:
A small helper for the frequency pattern

### DIFF
--- a/nil/client/rpc/client.go
+++ b/nil/client/rpc/client.go
@@ -378,17 +378,10 @@ func (c *Client) GetCode(ctx context.Context, addr types.Address, blockId any) (
 	if err != nil {
 		return types.Code{}, err
 	}
-
-	raw, err := c.call(ctx, Eth_getCode, addr, blockNrOrHash)
+	codeHex, err := simpleCall[string](ctx, c, Eth_getCode, addr, blockNrOrHash)
 	if err != nil {
 		return types.Code{}, err
 	}
-
-	var codeHex string
-	if err := json.Unmarshal(raw, &codeHex); err != nil {
-		return types.Code{}, err
-	}
-
 	return hexutil.FromHex(codeHex), nil
 }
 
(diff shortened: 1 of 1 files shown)


nil/client/rpc/client.go
```diff
@@ -1043,21 +978,27 @@ func (c *Client) GetDebugContract(
...
-	return DebugRPCContract, err
+	return simpleCall[*jsonrpc.DebugRPCContract](ctx, c, Debug_getContract, contractAddr, blockRef)
 }
 
 func (c *Client) DoPanicOnShard(ctx context.Context, shardId types.ShardId) (uint64, error) {
 	_, err := c.call(ctx, Dev_doPanicOnShard, shardId)
 	return 0, err
 }
+
+func simpleCall[ReturnType any](ctx context.Context, c *Client, method string, params ...any) (ReturnType, error) {
```
olegrok, line 989: Won't it work?
```go
func simpleCall[ReturnType any](ctx context.Context, c *Client, method string, params ...any) (ReturnType, error) {
	var result ReturnType
	res, err := c.call(ctx, method, params...)
	if err != nil {
		return result, err
	}

	err = json.Unmarshal(res, &result)
	return result, err
}
```

Un1oR: It seems like there would be an extra indirectness there at the `&result` type. Isn't it? Plus `result` would be a non-initialized pointer. Or am I missing something?

olegrok: > extra indirectness

What do you mean here? Everything will be better than using reflect.

> Plus result would be a non-initialized pointer.

`json.Unmarshal` will initialize your pointer. Because you pass your pointer by pointer (not by value)

shermike: If possible, I'd also suggest avoid using reflection.

olegrok (approved):

PR #692, opened 2025-03-28
[GH-#] Checking the error of initialization of syncers

Un1oR:
This could lead to crashes due to not receiving the last block


nil/services/nilservice/service.go
```diff
@@ -257,10 +257,12 @@ type syncersResult struct {
 	funcs   []concurrent.FuncWithSource
 	syncers []*collate.Syncer
 	wgInit  sync.WaitGroup
```
olegrok, line 259: Please use errgroup

https://pkg.go.dev/golang.org/x/sync/errgroup

olegrok (changes requested):

nil/services/nilservice/service.go
Un1oR: I don't understand exactly how you propose to apply it here. Where do I call `errgroup.Go`?

olegrok (approved):
Oh, well. My bad. I got your point.

PR #693, opened 2025-03-28
[explorer] Remove unused file

KlonD90:
We don't need trick with start.ts that used for load everything after opentelemetry since we don't use it should be easier


ukorvl:
Remove start.ts from `tsconfig` plz

olegrok (approved):

PR #699, opened 2025-03-30
Sync Committee: ProofBatch Task Result Handling

zadykian:
### Sync Committee: ProofBatch Task Result Handling

Follow-up #697:

* Do not trigger state reset if batch associated with the task does not exist;
* Extended `TaskStateChangeHandlerTestSuite` with more tests;

### DIFF
--- a/nil/services/synccommittee/core/task_state_change_handler.go
+++ b/nil/services/synccommittee/core/task_state_change_handler.go
@@ -13,6 +13,7 @@ import (
 )
 
 type ProvedBatchSetter interface {
+	BatchExists(ctx context.Context, batchId types.BatchId) (bool, error)
 	SetBatchAsProved(ctx context.Context, batchId types.BatchId) error
 }
 
@@ -45,6 +46,15 @@ func (h *taskStateChangeHandler) OnTaskTerminated(
 	task *types.Task,
 	result *types.TaskResult,
 ) error {
+	batchExists, err := h.batchSetter.BatchExists(ctx, task.BatchId)
+	if err != nil {
+		return fmt.Errorf("failed to check if batch with id=%s exists: %w", task.BatchId, err)
+	}
+	if !batchExists {
+		h.onBatchDoesNotExist(task, err, "OnTaskTerminated")
+		return nil
+	}
+
 	switch {
 	case result.IsSuccess():
 		log.NewTaskResultEvent(h.logger, zerolog.InfoLevel, result).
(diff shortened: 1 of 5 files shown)


nil/services/synccommittee/core/task_state_change_handler.go
```diff
@@ -45,6 +46,18 @@ func (h *taskStateChangeHandler) OnTaskTerminated(
 	task *types.Task,
 	result *types.TaskResult,
 ) error {
+	batchExists, err := h.batchSetter.BatchExists(ctx, task.BatchId)
+	if err != nil {
+		return fmt.Errorf("failed to check if batch with id=%s exists: %w", task.BatchId, err)
+	}
+	if !batchExists {
```
akokoshn, line 53: seem check
```
case errors.Is(err, types.ErrBatchNotFound):
		log.NewTaskEvent(h.logger, zerolog.WarnLevel, task).
			Err(err).
			Stringer(logging.FieldBatchId, task.BatchId).
			Msg("batch not found, skipping (SetBatchAsProved)")
		return nil
```
in `onTaskSuccess` no need now

zadykian: Batch might still be removed after `BatchExists(...)` call (which would return true) and before `SetBatchAsProved(...)` is called

nil/services/synccommittee/core/task_state_change_handler.go
```diff
@@ -57,9 +67,26 @@ func (h *taskStateChangeHandler) OnTaskTerminated(
...
+		Msg("task execution failed with critical error, state will be reset")
+
+	err := h.stateResetLauncher.LaunchPartialResetWithSuspension(ctx, task.BatchId)
+
+	switch {
+	case err == nil:
+		return nil
+
+	case errors.Is(err, types.ErrBatchNotFound):
+		h.onBatchDoesNotExist(task, err, "LaunchPartialResetWithSuspension")
```
akokoshn, line 85: If last good batch is not exist during partial reset - have to reset completely

zadykian: If batch doesn't exist at this point, it can only mean that it has already been removed during previous reset.

Assume that we have 5 batches in the storage, where `[E]` is the latest created.
All these batches are being proved concurrently:

```
[A] <-- [B] <-- [C] <-- [D] <-- [E]
```

Then, proof generated for batch `[C]` fails with a critical error, Sync Committee resets itself, and `[B]` becomes the new latest batch:

```
[A] <-- [B]
```

Right after that proof generation for batch `[D]` also fails. 
It doesn't exist in the storage anymore, which is fine - we don't have to do anything, batches sequence is already valid

akokoshn: got it, thanks!

akokoshn (approved):

PR #705, opened 2025-03-31
📄 [docs] Fix broken LICENSE links in README files

donatik27:
Updated incorrect links to the LICENCE file across multiple README.md files to consistently point to the correct LICENSE file.

### DIFF
--- a/create-nil-hardhat-project/README.MD
+++ b/create-nil-hardhat-project/README.MD
@@ -14,7 +14,7 @@
 * [Usage](#-usage)
 * [Contributing](#-contributing)
 * [Work in progress](#-work-in-progress)
-* [License](#license)
+* [Licence](#licence)
 
 ## 🚀 Overview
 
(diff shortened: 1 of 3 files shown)


khannanov-nil:
Thank you for your contribution! For minor problems, please open a GitHub issue instead of opening a PR.

khannanov-nil:
> Hi, thanks for the clarification! Just to make sure I understand — opening an issue for these types of problems helps keep things more organized on your end?
> 
> Also, in this case, since it's a simple but repeated fix across multiple places, wouldn't it be more efficient to just merge the PR directly and avoid needing to reapply the same change manually later on?
> 
> Let me know if you'd prefer I migrate this into an issue, happy to help however works best for the team 👍

@donatik27 thanks for your interest, it is appreciated!

Yes, the main reason for us preferring issues for such problems is that the procedure for merging PRs from contributors into the production branch is a complicated process that involves commit squashing and commit signing. In many cases, performing these actions takes longer than applying the proposed fix. I look at docs-related open issues frequently so there's no risk of me missing something.

Issue #703, opened 2025-03-31
UX enhancements for interactive tutorials


khannanov-nil:
# Problem statement

We successfully rolled out interactive tutorials but they need some UX enhancements to be more popular among developers.

# Solution

Implement the following changes requested by @Gezort 

- Add clear indicators where we expect developers to write their own code. Simply adding comments with a `TODO` remark to contracts should be sufficient 
- Either add the hash of the failed transaction to check failures or try to print the outcome of `nil debug TXN_HASH`
- Implement a mechanism for caching solutions for completed tutorials 
- Improve the clarity of some error messages

# Additional considerations

@Gezort also requested the following changes which may require some discussion

- Return the user to the tutorials menu on tutorial completion. I disagree: this will make it difficult to showcase that checks have been run successfully. The user may also want to modify their solution instead of trying another tutorial.
- Highlightign completed tutorials instead of reordering the menu. My opinion: the current system prevents visual clutter by adding a clear separator, and the user is much less likely to misclick on a tutorial they've already completed. 
- Changing the indent size to four spaces. I prefer retaining it at two spaces since the Playground uses the same indentation.

khannanov-nil:
a note about `nil debug`: to correctly debug a contract, we'd need to access the =nil; CLI inside `explorer_frontend`. `Nil.js` does not fully support the debug API exposed by Cometa. so at the moment providing the transaction hash with some clarifications about Cometa debugging seems like a much more feasible option.

PR #709, opened 2025-03-31
[tutorial] added misc UX improvements

khannanov-nil:
## Short Summary

This diff adds several UX improvements to interactive tutorials.

## What Changes Were Made

- Added comments suggesting where exactly developers should add their code to tutorials.
- When a tutorial check fails and there is a fail receipt, the logs display the transaction hash for debugging.
- Completed solutions are cached and displayed when a developer accesses a completed tutorial.
- The text for most errors has been made clearer. 

## Related Issue

Addresses #703 .

## Checklist

### DIFF
--- a/explorer_frontend/src/features/tutorial/init.ts
+++ b/explorer_frontend/src/features/tutorial/init.ts
@@ -1,11 +1,13 @@
-import { sample } from "effector";
+import { combine, sample } from "effector";
 import { persist } from "effector-storage/local";
 import { changeCode, loadedTutorialPage } from "../code/model";
+import { $contracts } from "../contracts/models/base";
 import { notFoundRoute } from "../routing/routes/routes";
 import { tutorialWithUrlStringRoute } from "../routing/routes/tutorialRoute";
 import {
   $completedTutorials,
   $tutorial,
+  $tutorialUserSolutions,
   $tutorials,
   fetchAllTutorialsFx,
   fetchTutorialEvent,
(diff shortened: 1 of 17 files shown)


explorer_frontend/src/features/tutorial/model.ts
```diff
@@ -39,6 +39,8 @@ export const $tutorials = tutorialDomain.createStore<Tutorial[]>([]);
 
 export const $completedTutorials = tutorialDomain.createStore<number[]>([]);
 
+export const $tutorialUserSolutions = tutorialDomain.createStore<Record<string, string>[]>([]);
```
KlonD90, line 42: no idea why do you need here array. Because your actually lookup is tutorial.slug as a key. So you could use this as key and make Record<slug, Object definition> instead of it and have proper object definition. Now you just name it Record<string, string> without any contract beside be Records of string

explorer_frontend/src/features/tutorial/init.ts
```diff
@@ -63,3 +90,24 @@ sample({
...
+
+    const currentSolutions = $tutorialUserSolutions.getState();
+
+    const filteredSolutions = currentSolutions.filter(
+      (solution) => solution.url !== urlSlug.urlSlug,
+    );
+
+    return [...filteredSolutions, res];
+  },
+  target: $tutorialUserSolutions,
```
KlonD90, line 112: if you want some state just use as source. Do not use getState. Also It's better to be written ```
$tutorialUserSolutions.on(
sample({clock: , source: }),
(newSolution, prevSolutions) => [newSolution, ...prevSolutions]
)
```

explorer_frontend/src/features/tutorial/model.ts
khannanov-nil: got it, removed the array

explorer_frontend/src/features/tutorial/init.ts
khannanov-nil: with the change to the store, it is no longer array

PR #711, opened 2025-03-31
[rpc] Do not return error if effective priority fee calculation failed

shermike:
It is a valid situation, in which case the status should be equal to `BaseFeeTooHigh`.

### DIFF
--- a/nil/services/rpc/rawapi/local_receipt.go
+++ b/nil/services/rpc/rawapi/local_receipt.go
@@ -7,6 +7,7 @@ import (
 
 	fastssz "github.com/NilFoundation/fastssz"
 	"github.com/NilFoundation/nil/nil/common"
+	"github.com/NilFoundation/nil/nil/common/logging"
 	"github.com/NilFoundation/nil/nil/internal/db"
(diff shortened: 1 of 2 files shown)


nil/services/rpc/rawapi/local_receipt.go
```diff
@@ -50,11 +50,11 @@ func (api *LocalShardApi) GetInTransactionReceipt(
 			return nil, err
 		}
 
-		priorityFee, ok := execution.GetEffectivePriorityFee(block.BaseFee, transaction)
-		if !ok {
-			return nil, errors.New("critical error: effective priority fee is invalid")
+		if priorityFee, ok := execution.GetEffectivePriorityFee(block.BaseFee, transaction); ok {
+			gasPrice = block.BaseFee.Add(priorityFee)
+		} else if receipt.Status != types.ErrorBaseFeeTooHigh {
+			api.logger.Error().Msgf("Calculation of EffectivePriorityFee failed with wrong status: %s", receipt.Status)
```
olegrok, line 56: Let's attach transaction hash to this message

PR #712, opened 2025-03-31
[playground] remove rpcUrl validation

ukorvl:
This diff contains two changes:
- Validation of the rpc url is removed. The reason behind that is that validation was too strict and user was not able to develop playground locally. 
- Rpc url is now set only after successful wallet creation. It's kinda obvious - there is no need in rpc url if you don't have a wallet. There was a buggy scenario when user tries to create a wallet with some rpc url, which passed validation, but request is not successful. In that case rpc url is going to be set without wallet creation and user is going to receive message "Something wrong with the network" every time the page is open.

### DIFF
--- a/explorer_frontend/biome.json
+++ b/explorer_frontend/biome.json
@@ -16,7 +16,7 @@
         "useKeyWithClickEvents": "off"
       },
       "correctness": {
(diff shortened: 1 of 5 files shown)


KlonD90 (approved):
lgtm

PR #713, opened 2025-03-31
relayer debug: add smart account creation

oclaw:
## Short Summary

This PR adds support for debug relayer CLI options

## What Changes Were Made

- Now with l2-debug CLI options salt and faucet service endpoint can be passed to service. It will try to automatically create smart account in nild in that case

## Checklist

### DIFF
--- a/nil/services/relayer/internal/l2/init.go
+++ b/nil/services/relayer/internal/l2/init.go
@@ -1,0 +1,129 @@
+package l2
+
+import (
+	"context"
+	"crypto/ecdsa"
+	"errors"
+	"fmt"
+	"os"
+
+	"github.com/NilFoundation/nil/nil/client"
+	"github.com/NilFoundation/nil/nil/client/rpc"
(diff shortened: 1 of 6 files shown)


nil/services/relayer/internal/l2/init.go
```diff
@@ -0,0 +1,130 @@
...
+	addr, err := debugDeployer.CreateSmartAccount(
+		types.BaseShardId,
+		salt,
+		amount,
+		fee,
+		&key.PublicKey,
+	)
+	if errors.Is(err, cliservice.ErrSmartAccountExists) {
+		logger.Info().Str("smart_account_address", addr.Hex()).Msg("smart account already exists")
+		return addr, nil
```
akokoshn, line 115: check if `config.SmartAccountAddress == ""` or `addr`

oclaw: This check was done, but in invalid place. Move it here and improved code a bit

akokoshn (approved):

PR #714, opened 2025-03-31
[docs] fix node_modules for docsaibackend.nix

ukorvl:
This diff fixes `node_modules` contents that are placed into the deb package.
Since we use `pnpm` as a node package manager, we have non-standard `node_modules` structure. It contains mostly symlinks to `pnpm` global store.

If i call `ls -la node_modules/next` i get `.pnpm/next@15.2.3_@opentelemetry+api@1.9.0_@playwright+test@1.51.1_react-dom@18.3.1_react@18.3.1__react@18.3.1/node_modules/next`. Obviously, it can not be used on the target deployment machine, because the `pnpm` store is not reproduced there.

### DIFF
--- a/nix/docsaibackend.nix
+++ b/nix/docsaibackend.nix
@@ -1,4 +1,4 @@
-{ lib, stdenv, callPackage, pnpm_10, nodejs }:
+{ lib, stdenv, callPackage, pnpm_10, nodejs, rsync }:
 
 stdenv.mkDerivation rec {
   name = "aibackend";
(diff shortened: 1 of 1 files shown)


Un1oR:
`node-linker=hoisted` doesn't solve the problem of getting rid of symlinks?

ukorvl:
> `node-linker=hoisted` doesn't solve the problem of getting rid of symlinks?

It is, but it does not work as i want. Here we still need to install dependencies in a regular way (because it is much more optimized, without pnpm global store it is more or less like npm in terms of installation speed i suppose)
The intention is to remove symlinks only in the copied node_modules.

KlonD90 (approved):

ukorvl:
This change is breaking nix. Closing.

PR #679, opened 2025-03-27
[GH-659] Basic protocol version change handling

Un1oR:
Basic protocol update check. The check is performed only at startup, but not at runtime — this will be finalized later.

PR #717, opened 2025-04-01
[GH-#] Goroutine labels

Un1oR:
For tasks started via `RunWithTimeout`, I've added labels that are mapped to the corresponding goroutines in the IDE (at least in GoLand).

BEFORE

BEFORE

AFTER

AFTER

It works only when the `debugger` build tag is raised, so it doesn't add any overhead in production.

Read more in the JetBrains [blog article](https://blog.jetbrains.com/go/2020/03/03/how-to-find-goroutines-during-debugging/#using-a-custom-library-to-enable-debugging-labeling).

In particular this technique allowed me to find an [uncancelable waiting](https://github.com/NilFoundation/nil/pull/679/files#diff-ab230848b41511019d4484b9165904556e6b86d41eef2bf72c8279933b73b1b8L79) that prevented an archive node from terminating when an error occurred.


nil/services/nilservice/service.go
```diff
@@ -708,22 +721,24 @@ func createShards(
...
-					logger.Error().
-						Err(err).
-						Stringer(logging.FieldShardId, shardId).
-						Msg("Collator goroutine failed")
-					return err
-				}
-				return nil
-			}))
+			funcs = append(funcs, concurrent.MakeTask(
+				fmt.Sprintf("run collator %d", i),
```
olegrok, line 725: I guess we can use shardId in a following way: `[id] collator`

nil/common/concurrent/utils.go
```diff
@@ -71,29 +80,40 @@ func RunWithTimeout(ctx context.Context, timeout time.Duration, fs ...FuncWithSo
...
-			if err := fn.Func(ctx); err != nil {
+			if err := task.Func(ctx); err != nil {
 				once.Do(func() {
 					originError = &ExecutionError{
 						Err:   err,
-						Stack: fn.Stack,
+						Stack: task.Stack,
 					}
-					cancel()
+					cncl()
```
olegrok, line 105: build fails here

Un1oR: Thanks, I added to the commit line by line

nil/cmd/nil_block_generator/internal/commands/client.go
```diff
@@ -39,9 +39,11 @@ func backgroundNilNode(cfg *nildconfig.Config) {
 		cfg.Config,
 		database,
 		nil,
-		concurrent.WithSource(func(ctx context.Context) error {
-			return database.LogGC(ctx, cfg.DB.DiscardRatio, cfg.DB.GcFrequency)
-		}))
+		concurrent.MakeTask(
+			"badger log garbage collection",
```
olegrok, line 43: Just "badger GC"

nil/services/nilservice/service.go
```diff
@@ -497,26 +502,30 @@ func CreateNode(
...
-				logger.Error().
-					Err(err).
-					Stringer(logging.FieldShardId, cfg.Replay.ShardId).
-					Msg("Replayer goroutine failed")
-				return err
-			}
-			return nil
-		}))
+		funcs = append(funcs, concurrent.MakeTask(
+			"run replayer",
```
olegrok, line 506:
```suggestion
			"block-replay",
```

nil/services/nilservice/service.go
```diff
@@ -525,42 +534,23 @@ func CreateNode(
...
 
-	funcs = append(funcs, concurrent.WithSource(func(ctx context.Context) error {
-		if err := startAdminServer(ctx, cfg); err != nil {
-			logger.Error().Err(err).Msg("Admin server goroutine failed")
-			return err
-		}
-		return nil
-	}))
+	funcs = append(funcs, concurrent.MakeTask(
+		"start admin server",
```
olegrok, line 538:
```suggestion
		"admin-api",
```

nil/services/nilservice/service.go
```diff
@@ -583,6 +573,43 @@ func CreateNode(
...
+	syncersResult *syncersResult,
+	database db.DB,
+	logger logging.Logger,
+) []concurrent.Task {
+	if (cfg.RPCPort == 0 && cfg.HttpUrl == "") || rawApi == nil {
+		return tasks
+	}
+
+	return append(tasks, concurrent.MakeTask(
+		"start rpc server",
```
olegrok, line 589:
```suggestion
		"rpc-api",
```

PR #727, opened 2025-04-01
Bundle docs_ai_backend

knazarov:
## Short Summary

It looks like it's not very difficult to bundle all necessary npm dependencies together with the docs_ai_backend. What is needed is just to specify `output: standalone` in the next.js config.

I've tried to start the server with `node` from the resulting build and it worked. However, I can't do any in-depth testing and will wait for Shukhrat to do that part.

## Related Issue

Part of #728

## Breaking Changes (if any)

It is possible that I didn't ship everything that is needed in the resulting distribution, so help is needed.

## Checklist

### DIFF
--- a/docs_ai_backend/next.config.js
+++ b/docs_ai_backend/next.config.js
@@ -1,0 +1,3 @@
+module.exports = {
+  output: 'standalone',
+};
--- a/docs_ai_backend/package.json
+++ b/docs_ai_backend/package.json
@@ -5,7 +5,7 @@
   "scripts": {
     "dev": "next dev -p 7000",
     "build": "next build",
-    "start": "next start -p 8092"
+    "start": "PORT=8092 HOSTNAME=localhost node .next/standalone/docs_ai_backend/server.js"
   },
   "keywords": [],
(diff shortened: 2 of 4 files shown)


olegrok (approved):

ukorvl:
Let's change "start" script to "node ./path to server.js"?

khannanov-nil (changes requested):
There are two main issues that need to be investigated here. Both of them arise if the suggestion by @ukorvl is applied (i.e., the app is started via `next ./node .next/standalone/docs_ai_backend/server.js`.

1. The app starts on port `3000` by default but the production machine will expect it at port `8092`. Probably the port can be specified when the app is launched directly in the command.

2. Requests to `/api/services/db` do not work and they are needed to renew the DB. Specific error is:

```
[TypeError: fetch failed] {
  [cause]: [Error: getaddrinfo EAI_AGAIN undefined] {
    errno: -3001,
    code: 'EAI_AGAIN',
    syscall: 'getaddrinfo',
    hostname: 'undefined'
  }
}
```

If the app is launched via `npm run start`, the request works normally.

PR #734, opened 2025-04-02
[journald-forwarder] Implement store batch of events in one click request

Zerg1996:
## Short Summary

In this PR, the entire batch of logs is stored in ClickHouse in a single query execution. This helps to avoid overloading ClickHouse with too many small queries.

## What Changes Were Made

- journald-forwarder services now send batch of events into clickhouse
- updated journald-forwarder tests

## Checklist

### DIFF
--- a/nil/services/journald_forwarder/clickhouse.go
+++ b/nil/services/journald_forwarder/clickhouse.go
@@ -1,0 +1,162 @@
+package journald_forwarder
+
+import (
+	"context"
+	"fmt"
+	"strings"
+
+	"github.com/ClickHouse/clickhouse-go/v2"
+	"github.com/ClickHouse/clickhouse-go/v2/lib/driver"
+	"github.com/NilFoundation/nil/nil/common/logging"
+)
+
+const (
+	DefaultDatabase = "nil"
+	DefaultTable    = "events"
+)
+
+type ClickhouseConfig struct {
+	ClickhouseAddr string
+	ListenAddr     string
+	DbUser         string
+	DbDatabase     string
+	DbPassword     string
+}
+
+type Clickhouse struct {
+	ClickhouseConfig
+	connection clickhouse.Conn
+}
+
+func (s *Clickhouse) InsertColumnsInTable(
+	ctx context.Context,
+	database, tableName string,
+	columns, columnsType []string,
+) error {
+	columnClauses := make([]string, len(columns))
+	for i, column := range columns {
+		columnClauses[i] = fmt.Sprintf("ADD COLUMN %s %s", column, columnsType[i])
+	}
+
+	query := fmt.Sprintf(
+		"ALTER TABLE %s.%s %s;",
+		database,
+		tableName,
+		strings.Join(columnClauses, ", "),
+	)
+
+	return s.connection.Exec(ctx, query)
+}
+
+func (s *Clickhouse) InsertData(
+	ctx context.Context,
+	database string,
+	tableName string,
+	columns []string,
+	values [][]any,
+) error {
+	placeholders := make([]string, len(columns))
+	for i := range columns {
+		placeholders[i] = "?"
+	}
+
+	// Create multiple value placeholders for batch insert
+	valueRows := make([]string, len(values))
+	for i := range values {
+		valueRows[i] = fmt.Sprintf("(%s)", strings.Join(placeholders, ", "))
+	}
+
+	query := fmt.Sprintf(
+		"INSERT INTO %s.%s (%s) VALUES %s",
+		database,
+		tableName,
+		strings.Join(columns, ", "),
+		strings.Join(valueRows, ", "),
+	)
+
+	// Flatten values slice
+	flattenedValues := make([]any, 0, len(values)*len(columns))
+	for _, row := range values {
+		flattenedValues = append(flattenedValues, row...)
+	}
+
+	return s.connection.Exec(ctx, query, flattenedValues...)
+}
+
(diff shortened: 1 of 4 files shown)


nil/tests/journald_forwarder/journald_forwarder_test.go
```diff
@@ -73,38 +73,56 @@ func (s *SuiteJournaldForwarder) SetupSuite() {
...
+	}()
+
+	select {
+	case <-done:
+	case <-time.After(5 * time.Second):
+		s.T().Log("TearDownSuite timeout! Possible deadlock.")
+	}
+
 	if s.clickhouse != nil {
 		err := s.clickhouse.Process.Kill()
```
olegrok, line 96: Please consider

https://github.com/NilFoundation/nil/blob/2897ce0e8dfb503caddbf9035f1c28d3b6d51849/nil/cmd/exporter/internal/clickhouse/clickhouse_test.go#L46-L49

nil/tests/journald_forwarder/journald_forwarder_test.go
```diff
@@ -73,38 +73,56 @@ func (s *SuiteJournaldForwarder) SetupSuite() {
...
 		s.Require().NoError(err)
 	}
+
+	if s.connect != nil {
+		_ = s.connect.Close()
```
olegrok, line 101: Let's check error here

nil/services/journald_forwarder/clickhouse.go
```diff
@@ -0,0 +1,161 @@
...
+	defer rows.Close()
+
+	columns := make(map[string]any)
+	for rows.Next() {
+		var columnName string
+		if err := rows.Scan(&columnName); err != nil {
+			return nil, err
+		}
+		columns[columnName] = nil
+	}
```
olegrok, line 101: Error handling is missed here (and maybe in other places)
```suggestion
	}
	
	if err := rows.Err(); err != nil {
		return nil, err
	}
```

Zerg1996: Looks like we don't need such check, because in our case if Scan return error - we don't Scan next rows.

nil/services/journald_forwarder/clickhouse.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+type Clickhouse struct {
+	ClickhouseConfig
+	connect clickhouse.Conn
+}
+
+func (s *Clickhouse) InsertColumnsInTable(ctx context.Context,
+	database, tableName string,
+	columns, columnsType []string,
```
Un1oR, line 33: Let's avoid this argument formatting, where some of it is on different lines and some of it is on the same line?

nil/services/journald_forwarder/clickhouse.go
```diff
@@ -0,0 +1,161 @@
...
+	for i := range values {
+		valueRows[i] = fmt.Sprintf("(%s)", strings.Join(placeholders, ", "))
+	}
+
+	query := fmt.Sprintf(
+		"INSERT INTO %s.%s (%s) VALUES %s",
+		database,
+		tableName,
+		strings.Join(columns, ", "),
+		strings.Join(valueRows, ", "),
```
Un1oR, line 73: Is there not some achievable limit on the length of the query that we might encounter?

nil/services/journald_forwarder/clickhouse.go
```diff
@@ -0,0 +1,161 @@
...
+func (s *Clickhouse) GetTabelColumnNames(ctx context.Context, database, tableName string) (map[string]any, error) {
+	const columnQuery = "SELECT name FROM system.columns WHERE database = ? AND table = ?"
+
+	rows, err := s.connect.Query(ctx, columnQuery, database, tableName)
+	if err != nil {
+		return nil, err
+	}
+	defer rows.Close()
+
+	columns := make(map[string]any)
```
Un1oR, line 94: Is a dictionary really necessary here? I understand that we should not have duplicates. A sorted array (for which the diff operation should also be performed in linear time) would suffice.

nil/services/journald_forwarder/clickhouse.go
```diff
@@ -0,0 +1,161 @@
...
+	ClickhouseAddr string
+	ListenAddr     string
+	DbUser         string
+	DbDatabase     string
+	DbPassword     string
+}
+
+type Clickhouse struct {
+	ClickhouseConfig
+	connect clickhouse.Conn
```
Un1oR, line 28: Why not `connection'? It's strange to see a verb in a name.

nil/services/journald_forwarder/clickhouse.go
Un1oR: Okay, I guess we also have unordered columns in the events, in which case it's really easier to do the check in the hash table than to add sorting there as well.

nil/services/journald_forwarder/clickhouse.go
Zerg1996: Only on query size as I know.

nil/services/journald_forwarder/clickhouse.go
Zerg1996: Okay! Move ctx to newline.
But arguments with same type is looks pretty good on the same line.

nil/services/journald_forwarder/clickhouse.go
Un1oR: Should we provide for handling such a situation (and reducing the size of the batch) or do we consider such a situation impossible in our installation?

PR #736, opened 2025-04-02
[execution] don't inc ExtSeqno in case if tx is not included to block

olegrok:
The previous patch
(cbed7489a815956487b913a4d6ea96922e3ff28d) fixed the possibility of adding identical transactions to
multiple blocks. However, it introduced a regression.

Transactions that are not included in a block (and create so-called temporary receipts) could still
modify the state. This breaks replay, since the
account cannot be brought to the correct state
without that transaction.

This patch fixes the issue by rolling back ExtSeqno if the transaction is not included in a block.

This mainly affects deploy transactions. For
execution transactions, gas is paid if
`verifyExternal` succeeds, so state changes are valid in that case.

### DIFF
--- a/nil/internal/execution/block_generator.go
+++ b/nil/internal/execution/block_generator.go
@@ -273,6 +273,13 @@ func (g *BlockGenerator) handleTxn(txn *types.Transaction) error {
 		g.counters.ExecTransactions++
 	}
 
+	var err error
+	var seqno types.Seqno
+	if assert.Enable && txn.IsExternal() {
+		seqno, err = g.executionState.GetExtSeqno(txn.To)
+		check.PanicIfErr(err)
+	}
+
 	txnHash := g.executionState.AddInTransaction(txn)
 
 	var res *ExecutionResult
@@ -291,9 +298,21 @@ func (g *BlockGenerator) handleTxn(txn *types.Transaction) error {
 	if res.FatalError != nil {
 		return res.FatalError
 	}
-	g.addReceipt(res)
+	g.handleResult(res)
 	g.counters.CoinsUsed = g.counters.CoinsUsed.Add(res.CoinsUsed())
 
+	if assert.Enable && txn.IsExternal() {
+		newSeqno, err := g.executionState.GetExtSeqno(txn.To)
+		check.PanicIfErr(err)
+		if res.GasUsed == 0 {
+			check.PanicIfNotf(newSeqno == seqno,
+				"seqno changed during execution with GasUsed=0 (old %d, new: %d)", seqno, newSeqno)
+		} else {
+			check.PanicIfNotf(newSeqno == seqno+1,
+				"seqno was not changed correctly during execution (old %d, new: %d). Gas used: %d",
+				seqno, newSeqno, res.GasUsed)
+		}
+	}
 	return nil
 }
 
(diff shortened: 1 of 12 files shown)


nil/internal/execution/block_generator.go
```diff
@@ -294,6 +301,16 @@ func (g *BlockGenerator) handleTxn(txn *types.Transaction) error {
...
 	g.counters.CoinsUsed = g.counters.CoinsUsed.Add(res.CoinsUsed())
 
+	if assert.Enable && txn.IsExternal() {
+		newSeqno, err := g.executionState.GetExtSeqno(txn.To)
+		check.PanicIfErr(err)
+		if res.GasUsed == 0 {
+			check.PanicIfNotf(newSeqno == seqno, "seqno changed during execution")
+		} else {
+			check.PanicIfNotf(newSeqno == seqno+1,
+				"seqno was not changed during execution (old %d, new: %d). Gas used: %d", seqno, newSeqno, res.GasUsed)
```
Un1oR, line 311: Seems like a bad message
If `newSeqno` is equal to `seqno+2`, then the message "seqno was not changed during execution" will be wrong. And if `newSeqno == seqno`, then the `(old %d, new: %d)` part will have two identical values
I would reword)

nil/internal/execution/block_generator.go
```diff
@@ -370,6 +387,11 @@ func (g *BlockGenerator) addReceipt(execResult *ExecutionResult) {
 		// todo: fail generation here when collator performs full validation.
 		check.PanicIfNot(execResult.Failed())
 
+		// Transaction that doesn't create side-effects (receipts) shouldn't affect account state.
```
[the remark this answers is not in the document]
olegrok: Let's rename to `handleResult`

nil/internal/execution/block_generator.go
olegrok: Ok. Changed a bit. Now it's "...seqno was not changed **correctly**..."

nil/tests/regression/regression_test.go
```diff
@@ -243,7 +327,7 @@ func (s *SuiteRegression) TestNonRevertedErrDecoding() {
 	code, err := contracts.GetCode(contracts.NameTest)
 	s.Require().NoError(err)
 
-	payload := types.BuildDeployPayload(code, common.EmptyHash)
+	payload := types.BuildDeployPayload(code, common.Hash{0x03})
```
Un1oR, line 330: Why is something different here?

olegrok: Just to reduce chance of collision between tests.

nil/internal/execution/block_generator.go
```diff
@@ -370,6 +388,11 @@ func (g *BlockGenerator) addReceipt(execResult *ExecutionResult) {
...
+		seqno, err := g.executionState.GetExtSeqno(txn.To)
+		check.PanicIfErr(err)
+		check.PanicIfErr(g.executionState.SetExtSeqno(txn.To, seqno-1))
```
shermike, line 394: I'm not sure that it is the best place for reverting seqno. We increase it in the ExecutionState, it is quite far from here. What if `handleExternalTransaction` return error before calling `HandleTransaction`, thus, seqno won't be increased.
Maybe it is better to add this revert to the `ExecutionState::HandleTransaction` as defer function next to place where we increase seqno?

shermike, line 394: I'm not sure that it is the best place for reverting seqno. We increase it in the ExecutionState, it is quite far from here. What if `handleExternalTransaction` return error before calling `HandleTransaction`, thus, seqno won't be increased.
Maybe it is better to add this revert to the `ExecutionState::HandleTransaction` as defer function next to place where we increase seqno?

shermike (changes requested):

nil/internal/execution/block_generator.go
olegrok: Agree with you. Thanks for suggestion it's really better.

shermike (approved):

PR #738, opened 2025-04-03
[execution] Fix big transactions processing

shermike:
Previously, big transactions just stuck in the pool/blocks, and the node tried to execute them again and again. Thus, a kind of infinite loop was observed.
This patch fixes it by adding constraint to the EVM that transaction cannot consume more than block gas limit. Also, the check for exceeding the gas limit has now been moved after adding a transaction, otherwise such a transaction would never appear in the block.
This solution has a drawback that the block can exceed its gas limit by two times in the worst case. In the future, we'll probably want to solve it more appropriately by reducing the possible excess of the block gas limit.

Relates to #79

### DIFF
--- a/nil/tests/regression/regression_test.go
+++ b/nil/tests/regression/regression_test.go
@@ -3,6 +3,7 @@ package main
 import (
 	"math/big"
 	"testing"
+	"time"
 
 	"github.com/NilFoundation/nil/nil/common"
 	"github.com/NilFoundation/nil/nil/internal/contracts"
@@ -142,7 +143,7 @@ func (s *SuiteRegression) TestProposerOutOfGas() {
 	s.Require().True(receipt.Success)
 	s.Require().Equal("Success", receipt.Status)
 	s.Require().Len(receipt.OutReceipts, 1)
-	s.Require().Equal("OutOfGasDynamic", receipt.OutReceipts[0].Status)
+	s.Require().Equal("TransactionExceedsBlockGasLimit", receipt.OutReceipts[0].Status)
 }
 
 func (s *SuiteRegression) TestInsufficientFundsIncExtSeqno() {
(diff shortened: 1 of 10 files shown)


olegrok:
Is it related to this patch https://github.com/NilFoundation/nil/pull/447 ?

nil/tests/regression/regression_test.go
```diff
@@ -365,6 +365,43 @@ func (s *SuiteRegression) TestNonRevertedErrDecoding() {
...
+			types.NewFeePackFromGas(50_000_000),
+			types.NewZeroValue(),
+			nil)
+		s.Require().True(receipt.Success)
+		s.Require().Len(receipt.OutReceipts, 1)
+		s.Require().False(receipt.OutReceipts[0].Success)
+		s.Require().Equal("TransactionExceedsBlockGasLimit", receipt.OutReceipts[0].Status)
+	})
+
+	s.Run("External big transaction", func() {
```
olegrok, line 396: I guess you can use new txpool API to be sure that it doesn't stuck in the pool

shermike:
> Is it related to this patch #447 ?

Yes, it is. But it fixes also for external transactions. And add restriction to the EVM.

nil/tests/regression/regression_test.go
shermike: Good idea. Added, thanks.

nil/internal/vm/interpreter.go
```diff
@@ -268,6 +275,11 @@ func (in *EVMInterpreter) Run(contract *Contract, input []byte, readOnly bool) (
 		if err != nil {
 			break
 		}
+		if contract.Gas < blockLimitThreshold {
```
olegrok, line 278: To spend more gas than block limit you need to specify big FeeCredit and buy amount of gas that exceeds the limit. Can't we check it inside buyGas function to avoid calling EVM?To spend more gas than block limit you need to specify big FeeCredit and buy amount of gas that exceeds the limit. Can't we check it inside buyGas function to avoid calling EVM?

shermike: I thought about this. And it is ok for external transactions, but doesn't work for internal, since it can forward gas to outbound transactions. Probably we can buy only block limit in such case and forward leftover and rest of inbound gas. Let me check.

shermike: Actually, it doesn't work for external transactions as well, because they also can send another transactions.
But I slightly reworked initial implementation, now there are no changes inside EVM, but gas is cut off before running EVM.

olegrok (approved):
But seems you need to fix existing test

PR #744, opened 2025-04-03
[GH-127] Give each command a separate parameter object

### LINKED ISSUE #127: Erroneous global state mutation when configuring command line options
We have a [global variable](https://github.com/NilFoundation/nil/blob/691625c3edbd07fcce7e2bfcf5c612c277ac73b6/nil/cmd/nil/internal/wallet/params.go#L25-L25) being [changed](https://github.com/NilFoundation/nil/blob/691625c3edbd07fcce7e2bfcf5c612c277ac73b6/nil/cmd/nil/internal/wallet/call-readonly.go#L34-L34) while [registering commands](https://github.com/NilFoundation/nil/blob/691625c3edbd07fcce7e2bfcf5c612c277ac73b6/nil/cmd/nil/internal/wallet/wallet.go#L37-L37) in the CLI that is [used in other commands](https://github.com/NilFoundation/nil/blob/691625c3edbd07fcce7e2bfcf5c612c277ac73b6/nil/cmd/nil/internal/wallet/new.go#L45-L45).

_It probably doesn't make sense to be fixed if we completely migrate the CLI to nil.js and remove the old code._

Un1oR:
Closes #127

### DIFF
--- a/docs/tests/working-with-smart-contracts-cli.test.mjs
+++ b/docs/tests/working-with-smart-contracts-cli.test.mjs
@@ -14,24 +14,17 @@ import {
 const util = require("node:util");
 const exec = util.promisify(require("node:child_process").exec);
 
-const SALT = BigInt(Math.floor(Math.random() * 10000));
-
 const CONFIG_FILE_NAME = "./tests/tempWorkingWithSmartContracts.ini";
 
 const CONFIG_FLAG = `--config ${CONFIG_FILE_NAME}`;
 
-//startRetailerDeploymentCommand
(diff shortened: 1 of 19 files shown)


olegrok (approved):

Un1oR:
```
nil.docs>    × CLI deployment tests > external deployment of Retailer and Manufacturer is successful 691ms
nil.docs>      → Command failed: nil contract deploy ./tests/Retailer/Retailer.bin --shard-id 1 --salt 8432 --config ./tests/tempWorkingWithSmartContracts.ini
nil.docs> Error: rpc error: {"code":-32000,"message":"ContractAddressCollision"}
```
I don't understand why this started happening with the salt being random :thinking:

khannanov-nil (approved):

PR #746, opened 2025-04-03
fix duplication of shardId in logs

olegrok:
Zerolog doesn't deduplicate log entry tags. So it's enough to set shardId for root logger. Otherwise we will have shardId twice in JSON. This patch removes several FieldShardId fields in case if it's defined in the root logger.


Un1oR:
If we don't use Zerolog directly now, why not make under `if assert.Enable` a check for already set fields?

Un1oR (approved):

olegrok:
> If we don't use `zerolog` directly now, why not make under `if assert.Enable` a check for already set fields?

I'll think about it

PR #754, opened 2025-04-04
feat(rollup contract): add state reset function

x-mass:
## Short Summary

Add state reset function to the rollup contract so an admin could revert to specified state root.

## What Changes Were Made

- Added `resetState` function. When called, it traverses a chain of `batchInfoRecords` (chained via `oldStateRoot`) until `newStateRoot` equals `targetStateRoot`, removing batch info and corresponding `stateRootIndex` entries. Any unrelated or dangling batch entries are left untouched.
- Remove `lastCommittedBatchIndex` since it is not used anyway, and to keep it consistent during state reset we had to introduce another `committedBatchHistory`.

## Checklist

### DIFF
--- a/rollup-bridge-contracts/contracts/NilRollup.sol
+++ b/rollup-bridge-contracts/contracts/NilRollup.sol
@@ -18,7 +18,7 @@ contract NilRollup is
     INilRollup
 {
     /*//////////////////////////////////////////////////////////////////////////
-                             NILROLLUP-ERRORS   
+                             NILROLLUP-ERRORS
     //////////////////////////////////////////////////////////////////////////*/
 
     /// @dev Invalid owner address.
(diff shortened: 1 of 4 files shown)


akokoshn (commented):
We also need to implement call this method from SC, when it resets state: https://github.com/NilFoundation/nil/blob/e8e154d3299cb35f9a2e6246d0b1d608eae77715/nil/services/synccommittee/core/fetching/aggregator.go#L169

akokoshn (commented):
looks ok for me, let's wait for @defistar review

defistar:
> looks ok for me, let's wait for @defistar review

@akokoshn @x-mass looks good for me
this may cause merge conflict with current working branch for bridge contracts
we can port these changes on to it after the merge

defistar (approved):

PR #761, opened 2025-04-05
feat: implemented nil config commands

### LINKED ISSUE #454: [cli.js] Support config.ini
Currently, cli.js generates props and store them in config.ini file. The config.ini updates on new smart account creation. We need to implement commands to work with config.ini values.

Project: https://github.com/NilFoundation/nil/tree/main/clijs

- [x] config get: get particular value from config.ini
- [x] config init: init a new config.ini. Add forced option
- [x] config set: set particular value to config.ini
- [x] config show: show the whole content of config.ini file

Ref: https://github.com/NilFoundation/nil/blob/main/nil/cmd/nil/internal/config/command.go

0xNgoo:
## Short Summary
Implemented `clijs` config commands includng `get`, `show`, `init` and `set`

## What Changes Were Made

- implemented `nil config get` command
- implemented `nil config show` command
- implemented `nil config init` command
- implemented `nil config set` command

## Related Issue

Fixes #454
or  
Related to #[Issue Number]

## Checklist

### DIFF
--- a/clijs/src/commands/config/set.ts
+++ b/clijs/src/commands/config/set.ts
@@ -1,0 +1,48 @@
+import { Args } from "@oclif/core";
+import { BaseCommand } from "../../base.js";
+import { ConfigKeys } from "../../common/config.js";
+
+// Supported options in the config file
+const supportedOptions = new Set([
+  ConfigKeys.RpcEndpoint,
+  ConfigKeys.CometaEndpoint,
+  ConfigKeys.FaucetEndpoint,
+  ConfigKeys.PrivateKey,
+  ConfigKeys.Address,
+]);
+
+export default class ConfigSet extends BaseCommand {
+  static override description = "Set the value of a key in the config file";
+
+  static override examples = ["<%= config.bin %> <%= command.id %>"];
+
+  static args = {
+    key: Args.string({
+      description: "The key to set in the config file",
+      required: true,
+    }),
+    value: Args.string({
+      description: "The value to set for the key",
+      required: true,
+    }),
+  };
+
+  public async run(): Promise<string> {
+    const { args } = await this.parse(ConfigSet);
+    const { key, value } = args;
+
+    if (!this.configManager) {
+      this.error("Config manager is not initialized");
+    }
+
+    if (!supportedOptions.has(key as ConfigKeys)) {
+      this.error(`Key "${key}" is not known`);
+    }
+
+    this.configManager.updateConfig(ConfigKeys.NilSection, key, value);
+
+    const formattedOutput = `Set "${key}" to "${value}"`;
+
+    return formattedOutput;
+  }
+}
--- a/clijs/src/sea.ts
+++ b/clijs/src/sea.ts
@@ -10,6 +10,11 @@ import AbiCommand from "./commands/abi";
 import AbiDecode from "./commands/abi/decode";
 import AbiEncode from "./commands/abi/encode";
 import BlockCommand from "./commands/block";
+import Config from "./commands/config";
+import ConfigGet from "./commands/config/get";
+import ConfigInit from "./commands/config/init";
+import ConfigSet from "./commands/config/set";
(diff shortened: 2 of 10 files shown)


clijs/src/sea.ts
```diff
@@ -25,6 +25,12 @@ import SystemCommand from "./commands/system";
 import ChainId from "./commands/system/chain-id";
 import GasPrice from "./commands/system/gas-price";
 import Shards from "./commands/system/shards";
+import Config from "./commands/config";
```
Un1oR, line 28: Please keep this file in alphabetical order

Un1oR:
I guess the problem with `config show` not printing anything can be solved this way (similar to other commands):
```
--- a/clijs/src/commands/config/show.ts
+++ b/clijs/src/commands/config/show.ts
@@ -22,6 +22,7 @@ export default class ConfigShow extends BaseCommand {
         formattedOutput += `${key.padEnd(18)}: ${value}\n`;
       }
     }
+    this.log(formattedOutput);
 
     return formattedOutput;
   }
```

clijs/src/commands/config/set.ts
```diff
@@ -0,0 +1,48 @@
...
+      this.error("Config manager is not initialized");
+    }
+
+    if (!supportedOptions.has(key as ConfigKeys)) {
+      this.error(`Key "${key}" is not known`);
+    }
+
+    this.configManager.updateConfig(ConfigKeys.NilSection, key, value);
+
+    const formattedOutput = `Set "${key}" to "${value}"`;
```
Un1oR, line 44: No handling of the `quiet` flag

clijs/src/commands/config/set.ts
```diff
@@ -0,0 +1,48 @@
...
+      description: "The value to set for the key",
+      required: true,
+    }),
+  };
+
+  public async run(): Promise<string> {
+    const { args } = await this.parse(ConfigSet);
+    const { key, value } = args;
+
+    if (!this.configManager) {
```
Un1oR, line 34: Is this possible? We initialize it in `BaseCommand` and I don't see any such check in other commands. I find it unnecessary and only confusing.

clijs/test/commands/config/get.test.ts
```diff
@@ -0,0 +1,21 @@
...
+import { ConfigKeys } from "../../../src/common/config.js";
+import { CliTest } from "../../setup.js";
+
+describe("config:get", () => {
+  CliTest("gets a value from the config file", async ({ cfgPath, runCommand }) => {
+    const { result } = await runCommand(["config", "get", ConfigKeys.RpcEndpoint]);
+    const configManager = new ConfigManager(cfgPath);
+
+    expect(result).to.equal(
+      configManager.getConfigValue(ConfigKeys.NilSection, ConfigKeys.RpcEndpoint),
```
Un1oR, line 12: Aren't we expecting `undefined` here if we haven't set it previously? It seems to me that the `get` test should create (probably manually as writing a string to a file) a config of the expected content, and the test should check that we get that particular value (rather than just comparing the identity of the code inside the command with the same code outside it).

clijs/test/commands/config/init.test.ts
```diff
@@ -0,0 +1,11 @@
+import { describe, expect } from "vitest";
+import { CliTest } from "../../setup.js";
+
+describe("config:init", () => {
+  CliTest("initializes a new config file", async ({ runCommand }) => {
+    const { result } = await runCommand(["config", "init"]);
+
+    expect(result).toBeTypeOf("string");
+    expect(result).toBe("config initialized");
```
Un1oR, line 9: I think this test is also not very good and doesn't check the real state of affairs. I think you should compare the contents of the config file with the expected contents after executing the command.

idea404:
HI @0xNgoo please confirm you are still working on the feedback here. Otherwise I'll close this PR

Gezort:
close as inactive

PR #768, opened 2025-04-07
relayer: do not hardcode ABI of L1/L2 contracts

oclaw:
## What Changes Were Made

-
-
-

## Related Issue

Fixes #[Issue Number]  
or  
Related to #[Issue Number]

## Checklist

- [ ] I have read and followed the [Contributing Guide](https://github.com/NilFoundation/nil/blob/main/CONTRIBUTION-GUIDE.md)
- [ ] I have tested the changes locally


defistar (approved):

AndreyMlashkin:
Nix part looks good to me.
But I can't review the whole PR, as I am not familiar with go code or make-files in nil

nil/services/relayer/internal/l2/contract_wrapper.go
```diff
@@ -104,11 +96,7 @@ func NewL2ContractWrapper(
 			Msg("looks like L2 contract is not deployed")
 	}
 
-	abiFile, err := os.Open(config.ContractABIPath)
-	if err != nil {
-		return nil, fmt.Errorf("failed to open '%s' L2 ABI file: %w", config.ContractABIPath, err)
-	}
-	defer abiFile.Close()
+	abiFile := bytes.NewReader(GetL2BridgeMessengerABI())
```
x-mass, line 99: Mb return `ABI` instance from `GetL2BridgeMessengerABI()` func? Or rename it to sth like `GetL2BridgeMessengerABIString`

nil/services/relayer/Makefile.inc
```diff
@@ -0,0 +1,18 @@
+root_services = nil/services
+root_relayer = nil/services/relayer
+
+.PHONY: $(root_relayer)/embed_l2_abi
+$(root_relayer)/embed_l2_abi:
+	@cat $(root_services)/rollup-bridge-contracts-compiled/contracts/bridge/l2/L2BridgeMessenger.sol/L2BridgeMessenger.json | jq .abi > $(root_relayer)/internal/l2/L2BridgeMessenger.json.abi
```
x-mass, line 6: Imho, `.json` is a primary extension
```suggestion
	@cat $(root_services)/rollup-bridge-contracts-compiled/contracts/bridge/l2/L2BridgeMessenger.sol/L2BridgeMessenger.json | jq .abi > $(root_relayer)/internal/l2/L2BridgeMessenger.abi.json
```

oclaw: It was done to be aligned with current gitignore rules (they ignore .abi). I can add narrow rule for this file here but not sure if it is worth it

nil/services/relayer/internal/l2/contract_wrapper.go
oclaw: Good idea. Did ABI parsing in the `init()` function

.github/workflows/main.yml
```diff
@@ -106,7 +106,7 @@ jobs:
       - name: build binaries
         run: |
           export PATH=$PATH:$(go env GOPATH)/bin
-          make
+          make nild nil
```
akokoshn, line 109: i guess we can build all binaries excluding relayer here

nil/Makefile.inc
```diff
@@ -1,10 +1,18 @@
 include nil/services/synccommittee/Makefile.inc
+include nil/services/relayer/Makefile.inc
 
 root_client = nil/client
 root_vm = nil/internal/vm
 root_db = nil/internal/db
+root_services = nil/services
```
akokoshn, line 7: already defined in `nil/services/relayer/Makefile.inc`

oclaw: Except relayer and sync committee related stuff. Anyway we have agreed with @olegrok that these two are enough for basic build check (since we do not run the full nix build here)

.github/workflows/main.yml
oclaw: Except relayer and sync committee related stuff. Anyway we have agreed with @olegrok that these two are enough for basic build check (since we do not run the full nix build here)

nil/Makefile.inc
oclaw: cleaned up

.github/workflows/main.yml
akokoshn: ok

nil/services/relayer/internal/l1/abi.go
```diff
@@ -0,0 +1,3 @@
+package l1
+
+//go:generate bash -c "jq '.abi' ../../../rollup-bridge-contracts-compiled/contracts/bridge/l1/L1BridgeMessenger.sol/L1BridgeMessenger.json | go run github.com/ethereum/go-ethereum/cmd/abigen --abi - --pkg=l1 --out=./l1_bridge_messenger_contract_abi_generated.go"
```
olegrok, line 3: Is it possible to move this content directly into Makefile?

oclaw: Do you mean `jq` invocation? It generally is possible but will require to keep a local copy of the ABI file (like we are doing for l2 case).

nil/services/relayer/internal/l2/contract_wrapper.go
```diff
@@ -125,7 +105,7 @@ func NewL2ContractWrapper(
 		privateKey:       pk,
 		smartAccountAddr: smartAccountAddr,
 		contractAddr:     contractAddr,
-		abi:              contractABI,
+		abi:              GetL2BridgeMessengerABI(),
```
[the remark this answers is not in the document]
oclaw: In current implementation it is not possible. `//go embed` is executed at the compile time and `l2.init()` is invoked before `main` of the binary, so service will just not start without initialized ABI object

nil/services/relayer/internal/l1/abi.go
olegrok: No. I mean it's ok to call `jq '.abi' ../../../rollup-bridge-contracts-compiled/contracts/bridge/l1/L1BridgeMessenger.sol/L1BridgeMessenger.json | go run github.com/ethereum/go-ethereum/cmd/abigen --abi - --pkg=l1 --out=./l1_bridge_messenger_contract_abi_generated.go` directly from Makefile.

oclaw: Done. Now go generate includes only `abigen` invocation

olegrok: Is it possible to move `abigen` call also to Makefile?

oclaw: Theoretically yes. But what is the need to do that? We widely use `go:generate` command across the project

oclaw: From my point of view it is the most clear solution we have now - preprocessing of the file is done in Makefile and generation of Go code is done via Go tools

olegrok: Using `go generate` inside go files was a mistake. It seems like we need to maintain two build systems.

Even authors say that everything can be done via make (https://go.dev/blog/generate):

> Go generate **does nothing that couldn’t be done with Make or some other build mechanism**, but it comes with the go tool—no extra installation required—and fits nicely into the Go ecosystem.

I suggest to avoid introducing new usages of go generate in future and reduce its usage if possible.

olegrok: Just copy a content of `generate.go` file into Makefile. It should work.

olegrok (approved):

PR #772, opened 2025-04-07
Several improvements after debugging cluster with Stresser

shermike:
- Improve Stresser:
	- Remove ability to check transactions result
	- Now the workload file can contain several different tasks, each of which can be accessed by name
- Remove `Timeout` field from Block struct
- Support txpool status in indexer
- Remove the limit on the maximum number of internal transactions in the Proposer

### DIFF
--- a/nil/internal/collate/proposer.go
+++ b/nil/internal/collate/proposer.go
@@ -21,7 +21,6 @@ import (
 )
 
 const (
-	defaultMaxInternalTxns               = 1000
 	defaultMaxGasInBlock                 = types.DefaultMaxGasInBlock
 	maxTxnsFromPool                      = 1000
 	defaultMaxForwardTransactionsInBlock = 200
@@ -48,9 +47,6 @@ type proposer struct {
 func newProposer(params *Params, topology ShardTopology, pool TxnPool, logger logging.Logger) *proposer {
 	if params.MaxGasInBlock == 0 {
 		params.MaxGasInBlock = defaultMaxGasInBlock
-	}
-	if params.MaxInternalTransactionsInBlock == 0 {
-		params.MaxInternalTransactionsInBlock = defaultMaxInternalTxns
 	}
 	if params.MaxForwardTransactionsInBlock == 0 {
 		params.MaxForwardTransactionsInBlock = defaultMaxForwardTransactionsInBlock
@@ -367,7 +363,6 @@ func (p *proposer) handleTransactionsFromNeighbors(tx db.RoTx) error {
 
 	checkLimits := func() bool {
 		return p.executionState.GasUsed < p.params.MaxGasInBlock &&
-			len(p.proposal.InternalTxnRefs) < p.params.MaxInternalTransactionsInBlock &&
 			len(p.proposal.ForwardTxnRefs) < p.params.MaxForwardTransactionsInBlock
 	}
 
--- a/nil/internal/collate/proposer.go
+++ b/nil/internal/collate/proposer.go
@@ -22,7 +22,7 @@ import (
 
 const (
 	defaultMaxGasInBlock                 = types.DefaultMaxGasInBlock
-	maxTxnsFromPool                      = 1000
+	maxTxnsFromPool                      = 10_000
 	defaultMaxForwardTransactionsInBlock = 200
 
 	validatorPatchLevel = 1
--- a/nil/services/indexer/indexer.go
+++ b/nil/services/indexer/indexer.go
@@ -20,6 +20,8 @@ const (
 	InitialRoundsAmount = 1000
 
 	maxFetchSize = 500
+
+	timeoutWaitingRpc = 5 * time.Minute
 )
 
 type Indexer struct {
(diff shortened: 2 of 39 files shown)


nil/internal/collate/proposer.go
```diff
@@ -49,9 +48,6 @@ func newProposer(params *Params, topology ShardTopology, pool TxnPool, logger lo
 	if params.MaxGasInBlock == 0 {
 		params.MaxGasInBlock = defaultMaxGasInBlock
 	}
-	if params.MaxInternalTransactionsInBlock == 0 {
```
olegrok, line 52: Why?

nil/services/rpc/rawapi/local_txnpool.go
```diff
@@ -27,11 +27,11 @@ func (api *LocalShardApi) SendTransaction(ctx context.Context, encoded []byte) (
 }
 
 func (api *LocalShardApi) GetTxpoolStatus(ctx context.Context) (uint64, error) {
-	return uint64(api.txnpool.GetQueue().Len()), nil
+	return uint64(api.txnpool.GetPendingLength()), nil
 }
 
 func (api *LocalShardApi) GetTxpoolContent(ctx context.Context) ([]*types.Transaction, error) {
-	txns, err := api.txnpool.Peek(api.txnpool.GetQueue().Len())
+	txns, err := api.txnpool.Peek(1_000_000)
```
olegrok, line 34: txpool has "Size" field in config. And you can try to use this field instead of hardcoded constant.

nil/tests/economy/economy_test.go
```diff
@@ -81,6 +81,8 @@ func (s *SuiteEconomy) SetupSuite() {
 		ZeroState:            zeroState,
 		CollatorTickPeriodMs: 200,
 		RunMode:              nilservice.CollatorsOnlyRunMode,
+
+		DisableConsensus: true,
```
olegrok, line 85: Please, revert this change

nil/tests/timeouts_no_race.go
```diff
@@ -5,9 +5,9 @@ package tests
 import "time"
 
 const (
-	ReceiptWaitTimeout    = 15 * time.Second
+	ReceiptWaitTimeout    = 15 * time.Minute
```
olegrok, line 8: Debug?

nil/services/stresser/workload/send_value.go
```diff
@@ -0,0 +1,41 @@
...
+	batchLimit := len(w.params.Contracts)
+	options := &core.TxParams{
+		FeePack: types.NewFeePackFromGas(100_000_000),
+		Value:   types.NewValueFromUint64(1000),
+	}
+	for i := 0; i < w.Iterations; i += batchLimit {
+		wg := &sync.WaitGroup{}
+		for j := 0; j < batchLimit && i+j < w.Iterations; j++ {
+			wg.Add(1)
+			go func(idx int) {
```
olegrok, line 30: As I see you use this pattern several times. I suggest to move it into a separate function.

nil/internal/collate/proposer.go
olegrok: That's true but we need to support some quota for external messages. Otherwise once we can stop to process them at all.

shermike: There is no such quota yet. And for sure it won't be an absolute value of transactions count. It should be carefully researched, and probably it will be a percentage cap on gas for internal transactions. 
Also, since internal transactions cannot be endless by itself, i.e. external transactions start them, maybe quota is not needed at all. cc @alex-damjanovic

nil/client/rpc/client.go
```diff
@@ -164,6 +164,39 @@ func (b *BatchRequestImpl) SendTransactionViaSmartContract(
...
+func (b *BatchRequestImpl) SendExternalTransaction(
+	ctx context.Context,
+	calldata types.Code,
+	contractAddress types.Address,
+	pk *ecdsa.PrivateKey,
+	fee types.FeePack,
+) (uint64, error) {
+	id := len(b.requests)
+
+	extTxn, err := client.CreateExternalTransaction(ctx, b.client, calldata, contractAddress, fee, false, 0)
```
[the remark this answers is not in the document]
shermike: Actually, it is not used. So I removed it as it requires additional effort to support sequential seqno for the same contract. It should be done in a separate PR.

nil/internal/collate/proposer.go
alex-damjanovic: It is a viable concern @olegrok, and we thought about it. At the moment we are discussing splitting the available  block space between external and internal transactions, which would adjust on block-by-block basis. (Full external transaction gas portion  = a bit more space for external transaction gas in next block, etc) 
It is still in design/discussion phase.  I will let you know once we finalize it. 

But nevertheless, we will only use gas units, not number of transactions to measure capacity, no matter how we proceed.

olegrok: Well, thanks!

nil/services/stresser/workload/send_value.go
shermike: Yes, good idea, done. Thanks.

nil/services/indexer/indexer.go
```diff
@@ -273,6 +298,51 @@ func (e *Indexer) runBottomFetcher(ctx context.Context, shardId types.ShardId, t
...
+
+	numShards, err := e.client.GetNumShards(ctx)
+	if err != nil {
+		logger.Error().Err(err).Msg("Failed to get number of shards")
+	}
+
+	concurrent.RunTickerLoop(ctx, 1*time.Second, func(ctx context.Context) {
+		statuses := make([]*driver.TxPoolStatus, 0, numShards)
+		for shardId := range numShards {
+			txPool, err := e.client.GetTxpoolStatus(ctx, types.ShardId(shardId))
```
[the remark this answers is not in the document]
olegrok: Lol

Issue #770, opened 2025-04-07
Alert if sync is lagging behind


dmtrskv:
Archive node and indexer should alert if they are lagging behind on some shards.
This is done via otlp metrics on syncers and signoz alerts.

PR #781, opened 2025-04-09
[telemetry] Metrics for validators

dmtrskv:
Moved metrics from block generator to validator.
Validator is the correct place, because generator shouldn't produce side effects.

Separated block id from other metrics, so that it can be reported by syncers, not only by producers.

Part of https://github.com/NilFoundation/nil/issues/770

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,7 +82,7 @@ format: generated
 	GOPROXY= go mod vendor
 	golangci-lint fmt
 
-lint: format golangci-lint
+lint: format golangci-lint checklocks
 
 checklocks: generated
 	@export GOFLAGS="$$GOFLAGS -tags=test,goexperiment.synctest"; \
--- a/nil/internal/collate/metrics.go
+++ b/nil/internal/collate/metrics.go
@@ -1,8 +1,9 @@
-package execution
+package collate
 
 import (
 	"context"
 
+	"github.com/NilFoundation/nil/nil/internal/execution"
 	"github.com/NilFoundation/nil/nil/internal/telemetry"
 	"github.com/NilFoundation/nil/nil/internal/telemetry/telattr"
 	"github.com/NilFoundation/nil/nil/internal/types"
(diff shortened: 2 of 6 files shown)


Makefile
```diff
@@ -82,7 +82,7 @@ format: generated
 	GOPROXY= go mod vendor
 	golangci-lint fmt
 
-lint: format golangci-lint
+lint: format golangci-lint checklocks
```
dmtrskv, line 85: Looks reasonable to me, since I run `make lint` locally a lot.

nil/internal/telemetry/telattr/metric.go
```diff
@@ -5,6 +5,8 @@ import (
 	"go.opentelemetry.io/otel/metric"
 )
 
-func With(attrs ...attribute.KeyValue) metric.MeasurementOption {
+type MetricOption = metric.MeasurementOption
```
dmtrskv, line 8: so that we can use local types without referring to a third party

nil/internal/collate/metrics.go
```diff
@@ -87,31 +74,31 @@ func (mh *MetricsHandler) initMetrics(meter metric.Meter) error {
...
 
-func (mh *MetricsHandler) EndProcessingMeasurement(ctx context.Context, counters *BlockGeneratorCounters) {
+func (mh *MetricsHandler) EndProcessingMeasurement(ctx context.Context, counters *execution.BlockGeneratorCounters) {
 	mh.measurer.Measure(ctx)
 	mh.internalTxnCounter.Add(ctx, counters.InternalTransactions, mh.option)
 	mh.externalTxnCounter.Add(ctx, counters.ExternalTransactions, mh.option)
 	mh.deployTxnCounter.Add(ctx, counters.DeployTransactions, mh.option)
 	mh.execTxnCounter.Add(ctx, counters.ExecTransactions, mh.option)
 	mh.coinsUsed.Record(ctx, int64(counters.CoinsUsed.Uint64()), mh.option)
+	mh.gasPrice.Record(ctx, int64(counters.GasPrice.Uint64()), mh.option)
```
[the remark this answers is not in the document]
dmtrskv: little bullshitter

olegrok (approved):

PR #784, opened 2025-04-09
[txpool] Fix seqno gap

shermike:
The problem was in the following.
Consider the case where two transactions with `seqno` 1 and 4 are added to the same account. Previously, after executing the transaction with `seqno` 1, the next `seqno` was set to 4. As a result, transactions with `seqno` 2 and 3 were rejected with a `Seqno too low` error. Now, the next `seqno` is set to 2, allowing transactions with `seqno` 2 and 3 to be processed as expected.


olegrok (approved):
Isn't it possible to make `defaultBatchConcurrency=2` then?

Issue #489, opened 2025-03-06
Add relay to make private nodes available publicly


dmtrskv:
The cluster should be available for libp2p connections from the public network (mainly from archive nodes).

A libp2p relay node can be deployed on a machine from the cluster network on which libp2p ports are publicly available.

dmtrskv:
Added: relay service and ansible scripts for its deployment.

PR #785, opened 2025-04-09
[archive] Misc improvements concerning relays

dmtrskv:
Some improvements made after testing archive nodes with relays.

- Archive nodes serve blocks over libp2p: call SetBlockRequestHandler in syncer.
- Update to syncer init: no bootstrap peers is fine; more explicit handling of the absence of peers with whom to check the version.

Part of https://github.com/NilFoundation/nil/issues/489

### DIFF
--- a/nil/internal/collate/syncer.go
+++ b/nil/internal/collate/syncer.go
@@ -151,11 +151,6 @@ func (s *Syncer) fetchRemoteVersion(ctx context.Context) (NodeVersion, error) {
 }
 
 func (s *Syncer) fetchSnapshot(ctx context.Context) error {
-	if len(s.config.BootstrapPeers) == 0 {
-		s.logger.Warn().Msg("No bootstrap peers to fetch snapshot from")
-		return nil
-	}
-
 	var err error
 	for _, peer := range s.config.BootstrapPeers {
 		err = fetchSnapshot(ctx, s.networkManager, peer, s.db, s.logger)
@@ -171,6 +166,11 @@ func (s *Syncer) Init(ctx context.Context, allowDbDrop bool) error {
 		return nil
 	}
 
+	if len(s.config.BootstrapPeers) == 0 {
+		s.logger.Info().Msg("No bootstrap peers. Skipping initialization")
+		return nil
+	}
+
 	version, err := s.getLocalVersion(ctx)
 	if err != nil {
 		return err
@@ -178,10 +178,14 @@ func (s *Syncer) Init(ctx context.Context, allowDbDrop bool) error {
 
 	remoteVersion, err := s.fetchRemoteVersion(ctx)
 	if err != nil {
-		// todo: when all shards can handle the new protocol, we should return an error here
-		s.logger.Warn().Err(err).Msgf(
-			"Failed to fetch remote version. For now we assume that local version %s is up to date", version)
-		return nil
+		// Nodes with allowDbDrop are supposed to be secondary, so they must sync with some reliable peer.
+		// We need some nodes to start without fetching a remote version
+		// otherwise, the cluster won't ever start from scratch.
+		if !allowDbDrop {
+			s.logger.Warn().Err(err).Msg("Failed to fetch remote version. We'll assume that our version is up to date")
+			return nil
+		}
+		return err
 	}
 
 	if version.ProtocolVersion != remoteVersion.ProtocolVersion {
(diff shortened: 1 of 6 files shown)


olegrok (approved):
LGTM. But you need to fix tests:
```
2025-04-09T17:40:30.3062053Z nil> --- FAIL: TestArchiveNodeWithoutBootstrapPeers (112.81s)
2025-04-09T17:40:30.3062394Z nil>     --- FAIL: TestArchiveNodeWithoutBootstrapPeers/TestCheckBlocksGeneraion (31.24s)
2025-04-09T17:40:30.3062603Z nil>         suite_archive_node.go:95:
2025-04-09T17:40:30.3062959Z nil>                    Error Trace:    /build/source/nil/tests/common.go:140
2025-04-09T17:40:30.3063464Z nil>                                                            /build/source/nil/tests/archive_node/suite_archive_node.go:95
2025-04-09T17:40:30.3063855Z nil>                                                            /build/source/nil/tests/archive_node/basic_test.go:24
2025-04-09T17:40:30.3064115Z nil>                    Error:          Condition never satisfied
2025-04-09T17:40:30.3064707Z nil>                    Test:           TestArchiveNodeWithoutBootstrapPeers/TestCheckBlocksGeneraion
```

nil/tests/common.go
```diff
@@ -153,18 +153,10 @@ func WaitZerostate(t *testing.T, client client.Client, shardId types.ShardId) {
...
-	require.Eventuallyf(
-		t,
-		func() bool {
-			block, err := client.GetBlock(t.Context(), shardId, transport.BlockNumber(block.Number+1), false)
-			return err == nil && block != nil
-		},
-		ShardTickWaitTimeout,
-		ShardTickPollInterval,
-		"Failed to wait for shard %d tick after block %d", shardId, block.Number)
+	require.Eventually(t, func() bool {
```
olegrok, line 156: Now this code checks that at least one block is returned. Actually it can be zerostate block. It doesn't wait for block ticks. In case if consensus fails this function passes.

olegrok (changes requested):

nil/tests/common.go
olegrok: For network startup it's probably ok. But in case you need to be sure that some transaction doesn't break validators  after execution it won't work anymore

olegrok: But ok. It's probably not required for now.

olegrok (approved):

nil/tests/async_await/async_await_test.go
```diff
@@ -286,7 +280,7 @@ func (s *SuiteAsyncAwait) TestFactorial() {
 }
 
 func (s *SuiteAsyncAwait) TestFibonacci() {
-	data := s.AbiPack(s.abiTest, "fibonacci", int32(6))
+	data := s.AbiPack(s.abiTest, "fibonacci", int32(4))
```
dmtrskv, line 283: 6 was too much, because each calculation requires two async calls

nil/internal/execution/block_generator.go
```diff
@@ -153,6 +153,11 @@ func (p *BlockGenerator) CollectGasPrices(prevBlockId types.BlockNumber) []types
 			shardHash = mainBlock.PrevBlock
 		}
 
+		if shardHash.Empty() {
```
dmtrskv, line 156: it produced unneeded warnings

dmtrskv:
Moved test and log improvements to https://github.com/NilFoundation/nil/pull/799

nil/internal/collate/syncer.go
```diff
@@ -167,17 +162,26 @@ func (s *Syncer) Init(ctx context.Context, allowDbDrop bool) error {
...
 
 	remoteVersion, err := s.fetchRemoteVersion(ctx)
 	if err != nil {
-		// todo: when all shards can handle the new protocol, we should return an error here
-		s.logger.Warn().Err(err).Msgf(
-			"Failed to fetch remote version. For now we assume that local version %s is up to date", version)
-		return nil
+		// Nodes with allowDbDrop are supposed to be secondary, so they must sync with some reliable peer.
+		// We need some nodes to start without fetching a remote version
+		// otherwise, the cluster won't ever start from scratch.
```
Un1oR, line 179: Don't we have enough to return with `nil` by the condition `if len(s.config.BootstrapPeers) == 0` above? I don't really understand this comment and the conditions in this branch. Could you please explain in more detail?

dmtrskv: When we start the first node in the cluster and it has bootstrap peers, it will fail here if we return err.
So, we either must have some nodes without bootstrap peers, or do smth like I did, when we return error only for nodes that allow db drop (as they are probably some syncers).

Un1oR: What's wrong with having a certain number of nodes without bootstrap peers? After all, if these are the "initial" nodes, they have nowhere to have bootstrap peers. Isn't that the case now?

PR #789, opened 2025-04-10
[PoC] Message Queue contract

olegrok:
⚠️ **Note**: This patch is **not intended to be merged as-is**. It serves only as a demonstration of how a Message Queue contract could be implemented.

There are several issues that arise with this approach, and **not all of them have been resolved yet**.

---

## Summary of Changes

- A **Message Queue contract** is deployed on **each shard**.
- The contract currently provides only basic functionality:
  - Adding messages to the queue.
  - Clearing the queue.
- **Queue clearing is performed at the beginning of each block** via a **system call** sent by proposer.
- The most interesting part is **message persistence**: both the message data and the sender’s address are saved.
- Only `SmartAccount.send` method is ported to this approach.
- At the end of the block production cycle, all messages from the queue are collected and included in the **"out transaction tree"**.
  - In the future, we might **completely remove the transaction tree**.
  - For now, with both approaches coexisting, this hybrid implementation is used.
- **Malformed or invalid messages are currently dropped silently**.

---

## Potential Next Steps

- Reduce and consolidate existing message-sending interfaces.
- Replace the `Nil.sendTransaction` interface with `MessageQueue.sendRawTransaction`.
- Improve error handling and diagnostics.
- ...

### DIFF
--- a/nil/internal/execution/block_generator.go
+++ b/nil/internal/execution/block_generator.go
@@ -271,7 +271,6 @@ func (g *BlockGenerator) prepareExecutionState(proposal *Proposal, gasPrices []t
 				Msg("Failed to handle out message")
 			continue
 		}
-		fmt.Println("====================== out txn added", msg.Address, "======================")
 	}
 
 	g.executionState.ChildShardBlocks = make(map[types.ShardId]common.Hash, len(proposal.ShardHashes))
--- a/nil/internal/execution/mq_manager.go
+++ b/nil/internal/execution/mq_manager.go
@@ -1,7 +1,6 @@
 package execution
 
 import (
-	"encoding/json"
 	"fmt"
 
 	"github.com/NilFoundation/nil/nil/internal/config"
@@ -100,6 +99,9 @@ func HandleOutMessage(es *ExecutionState, msg *message) error {
 		return vm.ErrTransactionToMainShard
 	}
 
+	// TODO: support estimate fee for such messages
+	payload.FeeCredit = types.MaxFeePerGasDefault
+
 	// TODO: withdrawFunds should be implemneted
 	// if err := withdrawFunds(es, msg.Address, payload.Value); err != nil {
 	// 	return nil, fmt.Errorf("withdraw value failed: %w", err)
@@ -120,9 +122,5 @@ func HandleOutMessage(es *ExecutionState, msg *message) error {
 	}
 
 	_, err = es.AddOutTransaction(msg.Address, &payload)
-	if err == nil {
-		pJson, _ := json.MarshalIndent(payload, "", "  ")
-		fmt.Println("out transaction added", msg.Address, string(pJson))
-	}
 	return err
 }
--- a/nil/tests/basic/basic_test.go
+++ b/nil/tests/basic/basic_test.go
@@ -722,15 +722,15 @@ func (s *SuiteRpc) TestRpcBlockContent() {
 		block, err = s.Client.GetBlock(s.Context, types.BaseShardId, "latest", false)
 		s.Require().NoError(err)
 
-		return len(block.TransactionHashes) > 0
+		return len(block.TransactionHashes) > 1
 	}, 6*time.Second, 50*time.Millisecond)
 
 	block, err = s.Client.GetBlock(s.Context, types.BaseShardId, block.Hash, true)
 	s.Require().NoError(err)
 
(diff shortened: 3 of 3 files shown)


nil/internal/execution/block_generator.go
```diff
@@ -260,6 +260,19 @@ func (g *BlockGenerator) prepareExecutionState(proposal *Proposal, gasPrices []t
 		g.executionState.AppendForwardTransaction(txn)
 	}
 
+	messages, err := GetMessageQueueContent(g.executionState)
```
shermike, line 263: Shouldn't we get messages queue from other shards targeting to our shard?

nil/contracts/solidity/system/MessageQueue.sol
```diff
@@ -0,0 +1,29 @@
...
+    function getMessages() external view returns (Message[] memory) {
+        return queue;
+    }
+
+    function clearQueue() external {
+        require(msg.sender == address(this), "clearQueue: only MessageQueue contract can be caller of this function");
+        delete queue;
+    }
+
+    Message[] private queue;
```
shermike, line 28: Maybe it is worth to separate this array by shard. So the other shard can fetch only messages addressed to him: `getMessages(uint shardId)`

nil/internal/execution/block_generator.go
olegrok: We should do that by proposer. I suggest to implement step-by-step solution. It's the first step in this directon.

Right now an approach doesn't break existing code. We just copy a message queue content to an existing outbound messages tree. Ideally, we need to drop old approach and then it will be possible to use only queue account state.

nil/contracts/solidity/system/MessageQueue.sol
olegrok: Right now we don't have a special interface to send data to specific shard. Our smart accounts operates with raw calldata and we can extract destination shard only after decoding a message. But we don't have a SSZ decoder in Solidity. So here we have single queue for all messages. In future then we will pass destination shards/address explicitly it will be possible.

nil/contracts/solidity/system/MessageQueue.sol
```diff
@@ -0,0 +1,29 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.15;
+
+import "../lib/IMessageQueue.sol";
+
+struct Message {
+    bytes data;
+    address sender;
```
Gezort, line 8: it may be convenient to introduce more fields here (e.g. tokens)

nil/contracts/solidity/system/MessageQueue.sol
```diff
@@ -0,0 +1,29 @@
...
+}
+
+contract MessageQueue is IMessageQueue{
+    function sendRawTransaction(bytes calldata _message) external override {
```
idea404, line 12: What are thoughts around supporting attaching `value`?

nil/contracts/solidity/system/MessageQueue.sol
olegrok: > What are the limits on the size of the queue or message size in queue?

Right now there is no any limits but we can add them

> Also how would this compare to another approach such as emitting an event instead?

I didn't think about emitting events but this looks like an attempt to move all magic out of EVM. Sounds interesting but I'm not sure it's what we expect. Should be discussed. Thanks for idea.

nil/contracts/solidity/system/MessageQueue.sol
olegrok: In the first version I'd just want to check that it will work at all. basic test passed. We can discussed it and extends current implementation.

nil/internal/collate/proposer.go
```diff
@@ -165,6 +169,15 @@ func (p *proposer) fetchLastBlockHashes(tx db.RoTx) error {
 	return nil
 }
 
+func (p *proposer) handleMessageQueue(bn types.BlockNumber) error {
+	txn, err := execution.CreateMQPruneTransaction(p.params.ShardId, bn)
+	if err != nil {
+		return fmt.Errorf("failed to create MQ prune transaction: %w", err)
+	}
+	p.proposal.SpecialTxns = append(p.proposal.SpecialTxns, txn)
```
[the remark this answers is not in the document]
olegrok: If such TX fail we fail block validation on consensus and then new block in the next round will be suggested.

PR #790, opened 2025-04-10
Get unique columns name from all events before inserting them into table

Zerg1996:
## Short Summary

Previously, when processing a batch of logs, the system determined the list of column names based only on the event with the largest number of fields. As a result, if an event with fewer fields did not include certain columns that existed in the larger event, those columns would be missing during insertion, causing data to be lost in ClickHouse.

## What Changes Were Made

- Fetching all unique columns name from events
- Testing batch insert events

## Checklist


nil/tests/journald_forwarder/journald_forwarder_test.go
```diff
@@ -295,6 +295,55 @@ func (s *SuiteJournaldForwarder) TestBatchLogDataInsert() {
 	})
 }
 
+func (s *SuiteJournaldForwarder) TestInsertedValue() {
+	s.Run("Check insert", func() {
```
olegrok, line 299: Do you really need to wrap it into `s.Run()`?
It's a single subtest here you can drop that wrapper

nil/services/journald_forwarder/main.go
```diff
@@ -43,6 +43,18 @@ func NewLogServer(click *Clickhouse, logger logging.Logger) *LogServer {
 
 var fieldStoreClickhouseTyped = logging.FieldStoreToClickhouse + logging.GetAbbreviation("bool")
 
+func createUniqueColumnsEvent(events []Event) Event {
+	uniqueColumns := Event{}
+	for _, event := range events {
+		for key, value := range event {
+			if _, exists := uniqueColumns[key]; !exists {
+				uniqueColumns[key] = value
```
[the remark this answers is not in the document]
Zerg1996: I use it to get type of column. In case where I want to add missing columns, I also call this function.

nil/services/journald_forwarder/main.go
```diff
@@ -206,23 +211,19 @@ func (s *LogServer) Export(
...
 		s.logger.Error().Err(insertErr).Msgf("Error inserting data: %+v", eventsToStore)
 		return nil, err
 	}
 
-	diffNames := make([]string, 0, len(diffColumns))
-	diffTypes := make([]string, 0, len(diffColumns))
-	for key, value := range diffColumns {
+	diffNames := make([]string, 0, len(uniqueEvent))
+	diffTypes := make([]string, 0, len(uniqueEvent))
+	for key, value := range uniqueEvent {
```
knazarov, line 226: You don't need `value` here, just use `_`

knazarov: Oh, you do actually use the value. NVM

PR #793, opened 2025-04-10
[GH-776] Use local data for queries to other shards in readonly RawAPI methods

### LINKED ISSUE #776: Sporadic deployment error when archive node is running
Users complained that when they tried to deploy something through their smart account, they encountered the following kind of error:
```
failed to estimate call fee: failed to call method Call on shard 1: failed to call method Call on shard 2: failed to read block 0x000271b5d109c014407376dcad243fe56f45323f9c1bb038bc04ca04d679f51a: key not found in db
```

Disabling the archive node gets rid of the error. Apparently, the reason is that in the chain of `Call`-requests the second one sometimes goes from the validator to the archive node, which lags behind, so the block with the specified hash is not located on it.

Un1oR:
This PR resolves #776 in the following way: we avoid going to the lagging archive node by executing the entire chain of calls within a single node, since we have data from all shards available.

**N.B.** To better understand what is going on, it is recommended to familiarize yourself with the changes from commit to commit. Some changes are made iteratively across multiple commits and may not reflect the final state at every point, but each commit results in some valid self-sufficient state.

I was primarily addressing the problem of writing a test that stably reproduces the problem situation. It turned out to be not so easy to simulate the lagging archive node and the correct order of requests. For this purpose I needed preliminary refactoring with hiding a concrete implementation of `network.Manager` behind an interface.

Next, I tackled the challenge of providing different subsets of RawAPI more granularly via libp2p. To do this, I deeply reworked this subsystem, adding the ability to independently create and register an arbitrary set of RawAPIs.

At the end, I was finally able to orchestrate the instantiation of the right APIs in the right modes on the right shards, which resulted in changing the behavior in the previously written test to the desired behavior.


nil/tests/rpc_node/call_test.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+		return &validatorNetworkManager{
+			Manager:             manager,
+			getPeersForProtocol: getPeersForProtocol,
+		}, nil
+	}
+}
+
+func (s *SuiteRpcNodeCall) TestCall() {
+	defer s.Cancel()
```
olegrok, line 64: That should be done in `TearDownTest` function

nil/tests/rpc_node/call_test.go
```diff
@@ -0,0 +1,143 @@
...
+
+	s.Run("Prepare network", func() {
```
olegrok, line 66: This can be moved in a `SetupTest` function.

nil/tests/rpc_node/call_test.go
```diff
@@ -0,0 +1,143 @@
...
+	"github.com/stretchr/testify/suite"
+	"sync/atomic"
+	"testing"
+)
+
+type SuiteRpcNodeCall struct {
+	tests.ShardedSuite
+}
+
+type NetworkManagerFactory = func(ctx context.Context, cfg *nilservice.Config, db db.DB) (network.Manager, error)
```
olegrok, line 27: You can define this type in a previous patch right near config - https://github.com/NilFoundation/nil/pull/793/commits/14b46f14b1f5f0e9b072be923e3ba1e915b5987a#diff-3ce05a3487a74b5c8772e50abed56c59cc17d0e7b35b4a8925446661af9fe3a1R90

nil/services/nilservice/service.go
```diff
@@ -195,12 +195,16 @@ func getRawApi(
 	database db.DB,
```
olegrok, line 193: Commit message of this patch (Changing the way we specify that we should use the readonly API) should contain some description/motivation regarding this change

nil/services/nilservice/service.go
```diff
@@ -213,25 +212,32 @@ func getRawApi(
...
-				api, ok := shardApis[shardId].(*rawapi.LocalShardApi)
-				check.PanicIfNot(ok)
-				shardApis[shardId], err = rawapi.NewLocalRawApiAccessor(shardId, api)
+			err = nodeApiBuilder.WithLocalShardApiRo(shardId, database, txnPools[shardId], cfg.EnableDevApi)
+			if err != nil {
+				return nil, err
+			}
+			if !readonly {
+				err = nodeApiBuilder.WithLocalShardApiRw(shardId, database, txnPools[shardId], cfg.EnableDevApi)
+				if err != nil {
```
olegrok:
```
if err := f(); err != nil {
...
```

nil/services/nilservice/service.go
```diff
@@ -213,25 +212,32 @@ func getRawApi(
...
+	nodeApiBuilder := rawapi.NodeApiBuilder()
 	for shardId := range types.ShardId(cfg.NShards) {
 		var err error
 		if slices.Contains(myShards, uint(shardId)) {
-			shardApis[shardId] = rawapi.NewLocalShardApi(shardId, database, txnPools[shardId], readonly, cfg.EnableDevApi)
-			if assert.Enable {
-				api, ok := shardApis[shardId].(*rawapi.LocalShardApi)
-				check.PanicIfNot(ok)
-				shardApis[shardId], err = rawapi.NewLocalRawApiAccessor(shardId, api)
+			err = nodeApiBuilder.WithLocalShardApiRo(shardId, database, txnPools[shardId], cfg.EnableDevApi)
```
olegrok: `ShardApiRo` shouldn't accept txPools. Let's remove this argument or pass `nil` explicitly.

nil/services/rpc/rawapi/client_direct_emulator.go
```diff
@@ -0,0 +1,104 @@
...
+
+type shardApiRequestPerformerDirectEmulator[RawApiType shardApiBase] struct {
+	apiName       string
+	apiType       reflect.Type
+	transportType reflect.Type
+
+	shardApi RawApiType
+	codec    apiCodec
+
+	derived any
```
olegrok, line 51: Is it possible to make it typed or at least put a comment what is it?

nil/services/rpc/rawapi/shard_api.go
```diff
@@ -0,0 +1,76 @@
...
+	"github.com/NilFoundation/nil/nil/services/txnpool"
+)
+
+type shardApiBase interface {
+	shardId() types.ShardId
+	setNodeApi(nodeApi NodeApi)
+	setAsP2pRequestHandlersIfAllowed(ctx context.Context, networkManager network.Manager, logger logging.Logger) error
+}
+
+const apiNameRo = "rawapi_ro"
```
olegrok, line 22: Does it mean that we will call methods like `/shard/5/rawapi_ro/GetBlockHeader` instead of `/shard/5/rawapi/GetBlockHeader`?

nil/services/nilservice/service.go
```diff
@@ -12,7 +12,6 @@ import (
 	"time"
```
olegrok, line 11: This patch (RawAPI has been redesigned. Added the ability to operate any number of different ShardAPIs in any combination) also needs a good description.

nil/services/rpc/jsonrpc/txpool_api_test.go
```diff
@@ -47,7 +47,7 @@ func (suite *SuiteTxnPoolApi) SetupSuite() {
 	defer database.Close()
 
 	suite.api = rawapi.NodeApiBuilder(database, nil).
-		WithLocalShardApiRo(types.MainShardId, suite.pool).
+		WithLocalShardApiRw(types.MainShardId, suite.pool).
```
olegrok, line 50: That looks like a fixing a typo from previous patch

nil/services/rpc/rawapi/client_rw.go
```diff
@@ -46,6 +47,17 @@ func (api *shardApiClientRw) SendTransaction(ctx context.Context, transaction []
 		ctx, api, "SendTransaction", transaction)
 }
 
-func (api *shardApiClientRw) DoPanicOnShard(ctx context.Context) (uint64, error) {
-	return sendRequestAndGetResponseWithCallerMethodName[uint64](ctx, api, "DoPanicOnShard")
+func (api *shardApiClientRw) GetTransactionCount(
```
olegrok, line 50: The same

nil/tests/rpc_node/call_test.go
```diff
@@ -133,7 +133,7 @@ func (s *SuiteRpcNodeCall) TestCall() {
 		}
 
 		_, err = s.DefaultClient.EstimateFee(s.Context, callArgs, "latest")
-		s.ErrorContains(err, db.ErrKeyNotFound.Error())
+		s.NoError(err, db.ErrKeyNotFound.Error())
```
olegrok, line 136: Could you please describe how changes from this patch (Local readonly APIs are used for all shards on validators and archive nodes) fixed this issue?

olegrok (commented):
> This PR resolves https://github.com/NilFoundation/nil/issues/776 in the following way: we avoid going to the lagging archive node by executing the entire chain of calls within a single node, since we have data from all shards available.

I guess you can have the same issues for non-archive nodes. Validator must be in sync with shard it validates. But it can have one-two blocks lag for shards that it doesn't validate.

> N.B. To better understand what is going on, it is recommended to familiarize yourself with the changes from commit to commit. Some changes are made iteratively across multiple commits and may not reflect the final state at every point, but each commit results in some valid self-sufficient state.

I tried to do that but it's still hard. You perfectly split your patch into several commits. I like it. But you need to add at least small description to each commit.

nil/tests/rpc_node/call_test.go
Un1oR: I can't because of cyclic type dependency (`Config` <-> `NetworkManagerFactory`). I mean I can, but I can't refer to it inside `Config`.

nil/services/rpc/rawapi/shard_api.go
Un1oR: Yes

olegrok: Do we really need such inconsistency?

Un1oR: I think this is worth doing. First, since now the interfaces exported to libp2p are not linked, we might accidentally get an unwanted name collision. Second, we can imagine a situation where we might want to have a collision of method names in different interfaces. For example, the same `GetTransactionCount` could have a readonly version without `PendingBlock` support.

nil/services/rpc/jsonrpc/txpool_api_test.go
Un1oR: Actually, no. Moving the methods that work with txpool to the rw API is a separate sub-task. First I kept the status quo and then made these necessary changes.

nil/services/nilservice/service.go
Un1oR: Writing commit descriptions is what gave me the most trouble yesterday after rebase) I'll try to make the descriptions more detailed a bit later.

nil/services/nilservice/service.go
```diff
@@ -668,7 +669,11 @@ func Run(
 	return 0
 }
 
-func createNetworkManager(ctx context.Context, cfg *Config, database db.DB) (*network.BasicManager, error) {
+func CreateNetworkManager(ctx context.Context, cfg *Config, database db.DB) (network.Manager, error) {
+	if cfg.Network != nil && cfg.RunMode != NormalRunMode {
+		cfg.Network.DHTMode = dht.ModeClient
+	}
```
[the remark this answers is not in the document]
Un1oR: Frankly speaking, I have biiig questions about the cleanliness of the code in `CreateNode`, including the "magic" `if`s. Specifically here, there were already config changes inside the function, and I just put the ones that were outside next to them for consistency. I had a bad feeling implementing the customizability of the `networkManager` creation. It all seems rather crusty. I was even pondering whether we should switch to using a DI framework (like `fx`, as libp2p does in its implementation) to construct our "application" object. But this has its disadvantages. We've [faced a similar problem before](https://www.youtube.com/watch?v=QGWwMMylyKM&t=1036s), and based on that experience, I think that perhaps converting one complex factory method into a more granular builder could work, but I don't have an understanding of exactly how to do that well right away.

nil/tests/rpc_node/call_test.go
Un1oR: By calling `api.nodeApi.Call` inside `handleOutTransactions` we stop going off the network, hitting a potentially lagging archive node, and instead process the entire request chain locally. This works well. I need to better deal with what you write above about possible lag of shards that are not validated by the validator, but in fact now we don't encounter problems when the whole chain is processed inside the validator (albeit with calls through the network) in the absence of an archive node.

nil/services/rpc/rawapi/client_direct_emulator.go
Un1oR: I didn't manage to describe it using the limitations of Go generics right away. I was very happy to see that this field is passed to a function that just wants `any`. But I'll see what I can do, including a comment.

olegrok (approved):

PR #796, opened 2025-04-11
Drop await feature

olegrok:
We can't support it in a pure EVM. So it was decided to remove this feature.


olegrok:
> I'm not sure that we won't be able to implement Request/Response feature. Except the requests ’ chain (which we can drop), it doesn't require anything that cannot be implemented with pure EVM. So I'd propose keeping its support in interfaces and examples.

Could you please describe how should it work?
Right now we dump current EVM state and there is no such ability with pure EVM.
Am I missed something?

shermike:
It is not about `awaitCall` which indeed cannot be implemented. But `sendRequest` is doable, there already were several approaches to do that.
Here is from Ilya Marozau https://www.notion.so/nilfoundation/Resp-req-message-proposal-6fa01099150144b38c376f6da57f5128
<details>
  <summary>My approach</summary>

    contract NilAwaitable is NilBase {
        Control internal ctrl;

        struct Awaiter {
            function(uint256, bytes memory, bytes memory) internal callback;
            uint answer_id;
            bool active;
            bytes context;
        }

        struct Control {
            mapping(uint256=>Awaiter) awaiters;
            uint256 await_id;
        }

        function awaitCall(
            address dst,
            uint gas,
            uint value,
            bytes memory callData,
            bytes memory context,
            function(uint256, bytes memory, bytes memory) internal cb
        ) internal returns(bool) {
            uint256 await_id = ctrl.await_id;
            ctrl.awaiters[await_id] = Awaiter({callback: cb, answer_id: ctrl.await_id, active: true, context: context});
            ctrl.await_id += 1;

            bool success = Nil.asyncCall(dst, address(this), value, callData);

            require(success, "Precompiled contract call failed");

            return true;
        }

        function onFallback() internal {
            (uint32 func_id, uint256 answer_id, bytes memory data) = abi.decode(msg.data, (uint32, uint256, bytes));
            Awaiter storage awaiter = ctrl.awaiters[answer_id];
            require(awaiter.active);
            function(uint256, bytes memory, bytes memory) internal cb = awaiter.callback;
            delete ctrl.awaiters[answer_id];
            cb(answer_id, data, awaiter.context);
        }
    }
</details>

Also, Everscale did that through regular TON contracts storage.

olegrok:
> It is not about `awaitCall` which indeed cannot be implemented. But `sendRequest` is doable, there already were several approaches to do that. Here is from Ilya Marozau https://www.notion.so/nilfoundation/Resp-req-message-proposal-6fa01099150144b38c376f6da57f5128
> 

Well, let's postpone (or close) this PR then.

Initially, my goal is to reduce amount of ways how we create outbound messages.
Ideally we need to keep only one way - via asyncCalls. Once we should change an approach with precompiles completely to pure EVM. Maintaining of request-response model make this task a bit more messy.

explorer_frontend/src/features/tutorial/spec.ts
```diff
@@ -1,8 +1,5 @@
 import { TutorialLevel } from "./model";
 import {
-  tutorialFiveContracts,
-  tutorialFiveIcon,
-  tutorialFiveText,
```
dmtrskv, line 5: maybe tutorials deserve more independent names hinting at their content? @khannanov-nil

olegrok:
Please ignore this patch. I'll close it. Since we agreed to keep syncCall features. Only "await" part should be removed.

PR #798, opened 2025-04-11
feat(sc): handle CommitBatch errors, add simulation before tx submission

x-mass:
## Short Summary

Sync committee. Handle CommitBatch errors, add simulation before tx submission.

## What Changes Were Made
- Add simulation before submitting each transaction to L1 to catch errors early.
- If simulation fails, parse the RPC error; depending on its type, the `CommitBatch` method either panics or triggers a state resync from L1 (e.g., when multiple SC instances are running and local state is stale).
- Move L1 finalized state fetcher to a separate class for better modularity.
- Pass the fetcher into both aggregator and proposer components so they can detect outdated local state and initiate L1 synchronization.

## Checklist

### DIFF
--- a/nil/services/synccommittee/core/fetching/aggregator.go
+++ b/nil/services/synccommittee/core/fetching/aggregator.go
@@ -39,6 +39,7 @@ type AggregatorBlockStorage interface {
 	TryGetLatestBatchId(ctx context.Context) (*types.BatchId, error)
 	SetBlockBatch(ctx context.Context, batch *types.BlockBatch) error
 	GetFreeSpaceBatchCount(ctx context.Context) (uint32, error)
+	SetProvedStateRoot(ctx context.Context, stateRoot common.Hash) error
 }
 
 type AggregatorConfig struct {
@@ -65,7 +66,7 @@ type aggregator struct {
 	batchEncoder    encode.BatchEncoder
 	blobBuilder     blob.Builder
 	rollupContract  rollupcontract.Wrapper
-	resetter        *reset.StateResetter
+	resetter        *reset.StateResetLauncher
 	clock           clockwork.Clock
 	metrics         AggregatorMetrics
 	config          AggregatorConfig
@@ -77,7 +78,7 @@ func NewAggregator(
 	rpcClient RpcBlockFetcher,
 	blockStorage AggregatorBlockStorage,
 	taskStorage AggregatorTaskStorage,
-	resetter *reset.StateResetter,
+	resetter *reset.StateResetLauncher,
 	rollupContractWrapper rollupcontract.Wrapper,
 	clock clockwork.Clock,
 	logger logging.Logger,
@@ -110,7 +111,20 @@ func (agg *aggregator) Name() string {
 func (agg *aggregator) Run(ctx context.Context, started chan<- struct{}) error {
 	agg.logger.Info().Msg("Starting block fetching")
 
-	err := agg.workerAction.Run(ctx, started)
+	latestStateRoot, err := agg.blockStorage.TryGetProvedStateRoot(ctx)
+	if err != nil {
+		return err
+	}
+	if latestStateRoot == nil {
+		// root not initialized in current database
+		agg.logger.Warn().
+			Msg("L1 state root is not initialized, trying to use storage state root")
+		if err := agg.syncLatestFinalizedRoot(ctx); err != nil {
+			return fmt.Errorf("failed to set proved state root: %w", err)
+		}
+	}
+
+	err = agg.workerAction.Run(ctx, started)
 
 	if err == nil || errors.Is(err, context.Canceled) {
 		agg.logger.Info().Msg("Block fetching stopped")
@@ -168,13 +182,16 @@ func (agg *aggregator) handleProcessingErr(ctx context.Context, err error) error
 
 	case errors.Is(err, types.ErrBlockMismatch):
 		agg.logger.Warn().Err(err).Msg("Block mismatch detected, resetting state")
-		if err := agg.resetter.ResetProgressNotProved(ctx); err != nil {
+		if err := agg.resetter.LaunchResetToL1WithSuspension(ctx, agg); err != nil {
 			return fmt.Errorf("error resetting state: %w", err)
 		}
 		return nil
 
 	case errors.Is(err, storage.ErrStateRootNotInitialized):
-		agg.logger.Warn().Err(err).Msg("State root not initialized, skipping")
+		agg.logger.Warn().Err(err).Msg("State root not initialized, trying to fetch it from L1")
+		if err := agg.syncLatestFinalizedRoot(ctx); err != nil {
+			return fmt.Errorf("failed to set proved state root: %w", err)
+		}
 		return nil
 
 	case errors.Is(err, storage.ErrCapacityLimitReached):
@@ -372,7 +389,7 @@ func (agg *aggregator) handleBlockBatch(ctx context.Context, batch *types.BlockB
 	}
 
 	if err := agg.rollupContract.CommitBatch(ctx, sidecar, batch.Id.String()); err != nil {
-		return fmt.Errorf("error committing batch, latestMainHash=%s: %w", batch.LatestMainBlock().Hash, err)
+		return agg.handleCommitBatchError(ctx, batch, err)
 	}
 
 	if err := agg.createProofTasks(ctx, batch); err != nil {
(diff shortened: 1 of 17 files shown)


nil/services/synccommittee/core/proposer.go
```diff
@@ -147,8 +112,8 @@ func (p *proposer) updateStateIfReady(ctx context.Context) error {
 		}
 
 		// another actor has already sent an update for this batch, we need to refetch state from contract
-		p.logger.Warn().Msg("batch is already finalized, skipping UpdateState tx")
-		if err := p.updateStoredStateRootFromContract(ctx); err != nil {
+		p.logger.Warn().Msg("batch is already finalized, skipping UpdateState tx, syncing state with L1")
```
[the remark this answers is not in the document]
x-mass: It will have no effect on the proposer or any other component. If the state root wasn't initialized previously, initializing it without creating batches won't generate any new proposal data.
Currently, this is handled by the aggregator, which both sets the state root and creates a new batch.

nil/services/synccommittee/core/fetching/aggregator.go
```diff
@@ -65,6 +66,7 @@ type aggregator struct {
 	batchEncoder    encode.BatchEncoder
 	blobBuilder     blob.Builder
 	rollupContract  rollupcontract.Wrapper
+	l1Syncer        *l1statesyncer.L1StateSyncer
```
oclaw, line 69: rename suggestion: `package l1statesyncer -> package l1state`, `L1StateSyncer -> Syncer`, will be same informative and much shorter :)

nil/services/synccommittee/core/fetching/aggregator.go
```diff
@@ -372,6 +376,16 @@ func (agg *aggregator) handleBlockBatch(ctx context.Context, batch *types.BlockB
 	}
 
 	if err := agg.rollupContract.CommitBatch(ctx, sidecar, batch.Id.String()); err != nil {
+		if errors.Is(err, rollupcontract.ErrBatchAlreadyCommitted) ||
+			errors.Is(err, rollupcontract.ErrBatchAlreadyFinalized) {
+			// for some reason, we attempted to prove a batch that has already been proved,
+			// sync the latest proved root with the L1 contract.
+			agg.logger.Warn().Msg("batch is already committed, syncing state with L1")
```
oclaw, line 383: Let's add batch id to the log

nil/services/synccommittee/internal/l1statesyncer/l1_state_syncer_test.go
```diff
@@ -0,0 +1,129 @@
...
+	l1Syncer         *L1StateSyncer
+	callContractMock *testaide.CallContractMock
+}
+
+func TestL1SyncerSuite(t *testing.T) {
+	t.Parallel()
+	suite.Run(t, new(L1SyncerTestSuite))
+}
+
+func (s *L1SyncerTestSuite) SetupSuite() {
```
oclaw, line 44: Why not SetupTest()? Do you need to keep db state between tests for some reason? (I see that we only have one now, but anyway)

nil/services/synccommittee/internal/rollupcontract/wrapper.go
```diff
@@ -207,6 +220,69 @@ func (r *wrapperImpl) logReceiptDetails(receipt *ethtypes.Receipt) {
...
+	args := make(map[string]any)
+	err = errorMethod.Inputs.UnpackIntoMap(args, revertData[4:])
+	if err != nil {
+		return fmt.Errorf("args upack error: %w", err)
+	}
+
+	return contractError{errorMethod.Name, args}
+}
+
+func (r *wrapperImpl) getAddress() (ethcommon.Address, error) {
```
oclaw, line 258: Do we need to do this case and transform each time? Maybe precompute it in constructor and keep as a field?

nil/services/synccommittee/internal/rollupcontract/commit_batch.go
```diff
@@ -73,14 +76,40 @@ func (r *wrapperImpl) CommitBatch(
...
 	if receipt.Status != ethtypes.ReceiptStatusSuccessful {
-		return errors.New("CommitBatch tx failed")
+		// Re-simulate the transaction on top of the block it originally failed in.
+		// Note: The execution order of transactions in the block is not preserved during simulation,
+		// so results may differ — but we attempt to identify the cause of failure anyway.
+		err = r.simulateTx(ctx, signedTx, receipt.BlockNumber)
+		if err != nil {
+			return r.handleCommitBatchTxError(fmt.Errorf("post-submition simulation: %w", err), batchIndex)
+		}
+		return errors.New("CommitBatch tx failed, can't establish the reason")
```
oclaw, line 89: `identify the reason` describes the situation better I believe

nil/services/synccommittee/internal/rollupcontract/commit_batch.go
```diff
@@ -73,14 +76,40 @@ func (r *wrapperImpl) CommitBatch(
...
+		// abigen doesn't generate error types, have to specify them manually
+		switch cerr.MethodName {
+		case "ErrorBatchAlreadyFinalized":
+			return ErrBatchAlreadyFinalized
+		case "ErrorBatchAlreadyCommitted":
+			return ErrBatchAlreadyCommitted
+		case "ErrorInvalidBatchIndex":
+			panic("invalid batch index encountered: " + batchIndex)
+		case "ErrorInvalidVersionedHash":
+			panic("invalid versioned hash was computed: " + batchIndex)
```
oclaw, line 107: Do we really want to panic here?

nil/services/synccommittee/internal/l1statesyncer/l1_state_syncer_test.go
x-mass: We don't need to keep db state between tests, and it is reset inside `SetupTest`.

nil/services/synccommittee/internal/rollupcontract/commit_batch.go
x-mass: If we received the wrong batch index in storage, it likely indicates that something serious has gone wrong in a way it wasn't supposed to.

nil/services/synccommittee/core/proposer.go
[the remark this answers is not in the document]
akokoshn: ok, if aggregator responsible for initializing `latestProved` in DB, make sense call sync when aggregator start - begin of the `Run` method (without waiting first error)

nil/services/synccommittee/internal/rollupcontract/commit_batch.go
akokoshn: `ErrorInvalidVersionedHash` and `ErrorInvalidBatchIndex` means wrong data for `commitBatch` (blobs, or batch index format).
In this case we should stop SC, reset state root to the latest proved, reset all batches, fix and run again.
On the dev environment make sense just reset all batches and rise specific log/metric event for alert `CorruptedBatchData`

x-mass: Added the same handling as for `ErrBatchAlreadyCommitted` and `ErrBatchAlreadyFinalized`, which is resetting the state to L1 latest finalized root.

nil/services/synccommittee/core/proposer.go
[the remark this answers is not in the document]
x-mass: Added to aggregator, got rid of L1 syncer at all. Reset logic is inside `StateResetLauncher` now.

nil/services/synccommittee/core/reset/state_reset_launcher.go
```diff
@@ -64,41 +79,151 @@ func (l *stateResetLauncher) LaunchPartialResetWithSuspension(
...
 	case <-time.After(gracefulShutdownTimeout):
-		l.logger.Fatal().Err(err).Msgf("Service did not stop after %s, force termination", gracefulShutdownTimeout)
+		l.logger.Fatal().Err(resumeErr).
+			Msgf("Service did not stop after %s, force termination", gracefulShutdownTimeout)
 	case <-stopped:
 	}
 }
+
+func (l *StateResetLauncher) withExclusiveLock(fn func() error) error {
+	l.mu.Lock()
```
akokoshn, line 213: may be make `isRunning` atomic?

nil/services/synccommittee/core/fetching/aggregator.go
```diff
@@ -168,13 +173,16 @@ func (agg *aggregator) handleProcessingErr(ctx context.Context, err error) error
 
 	case errors.Is(err, types.ErrBlockMismatch):
 		agg.logger.Warn().Err(err).Msg("Block mismatch detected, resetting state")
-		if err := agg.resetter.ResetProgressNotProved(ctx); err != nil {
+		if err := agg.resetter.LaunchNotProvedResetWithSuspension(ctx, agg); err != nil {
```
zadykian, line 176: I'm still sure that we only need two kinds of reset:

1. Partial reset triggered when the proof of a specific batch fails – `LaunchPartialResetWithSuspension`
2. Full reset, which includes discarding all batches and refetching the state root – `LaunchResetToL1WithSuspension`

Currently, the "Not Proved" reset is only triggered by the `Aggregator` when `ErrBlockMismatch` is detected, which is not entirely correct.

Here is an example where this would break:

-> Aggregator creates batch `A`, which contains a single block from shard `X`  
-> ProofProvider receives a new task and launches proof generation  
-> The batch is proved, but L1 becomes unavailable for some time, so batch `A` remains in a "pending" state  
-> Someone triggers a reset or rollback on the L2 side  
-> Aggregator tries to construct batch `A+1`, but fails due to a parent-child block hashes mismatch  
-> It handles `ErrBlockMismatch` and launches the "Not Proved" reset, but batch `A` – which contains a block that is no longer valid – has already been proved  
-> Aggregator becomes fully stuck

The key point here is that the call to `LaunchNotProvedResetWithSuspension` should be replaced. I see two options:

1. Traverse the full sequence of created batches, find the latest valid one, and reset all the following batches  
2. Trigger a full reset, which is already implemented and used in multiple places

The second option is the simpler of the two. I'm suggesting we stick to it.  
By doing so, we may discard some valid batches, but they can easily be reconstructed after the reset

nil/services/synccommittee/core/fetching/aggregator.go
```diff
@@ -110,6 +111,10 @@ func (agg *aggregator) Name() string {
 func (agg *aggregator) Run(ctx context.Context, started chan<- struct{}) error {
 	agg.logger.Info().Msg("Starting block fetching")
 
+	if err := agg.syncLatestFinalizedRoot(ctx); err != nil {
```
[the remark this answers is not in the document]
zadykian: Proceeding with an inaccessible L1 can potentially lead to an arbitrary number of invalid batches being created and proven

In that case, the stored state root — which may be outdated at this point — will be used without synchronization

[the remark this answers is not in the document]
akokoshn: In this case we should sync with L1 on handling `BlockMissMatchErr`
If L1 becomes accessible during processing - get error on `CommitBatch` or `UpdateState`, then sync and reset storage

nil/services/synccommittee/core/fetching/aggregator.go
```diff
@@ -418,3 +453,39 @@ func (agg *aggregator) prepareForBatchCommit(
 
 	return agg.rollupContract.PrepareBlobs(ctx, blobs)
 }
+
+func (agg *aggregator) getLatestFinalizedRoot(ctx context.Context) (common.Hash, error) {
```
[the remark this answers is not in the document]
akokoshn: as far i see method returns `agg.rollupContract.LatestFinalizedStateRoot` or `genesis`

nil/services/synccommittee/core/fetching/aggregator.go
x-mass: Replaced ResetNotProved with ResetAll

akokoshn (commented):
Looks ok from my point of view.
Let's wait for @zadykian

PR #801, opened 2025-04-13
[evm] Support debug logging in contracts

shermike:
Add the ability to log in contracts: `console.log(fmt, args...)`.

To do this, a new precompile was created, which takes the formatting string as the first parameter, and the format arguments as the rest.

The maximum number of arguments is 4. This restriction is necessary because a separate Solidity function is generated in the console library for all permutations of argument types.
Only three types of arguments are supported in `console.log()`: `uint256`, `address`, `string`.

The script `genlog.py` generates two files: `console.sol`, which contains all log functions to be used in Solidity contracts, and `contracts_generated.go`, which contains a map from function selectors to argument types for each log function defined in `console.sol`.

Format string is straightforward, there are only two specifiers: `%_` - suitable for all types and `%x` - print integer as a hex. E.g. `console.log("test args: int=%x, address=%_, string=%_", 123, addr, "some");` will print:
```
INF [solidity]	 nil/internal/vm/precompiled.go:1215 > test args: int=0x7b, address=0xabcdef, string=some
```


.github/workflows/main.yml
```diff
@@ -102,6 +102,9 @@ jobs:
           brew install go solidity protobuf coreutils
           go install golang.org/x/tools/cmd/stringer@latest
           go install google.golang.org/protobuf/cmd/protoc-gen-go@latest
+#          pip3 install pycryptodome --break-system-packages
+#          python3 -c "from Crypto.Hash import keccak"
+          pip install safe-pysha3 --break-system-packages
```
Un1oR, line 107:
```suggestion
          python3 -m pip install safe-pysha3
```

nil/contracts/generate.go
```diff
@@ -2,6 +2,8 @@ package contracts
 
 import "embed"
 
+//go:generate python3 genlog.py ./solidity/lib/console.sol ../internal/vm/console/console_generated.go
```
olegrok, line 5: It's better to move it directly to Makefile to avoid rebuild on each `make` invocation.

nil/internal/vm/console/console.go
```diff
@@ -0,0 +1,91 @@
...
+
+	params, ok := LogSignatures[funcId]
+	if !ok {
+		return "", fmt.Errorf("unknown log signature: %x", funcId)
+	}
+	if params[0] != StringTy {
+		return "", errors.New("first parameter must be a format string")
+	}
+	format := readString(input, 0)
+	res := ""
```
olegrok, line 28: You can use `strings.Builder` here (https://pkg.go.dev/strings#Builder)

olegrok (approved):

nil/internal/vm/console/console.go
shermike: Indeed. Thanks.

PR #804, opened 2025-04-14
[vm] drop sendRawTransaction precompile

olegrok:
We plan to move messaging to smart contracts. So inside smart contracts it's
impossible to operate with SSZ. So we need to drop sendTransaction precompile.


shermike (approved):

nil/internal/contracts/testaide.go
```diff
@@ -114,24 +114,11 @@ func GetCounterValue(t *testing.T, data []byte) int32 {
 	return val
 }
 
-func NewSmartAccountSendCallData(t *testing.T,
-	bytecode types.Code, gasLimit types.Gas, value types.Value,
-	tokens []types.TokenBalance, contractAddress types.Address, kind types.TransactionKind,
+func NewSmartAccountSendCallData(t *testing.T, bytecode types.Code,
```
Un1oR, line 117:
```suggestion
func NewSmartAccountAsyncCallCallData(t *testing.T, bytecode types.Code,
```

PR #805, opened 2025-04-14
[PoC] Pure tokens sync transfer

Zerg1996:
Part of #792


smart-contracts/contracts/Relayer.sol
```diff
@@ -46,4 +46,14 @@ contract Relayer {
         (bool success, bytes memory data) = dst.call{value: value}(callData);
         require(success, "Call failed");
     }
+
+    function transfer(
```
shermike, line 50: I think it is better to move this method to the `TokenManager`.

Issue #742, opened 2025-04-03
Fallback for Faucet and Cometa to regular RPC API endpoints in clijs


Un1oR:
The CLI on Go does not oblige us to set all three endpoints. If `faucet_endpoint` or `cometa_endpoint` are not set in the config, then `rpc_endpoint` is [used as them](https://github.com/NilFoundation/nil/blob/2897ce0e8dfb503caddbf9035f1c28d3b6d51849/nil/cmd/nil/common/rpc_client.go#L29-L41). This is [not the case](https://github.com/NilFoundation/nil/blob/2897ce0e8dfb503caddbf9035f1c28d3b6d51849/clijs/src/base.ts#L106-L121) in `clijs`. We need to support falllback logic in `clijs` as well.

PR #806, opened 2025-04-15
[GH-742] [cli.js] fallback for cometa and faucet endpoints in config

Gezort:
this fixes problem described in #742

### DIFF
--- a/clijs/src/common/config.ts
+++ b/clijs/src/common/config.ts
@@ -162,10 +162,10 @@ class ConfigManager {
     }
   }
 
-  public getConfigValue(section: string, key: string): string {
+  public getConfigValue(section: string, key: string, fallback?: string): string | undefined {
     const config = this.loadConfig();
     const sectionConfig = config[section] as Record<string, string>;
-    return sectionConfig?.[key];
+    return sectionConfig?.[key] ?? fallback;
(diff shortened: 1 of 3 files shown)


clijs/src/common/config.ts
```diff
@@ -194,6 +195,16 @@ class ConfigManager {
...
     this.saveConfig(config);
   }
+
+  private fillFallbacks() {
+    const rpc = this.getConfigValue(ConfigKeys.NilSection, ConfigKeys.RpcEndpoint);
+    if (this.getConfigValue(ConfigKeys.NilSection, ConfigKeys.CometaEndpoint) === undefined) {
+      this.updateConfig(ConfigKeys.NilSection, ConfigKeys.CometaEndpoint, rpc);
+    }
+    if (this.getConfigValue(ConfigKeys.NilSection, ConfigKeys.FaucetEndpoint) === undefined) {
+      this.updateConfig(ConfigKeys.NilSection, ConfigKeys.FaucetEndpoint, rpc);
```
Un1oR, line 205: I don't think we should multiply explicit values in the config if we can do without setting them explicitly. We only add unnecessary work for users.

olegrok (approved):

PR #808, opened 2025-04-15
Sync Committee: Full Local Reset

zadykian:
### Sync Committee: Full Local Reset

`ErrBlockMismatch` error now triggers a full reset of the local state—discarding all batches, including those that have already been proved.

This may be necessary if an L2 rollback/reset occurs after a batch is proved, but before the corresponding state update is submitted to L1.

In such cases, the proved batch may contain mismatched block(s).

### DIFF
--- a/nil/services/synccommittee/core/fetching/aggregator.go
+++ b/nil/services/synccommittee/core/fetching/aggregator.go
@@ -168,7 +168,7 @@ func (agg *aggregator) handleProcessingErr(ctx context.Context, err error) error
 
 	case errors.Is(err, types.ErrBlockMismatch):
 		agg.logger.Warn().Err(err).Msg("Block mismatch detected, resetting state")
-		if err := agg.resetter.ResetProgressNotProved(ctx); err != nil {
+		if err := agg.resetter.ResetProgressFull(ctx); err != nil {
 			return fmt.Errorf("error resetting state: %w", err)
 		}
 		return nil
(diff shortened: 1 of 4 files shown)


nil/services/synccommittee/core/fetching/aggregator.go
```diff
@@ -168,7 +168,7 @@ func (agg *aggregator) handleProcessingErr(ctx context.Context, err error) error
 
 	case errors.Is(err, types.ErrBlockMismatch):
 		agg.logger.Warn().Err(err).Msg("Block mismatch detected, resetting state")
-		if err := agg.resetter.ResetProgressNotProved(ctx); err != nil {
+		if err := agg.resetter.ResetProgressFull(ctx); err != nil {
```
akokoshn, line 171: We still need 2 kind of reset:
1. Not proved - if some thing wrong happens in collection batch, proof generation/verification or on `nild` after latest proved point.
2. Full - state of `nild` become inconsistent early latest proved point
The second option leads reset of L1 contract.

If we found inconsistency, need check if the latest proved state root valid on `nild`. If so - reset not proved otherwise - full reset

akokoshn (changes requested):

nil/services/synccommittee/core/fetching/aggregator.go
zadykian: Let's move our discussion to #798, closing this PR for now

PR #809, opened 2025-04-15
[proposer] fix TxIdGap error on L1 contract tx

olegrok:
This patch fixes TxIdGap error that was caused by internal system messages from/to main shard. Here we assign manually that id for system transactions.

### DIFF
--- a/nil/internal/collate/proposer.go
+++ b/nil/internal/collate/proposer.go
@@ -191,7 +191,9 @@ func (p *proposer) handleL1Attributes(tx db.RoTx, mainShardHash common.Hash) err
 		}
 	}
 
-	txn, err := CreateL1BlockUpdateTransaction(block)
+	txId := p.executionState.InTxCounts[types.MainShardId]
+	p.executionState.InTxCounts[types.MainShardId] = txId + 1
+	txn, err := CreateL1BlockUpdateTransaction(block, txId)
 	if err != nil {
 		return fmt.Errorf("failed to create L1 block update transaction: %w", err)
 	}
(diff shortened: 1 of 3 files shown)


nil/internal/collate/proposer.go
```diff
@@ -191,7 +191,9 @@ func (p *proposer) handleL1Attributes(tx db.RoTx, mainShardHash common.Hash) err
 		}
 	}
 
-	txn, err := CreateL1BlockUpdateTransaction(block)
+	txId := p.executionState.InTxCounts[types.MainShardId]
+	p.executionState.InTxCounts[types.MainShardId]++
```
avm, line 195: This won't work if there isn't a record for MainShard in InTxCounts. You need to write `...InTxCounts[types.MainShardId] = txId + 1`

olegrok: Fixed

nil/internal/collate/proposer.go
```diff
@@ -191,7 +191,9 @@ func (p *proposer) handleL1Attributes(tx db.RoTx, mainShardHash common.Hash) err
 		}
 	}
 
-	txn, err := CreateL1BlockUpdateTransaction(block)
+	txId := p.executionState.InTxCounts[types.MainShardId]
+	p.executionState.InTxCounts[types.MainShardId] = txId + 1
```
knazarov, line 195: Wouldn't we increment the InTxCounts twice? Once here, and once in AcceptInternalTransaction?

olegrok: No. Proposer and validator create different execution states

knazarov: Shouldn't it be `txId := p.executionState.InTxCounts[types.MainShardId]` instead?

olegrok: Right now it will work. But in case if we will have more system contracts in future we should increment id. Otherwise all transaction will have the same TxId. That will produce TxGap error

avm (approved):

nil/internal/collate/proposer.go
knazarov: I guess we'll migrate to contract-based messaging before that happens (hopefully)

PR #811, opened 2025-04-15
[vm] drop awaitCall feature & support sendRequest  in eth_call

olegrok:
1)

This feature was experimental and it can be reimplemented in pure EVM.
So we decided to drop it. This patch does that.

In docs we replaced awaitCall with sendRequest. It was too hard to do that
in "sharded-voting-protocol". It requires significant rework on this contracts
and they also seems to be untested. Here we just removed this example.

2)

Support sendRequest. This feature was missed in eth_call/eth_estimateFee. This patch adds it.


khannanov-nil (changes requested):
At the moment, `awaitCall()` is not fully removed from docs by this PR. The docs will be left in a state where sections discussing `awaitCall()` have code snippets without `awaitCall()`. I will remove these sections as part of this PR tomorrow.

khannanov-nil (approved):
LGTM but please rename the `Awaiter.sol` and `AwaiterCounter.sol` contracts since they no longer await anything. They can be removed too, as well as their associated test.

Issue #786, opened 2025-04-09
Easy node startup


dmtrskv:
One should be able to start a node providing only the chain, like this:
`nild archive --chain=testnet`

It requires at least some data (for each of the chains) hard-coded:
- Genesis block: version or full zero-state.
- Relay addresses.
- Archive node identities (reachable via relays).

We should also ensure that corresponding parameters don't change unexpectedly when running `nild gen-configs` (with an opt-out).

PR #813, opened 2025-04-15
[nild] Argument --chain adds test/main-net defaults

dmtrskv:
## Short Summary

An archive node outside of the cluster will be run simply with `nild archive --chain devnet`.

*The PR is a draft until the relay is up.*

## What Changes Were Made

- Argument `--chain` in `nild` and default config depending on it.
- Defaults for the actual network: relay and bootstrap addresses. *WIP*: now on test cluster `--chain test`.
- Validation of cluster spec against the defaults in `nild gen-configs` with `--chain`. E.g., each bootstrap identity and each relay from the defaults must be present in the cluster configs.

## Related Issue

Fixes https://github.com/NilFoundation/nil/issues/786

## Future Tasks

More defaults, including zero state.

### DIFF
--- a/nil/services/nilservice/defaults.go
+++ b/nil/services/nilservice/defaults.go
@@ -1,0 +1,83 @@
+package nilservice
+
+import (
+	"fmt"
+
+	"github.com/NilFoundation/nil/nil/common/check"
+	"github.com/NilFoundation/nil/nil/internal/network"
+)
+
+const (
+	ChainNameDevnet = "devnet"
+
+	ChainNameTest = "test" // todo: remove after tests
+)
+
+type ChainDefaults struct {
+	RelayAddresses     []string
+	BootstrapAddresses []string
+
+	Relays         network.AddrInfoSlice
+	BootstrapPeers network.AddrInfoSlice
+}
+
+var (
+	TestDefaults = newChainDefaults(
+		[]string{
+			"/ip4/100.100.35.125/tcp/3149/p2p/16Uiu2HAmFhv4FVhp3syxJ7uQCXm86Rbiu9Y6Cox7zBD757mBNNdK",
+		},
+		[]string{
+			"/ip4/100.100.35.125/tcp/3149/p2p/16Uiu2HAmFhv4FVhp3syxJ7uQCXm86Rbiu9Y6Cox7zBD757mBNNdK/p2p-circuit/" +
+				"p2p/16Uiu2HAmErDZcwxP6KQ6x1oPn8bY1Kkg9o39LvACcKjzJNTYpDNd",
+		},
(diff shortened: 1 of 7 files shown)


nil/services/nilservice/defaults.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/NilFoundation/nil/nil/common/check"
+	"github.com/NilFoundation/nil/nil/internal/network"
+)
+
+const (
+	ChainNameTestnet = "testnet"
+	ChainNameMainnet = "mainnet"
+)
+
+type ChainDefaults struct {
```
Un1oR, line 15: I don't really understand how we're going to maintain this. As far as I understand it, in `devnet` (where is it by the way? We don't have `testnet` or `mainnet` right now, and they are the only ones represented here) we drop the configuration from time to time and validator identities can change. In that sense, I don't think a static hardcode is going to help us. Plus there's a chance of getting a chicken and egg problem when we're preparing a release in which we plan to deploy a new configuration, and we don't yet know the identifiers that we want to zip into the `nild` version we'll be releasing.

Why don't we put a method in the JSON-RPC API that will output actual validator addresses that can be used as `bootstrapPeers`? Then we could make a command that takes an RPC URL, goes to it, and gets the actual values. This would work for any deployments.

dmtrskv: We'll have to hard-code the url then. And also somehow guarantee that the service was not compromised.
libp2p guarantees identity verification and also allows to connect to the network without any mediators like RPC.
Here we have just a short list of peers that will be used by the bootstrap. I reduced it to the single archive node in a follow-up push (I don't think that we want to open validator nodes right away).

We won't use this on dev, I suppose.
And on prod we're not supposed to clear network keys a lot. (I doubt that we clear them regularly on dev either.)

I'm using nil-dev for tests now.
I'll leave `devnet` (or `testnet`?) probably in the end if that's how our chain is known to the public.

Un1oR: First, let's synchronize in terminology. As I understand it,
`mainnet` — not present
`testnet` — not present
`devnet` == `nil-prod`
🤷‍♂️ == `nil-dev`
Do you understand these terms in the same way?

How justified is it to worry about JSON-RPC being compromised? We have HTTPS there. If we assume the user will download `nild` from our site and trust it, why shouldn't they trust another HTTPS request?

dmtrskv: I've updated the chain names.

Concerning the RPC, let's discuss on Thursday.
Also, we can always change this solution to RPC later.

dmtrskv:
Will be implemented in a different fashion

PR #820, opened 2025-04-16
drop poseidon hash usage

olegrok:
At first it's quite slow.
At second we won't need to use our own precompiles.


Gezort:
we need to bump versions of our js packages as well

shermike (approved):

KlonD90 (changes requested):
Should use https://www.npmjs.com/package/@noble/hashes also some random package updates are not welcome too

KlonD90 (approved):
lgtm

PR #834, opened 2025-04-17
merge sendRequest precompile into asyncCall

olegrok:
Here we merge sendRequest precompile code into asyncCall precompile.
At first it allows to eliminate code duplication and then it's the first step before implementation Awaiter contract that will include some request/response machinery for response processing.

### DIFF
--- a/nil/contracts/solidity/tests/TokensTest.sol
+++ b/nil/contracts/solidity/tests/TokensTest.sol
@@ -70,7 +70,9 @@ contract TokensTest is NilTokenBase {
             address(this),
             feeCredit,
             tokens,
-            bytes.concat(code, bytes32(salt))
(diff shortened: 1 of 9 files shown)


shermike:
LGTM, but tbh I'm not sure that I understand the purpose of such change. Both `asyncCall` and `sendRequest` precompiles will be removed, so why we need to move `sendRequest` to `asyncCall`?

olegrok:
> LGTM, but tbh I'm not sure that I understand the purpose of such change. Both `asyncCall` and `sendRequest` precompiles will be removed, so why we need to move `sendRequest` to `asyncCall` now?

To remove one precompile right now. And to have common code for all communication types.
E.g. it was possible to send transactions to non-existent shard. So after this patch we will have common validation path for all calls.

PR #838, opened 2025-04-18
[EVM] Implement "Awaiter" contract

olegrok:
This patch reimplements some async context logic in pure EVM.


nil/internal/execution/account_state.go
```diff
@@ -274,6 +265,8 @@ func (as *AccountState) GetAsyncContext(index types.TransactionIndex) (*types.As
 func (as *AccountState) GetAndRemoveAsyncContext(index types.TransactionIndex) (*types.AsyncContext, error) {
 	ctx, exists := as.AsyncContext[index]
 	if exists {
+		// TODO?
```
shermike, line 268: Remove `TODO`?

shermike: Actually, it doesn't matter, as we will remove it anyway. But I suppose we won't call it twice for the same context id, thus I even think we can replace it with assert: `check.panicIf(exist)`.

PR #839, opened 2025-04-18
[rollup-bridge-contracts] Compile L1BridgeMessanger ABI by solc

akokoshn:
## Short Summary

Avoid using hardhart for build GO code.
The main idea is compile ABI required for `relayer` by `solc`.
It's possible if use solidity code without external dependencies, so just move required structures/events/functions to the stand alone library `L1BridgeMessengerEvents.sol`

## What Changes Were Made

- Move MessageSent event and required structures to library L1BridgeMessengerEvents.sol
- Remove rollup-bridge-contracts from dependency of the nil derivation
- Compile ABI of L1BridgeMessange, reuired for relayer from L1BridgeMessengerEvents.sol with using solc

## Checklist


rollup-bridge-contracts/contracts/bridge/l2/interfaces/IL2BridgeMessenger.sol
```diff
@@ -3,12 +3,13 @@ pragma solidity 0.8.28;
 
 import { IBridgeMessenger } from "../../interfaces/IBridgeMessenger.sol";
 import { NilConstants } from "../../../common/libraries/NilConstants.sol";
+import { IRelayMessage } from "IRelayMessage.sol";
```
defistar, line 6: import to be changed to:

```
import { IRelayMessage } from "./IRelayMessage.sol";
```

akokoshn: thanks!, done

nix/nil.nix
```diff
@@ -53,6 +48,7 @@ buildGo124Module rec {
     "^nil(/.*)?$"
     "^smart-contracts(/.*)?$"
     "^uniswap(/.*)?$"
+    "^rollup-bridge-contracts(/.*)?$"
```
oclaw, line 51: As far as I know it will affect CI caching. @AndreyMlashkin might have some concerns here

nil/services/relayer/generate.go
```diff
@@ -0,0 +1,8 @@
+package relayer
+
+import "embed"
+
+//go:generate bash -c "solc ../../../rollup-bridge-contracts/contracts/bridge/l1/interfaces/IRelayMessage.sol --abi --overwrite -o ../rollup-bridge-contracts-compiled-abi/contracts/bridge/l1/L1BridgeMessenger.sol --allow-paths .,../../../rollup-bridge-contracts/contracts/common/libraries --no-cbor-metadata --metadata-hash none --pretty-json"
```
oclaw, line 5: Why do we need to put this into the Go file? This artrifacts are not used by any piece of go code direclty.
And according to the current coding guidelines we are trying to minimize introduction of new `generate.go` file in general

nix/nil.nix
akokoshn: right, here should be added only `IRelayerMessage.sol`  l1/l2 interfaces

nil/services/relayer/generate.go
```diff
@@ -0,0 +1,8 @@
...
+//go:generate bash -c "solc ../../../rollup-bridge-contracts/contracts/bridge/l2/interfaces/IRelayMessage.sol --abi --overwrite -o ../rollup-bridge-contracts-compiled-abi/contracts/bridge/l2/L2BridgeMessenger.sol --allow-paths .,../../../rollup-bridge-contracts/contracts/common/libraries --no-cbor-metadata --metadata-hash none --pretty-json"
+
+var Fs embed.FS
```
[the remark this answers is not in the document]
akokoshn: Good catch, thanks
it just copy past)

nil/services/relayer/generate.go
akokoshn: got it,
no problem to call `solc` from the make

nix/nil.nix
akokoshn: @AndreyMlashkin , doe it possible to add single file by regexp?
```
rollup-bridge-contracts/contracts/bridge/l1/interfaces/IRelayMessage.sol
```

nil/services/relayer/Makefile.inc
```diff
@@ -1,13 +1,17 @@
-root_services = nil/services
 root_relayer = nil/services/relayer
 
+.PHONY: $(root_relayer)/compile_abi
+$(root_relayer)/compile_abi:
+	solc $(root_relayer)/../../../rollup-bridge-contracts/contracts/bridge/l1/interfaces/IRelayMessage.sol --abi --overwrite -o $(root_relayer)/rollup-bridge-contracts-compiled-abi/contracts/bridge/l1/L1BridgeMessenger.sol --allow-paths .,$(root_relayer)/../../../rollup-bridge-contracts/contracts/common/libraries --no-cbor-metadata --metadata-hash none --pretty-json
```
oclaw, line 5: let's add another `root_contracts` variable instead of doing ../../../ :)

oclaw (approved):

PR #840, opened 2025-04-21
[mpt] fix losing of data on delete from small MPT

olegrok:
Seems we should do the same as for update.
Otherwise we lost data in case if len(root) < 32.


nil/internal/mpt/mpt_test.go
```diff
@@ -1,11 +1,12 @@
-package mpt
+package mpt_test
```
shermike, line 1: Why is that?

olegrok: Here we use only public interfaces and don't use anything internal. So we can add "_test" suffix for that package.

> The test file can be in the same package as the one being tested, or in a corresponding package with the suffix "_test". 

https://pkg.go.dev/testing

olegrok: I can revert this part if you insist

shermike (approved):

Issue #846, opened 2025-04-22
Simple initialization of archive node


Un1oR:
We should have a simple one-line way to run an archive node that will be configured on the specified cluster

Un1oR:
Duplicates #786

PR #847, opened 2025-04-22
[GH-846] Added `--bootstrap-url` parameter for simple initialization of archive node

Un1oR:
This is an alternative approach to the one proposed in #813.

Pros:
- No hardcodes
- No infrastructure lock-in, works fine locally, with or without relays

Caveats:
- Establishing a connection using relays requires additional research, debugging and experimentation with `EnableAutoNATv2` and `EnableHolePunching` in real or near real conditions. To avoid being blocked by this now, I've enabled the `WithInfiniteLimits` setting on relays, and `WithAllowLimitedConn` on nodes. This should work "cheap but surly".

### DIFF
--- a/nil/cmd/nild/main.go
+++ b/nil/cmd/nild/main.go
@@ -4,7 +4,9 @@ import (
 	"context"
 	"errors"
 	"os"
+	"time"
 
+	rpc_client "github.com/NilFoundation/nil/nil/client/rpc"
 	"github.com/NilFoundation/nil/nil/cmd/nild/nildconfig"
 	"github.com/NilFoundation/nil/nil/common/check"
 	"github.com/NilFoundation/nil/nil/common/concurrent"
@@ -32,7 +34,7 @@ var logFilter string
 func main() {
 	logger := logging.NewLogger("nild")
 
-	cfg := parseArgs()
+	cfg := parseArgs(logger)
 
 	logging.ApplyComponentsFilter(logFilter)
 
@@ -100,7 +102,43 @@ func addBasicFlags(fset *pflag.FlagSet, cfg *nildconfig.Config) {
 		&cfg.CollatorTickPeriodMs, "collator-tick-ms", cfg.CollatorTickPeriodMs, "collator tick period in milliseconds")
 }
 
-func parseArgs() *nildconfig.Config {
+func doBootstrapRequestAndPatchConfig(
+	ctx context.Context,
+	bootstrapUrl string,
+	cfg *nildconfig.Config,
+	logger logging.Logger,
+) error {
+	ctx, cancel := context.WithTimeout(ctx, 15*time.Second)
+	defer cancel()
+
+	bootstrapConfig, err := rpc_client.NewClient(bootstrapUrl, logger).GetBootstrapConfig(ctx)
+	if err != nil {
+		return err
+	}
+
+	// Patch config
+	cfg.NShards = bootstrapConfig.NShards
+
+	if cfg.ZeroState != nil {
+		logger.Warn().Msg("overriding zero state config")
+	}
+	cfg.ZeroState = bootstrapConfig.ZeroStateConfig
+
+	if cfg.Network.TcpPort == 0 {
+		tcpPort := 3000 // Some port to work through libp2p
+		logger.Info().Msgf("setting TCP port to %d", tcpPort)
+		cfg.Network.TcpPort = tcpPort
+	}
+
+	cfg.BootstrapPeers = append(cfg.BootstrapPeers, bootstrapConfig.BootstrapPeers...)
+
+	cfg.Network.DHTBootstrapPeers = append(cfg.Network.DHTBootstrapPeers, bootstrapConfig.DhtBootstrapPeers...)
+	cfg.Network.DHTEnabled = true
+
+	return nil
+}
+
+func parseArgs(logger logging.Logger) *nildconfig.Config {
 	cfg, err := loadConfig()
 	check.PanicIfErr(err)
 
@@ -179,17 +217,30 @@ func parseArgs() *nildconfig.Config {
 	replayCmd.Flags().Var(&cfg.Replay.BlockIdLast, "last-block", "last block id to replay")
 	replayCmd.Flags().Var(&cfg.Replay.ShardId, "shard-id", "shard id to replay block from")
 
+	var bootstrapUrl string
 	archiveCmd := &cobra.Command{
 		Use:   "archive",
 		Short: "Run nil archive node",
-		Run: func(cmd *cobra.Command, args []string) {
+		RunE: func(cmd *cobra.Command, args []string) error {
 			cfg.RunMode = nilservice.ArchiveRunMode
+
+			if bootstrapUrl != "" {
+				return doBootstrapRequestAndPatchConfig(cmd.Context(), bootstrapUrl, cfg, logger)
+			}
+
+			return nil
 		},
 	}
 
 	addBasicFlags(archiveCmd.Flags(), cfg)
 	cmdflags.AddNetwork(archiveCmd.Flags(), cfg.Network)
 	cmdflags.AddTelemetry(archiveCmd.Flags(), cfg.Telemetry)
+	// N.B. Despite the fact that we override the config with the loaded configuration,
+	// we still handle this flag in the usual way rather than resorting to a hack like GetConfigNameFromArgs.
+	// Network-supplied values should win by default, but because we merge only a small, well-defined
+	// subset of settings, any rare conflicts can be resolved explicitly and safely.
+	archiveCmd.Flags().StringVar(
+		&bootstrapUrl, "bootstrap-url", "", "url to fetch initial configuration from (genesis config etc.")
 
 	rpcCmd := &cobra.Command{
 		Use:   "rpc",
--- a/nil/internal/network/addr_info.go
+++ b/nil/internal/network/addr_info.go
@@ -1,11 +1,18 @@
 package network
 
 import (
+	"errors"
+
 	"github.com/NilFoundation/nil/nil/common"
 	"github.com/libp2p/go-libp2p/core/peer"
+	"gopkg.in/yaml.v3"
 )
 
 type AddrInfo peer.AddrInfo
+
+func (a AddrInfo) Empty() bool {
+	return errors.Is(a.ID.Validate(), peer.ErrEmptyPeerID)
+}
 
 func (a *AddrInfo) Set(value string) error {
 	addr, err := peer.AddrInfoFromString(value)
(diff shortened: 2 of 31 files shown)


nil/cmd/nild/main.go
```diff
@@ -100,7 +104,49 @@ func addBasicFlags(fset *pflag.FlagSet, cfg *nildconfig.Config) {
...
+	client := rpc_client.NewRawClient(bootstrapUrl, logger)
+	response, err := client.RawCall(ctx, "debug_getBootstrapConfig")
+	if err != nil {
+		logger.Error().Err(err).Msg("failed to fetch bootstrap config")
+		return err
+	}
+
+	var bootstrapConfig rpctypes.BootstrapConfig
+	err = json.Unmarshal(response, &bootstrapConfig)
+	if err != nil {
```
olegrok, line 120:
```suggestion
	if err = json.Unmarshal(response, &bootstrapConfig); err != nil {
```

nil/cmd/nild/main.go
```diff
@@ -100,7 +104,49 @@ func addBasicFlags(fset *pflag.FlagSet, cfg *nildconfig.Config) {
 		&cfg.CollatorTickPeriodMs, "collator-tick-ms", cfg.CollatorTickPeriodMs, "collator tick period in milliseconds")
 }
 
-func parseArgs() *nildconfig.Config {
+func doBootstrapRequestAndPatchConfig(bootstrapUrl string, cfg *nildconfig.Config, logger logging.Logger) error {
```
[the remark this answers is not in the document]
Un1oR: We don't have a normal context in the place of use outside yet, so I don't see much point in doing this

nil/internal/network/addr_info.go
```diff
@@ -44,6 +51,18 @@ func (a *AddrInfo) UnmarshalText(text []byte) error {
 	return a.Set(string(text))
 }
 
+func (a AddrInfo) MarshalYAML() (any, error) {
```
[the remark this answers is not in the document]
Un1oR: Enough, but `musttag` fails because it [doesn't look](https://github.com/go-simpler/musttag/blob/9abbf3472a763b6d296c5dbe825532d6d5a9ca08/builtins.go#L21-L25) at `MarshalText`/`UnmarshalText` for yaml. I want to try to PR into it, but so far I haven't found a better way (other than changing the alias to nest the structure with `yaml:",inline"`, but I like that even less).

nil/cmd/nild/main.go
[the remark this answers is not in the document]
olegrok: What's about `cmd.Context()`?

nil/internal/network/addr_info.go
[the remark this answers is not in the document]
olegrok: Well... Could you try to use `MarshalYAML`/`UnmarshalYAML` as a wrappers around `MarshalText`/`UnmarshalText`?

E.g.
```
func (a *AddrInfo) UnmarshalYAML(value *yaml.Node) error {
	return a.UnmarshalText(value.Value)
}
```

nil/cmd/nild/main.go
```diff
@@ -100,7 +104,49 @@ func addBasicFlags(fset *pflag.FlagSet, cfg *nildconfig.Config) {
...
+	ctx, cancel := context.WithTimeout(context.Background(), 15*time.Second)
+	defer cancel()
+
+	client := rpc_client.NewRawClient(bootstrapUrl, logger)
```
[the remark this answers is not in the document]
Un1oR: This is partly a chicken and egg problem. We have a known entry point in the form of JSON-RPC API, there is an easy way to query it without initializing complex entities. This is what we are doing to find out the addresses of at least some peers in libp2p so that we can work with them further. I think the current approach is pretty simple and straightforward, and requires minimal configuration in a global sense.

nil/cmd/nild/main.go
```diff
@@ -100,7 +104,49 @@ func addBasicFlags(fset *pflag.FlagSet, cfg *nildconfig.Config) {
...
+	if err != nil {
+		logger.Error().Err(err).Msg("failed to fetch bootstrap config")
+		return err
+	}
+
+	var bootstrapConfig rpctypes.BootstrapConfig
+	err = json.Unmarshal(response, &bootstrapConfig)
+	if err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
Un1oR: Yeah, we can take it out there.

nil/internal/network/addr_info.go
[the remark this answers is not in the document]
Un1oR: Yes, that option seems to work. Thank you
```go
func (a *AddrInfo) MarshalYAML() (interface{}, error) {
	return a.MarshalText()
}

func (a *AddrInfo) UnmarshalYAML(value *yaml.Node) error {
	return a.UnmarshalText([]byte(value.Value))
}
```

nil/services/rpc/types/bootstrap_config.go
```diff
@@ -0,0 +1,13 @@
+package types
```
dmtrskv, line 1: (Not directly relevant to the PR.)

The package should've been called `rpctypes` so that one doesn't have to rename it all the time

(actually, I have the same thought about `connection_manager` which is contra at least two go conventions: no snake_case and a manager cannot be a package; the package should be what the manager manages; I suggest smth like `connections` there)

nil/services/rpc/rawapi/internal/node_builder.go
```diff
@@ -49,7 +53,8 @@ func (nb *nodeApiBuilder) WithLocalShardApiRo(shardId types.ShardId) *nodeApiBui
 }
 
 func (nb *nodeApiBuilder) WithLocalShardApiRw(shardId types.ShardId, txnpool txnpool.Pool) *nodeApiBuilder {
-	var localShardApi shardApiRw = newLocalShardApiRw(newLocalShardApiRo(shardId, nb.db), txnpool)
+	var bootstrapConfig *rpctypes.BootstrapConfig // = nil — not used in rw API methods
+	var localShardApi shardApiRw = newLocalShardApiRw(newLocalShardApiRo(shardId, nb.db, bootstrapConfig), txnpool)
```
dmtrskv, line 57: why not `shardApiRw := newLocalShardApiRw(newLocalShardApiRo(shardId, nb.db, nil), txnpool)`?

nil/services/rpc/rawapi/internal/node_builder.go
```diff
@@ -38,8 +39,11 @@ func (nb *nodeApiBuilder) BuildAndReset() NodeApi {
 	return &rv
 }
 
-func (nb *nodeApiBuilder) WithLocalShardApiRo(shardId types.ShardId) *nodeApiBuilder {
-	var localShardApi shardApiRo = newLocalShardApiRo(shardId, nb.db)
+func (nb *nodeApiBuilder) WithLocalShardApiRo(
+	shardId types.ShardId,
+	bootstrapConfig *rpctypes.BootstrapConfig,
+) *nodeApiBuilder {
+	var localShardApi shardApiRo = newLocalShardApiRo(shardId, nb.db, bootstrapConfig)
```
dmtrskv, line 46: let's stick to `:=`

nil/internal/network/host.go
```diff
@@ -88,7 +89,7 @@ func newHost(ctx context.Context, conf *Config) (Host, logging.Logger, error) {
 	}
 
 	if conf.ServeRelay {
-		options = append(options, libp2p.EnableRelayService())
+		options = append(options, libp2p.EnableRelayService(relay.WithInfiniteLimits()))
```
dmtrskv, line 92: add a comment please

nil/internal/network/addr_info.go
```diff
@@ -1,12 +1,19 @@
...
+
 	"github.com/NilFoundation/nil/nil/common"
 	"github.com/libp2p/go-libp2p/core/peer"
+	"gopkg.in/yaml.v3"
 )
 
 type AddrInfo peer.AddrInfo
 
+func (a AddrInfo) Empty() bool {
+	return errors.Is(a.ID.Validate(), peer.ErrEmptyPeerID)
```
dmtrskv, line 14: that's a really weird one, but ok, libp2p provides it that way

nil/services/nilservice/service.go
```diff
@@ -322,7 +346,11 @@ func createSyncers(
 				res.result = err
 				res.wgInit.Done()
 			}()
-			if err = initSyncers(ctx, res.syncers, cfg.AllowDbDrop); err != nil {
+			if err = initSyncers(
+				libp2pnetwork.WithAllowLimitedConn(ctx, "allow working through relay with traffic proxying"),
```
dmtrskv, line 350: shouldn't it be done on some lower level?

Un1oR: I think ultimately we need to deal with establishing a direct connection, but as I wrote in the PR description, it's not the most trivial task. So such an obvious "plug" that the eye clings to seems to me preferable to a solution that will pretend to be good.

nil/services/rpc/rawapi/internal/node_builder.go
Un1oR: Answered in a neighboring comment. I want to see the interface type explicitly.

nil/cmd/nild/main.go
```diff
@@ -179,17 +220,30 @@ func parseArgs() *nildconfig.Config {
...
 		},
 	}
 
 	addBasicFlags(archiveCmd.Flags(), cfg)
 	cmdflags.AddNetwork(archiveCmd.Flags(), cfg.Network)
 	cmdflags.AddTelemetry(archiveCmd.Flags(), cfg.Telemetry)
+	// N.B. Despite the fact that we override the config with the loaded configuration, we handle this flag as usual,
+	// and not through a hack like GetConfigNameFromArgs. There are two reasons for this:
+	// - We patch a limited number of settings and are confident that other processing should not depend on them
+	// - We believe that the loaded settings should take precedence over all others, so we apply them last.
```
dmtrskv, line 244: actually, the config hack is there to use config values as defaults for cli arguments
since this var is not from the config and its default is fixed I don't think there's a need in the comment
it kinda deceives

nil/services/rpc/rawapi/internal/node_builder.go
dmtrskv: To see the explicit interface type for what purpose?
(it's usually shown by an IDE btw)

nil/services/rpc/rawapi/pb/conversion_bootstrap_config.go
```diff
@@ -0,0 +1,71 @@
...
+				return nil, err
+			}
+			config.BootstrapPeers[i] = addrInfo
+		}
+		return config, nil
+
+	case *BootstrapConfigResponse_Error:
+		return nil, res.Error.UnpackProtoMessage()
+	}
+	return nil, errors.New("unexpected response type")
```
dmtrskv, line 70: makes sense to return the type in the error

dmtrskv (approved):

nil/services/rpc/rawapi/internal/node_builder.go
Un1oR: The point is that the "constructor" returns a pointer to the implementation (this may also be controversial, but so far I find it more practical). At the point of assignment, I emphasize that the code should not rely on the details of the implementation, but rather work with the interface.
Yes, this is probably really far-fetched, but I found it convenient.

nil/cmd/nild/main.go
Un1oR: I agree that the first clause is incorrect. I have corrected the comment.

PR #853, opened 2025-04-23
Add integration test of the rollup contracts

oclaw:
## Short Summary

End-to-end testing scripts for checking whole L1-L2 bridge work

Added to rollup-bridge-contracts additional integration testing step for end-to-end check of deposit message handling (full description of the test is in the `test_integration.sh`). Idea of moving it into separate derivation turned to overcomplicate things so it is how it is

Current limitations:
- L2 contracts are not deployed to nild yet
- Deposit message relaying is not ensured (because of the L2 contracts issue)

## What Changes Were Made

- Added shell script for running the integration test
- Added invocation of the script into the rollup-bridge-contracts derivation
- Added readiness probe to the relayer service (it does not have any exposed endpoints by design so it is the easiest way to make sure service is up and running)

## Checklist


nil/cmd/relayer/main.go
```diff
@@ -106,6 +106,7 @@ func addRunCommandFlags(runCmd *cobra.Command, cfg *Config) error {
 	runCmd.Flags().StringVarP(&cfgFile, "config", "c", "", "config file")
 
 	runCmd.Flags().StringVar(&cfg.DbPath, "db-path", "relayer.db", "path to database")
+	runCmd.Flags().IntVar(&cfg.ReadinessPort, "readiness-port", cfg.ReadinessPort, "port for readiness probe")
```
akokoshn, line 109: If `ReadinessPort` is listener port of relayer, may be better name `OwnPort` or `own-endpoint` like for sync committee.

nil/services/relayer/service.go
```diff
@@ -186,5 +192,24 @@ func (rs *RelayerService) Run(ctx context.Context) error {
 		return rs.L2TransactionSender.Run(ctx, transactionSenderStarted)
 	})
 
+	<-eventListenerStarted
+	<-finalityEnsurerStarted
+	<-transactionSenderStarted
+
+	readinessServer := &http.Server{
```
akokoshn, line 199: May be reuse [task listener approach](https://github.com/NilFoundation/nil/blob/6a3e7477c590a1710bea0c7713f5586961ae841e/nil/services/synccommittee/core/service.go#L99)
Also it may be useful to implement debug cli for check relayer status (stored events, ...)

nix/bridge-test.nix
```diff
@@ -0,0 +1,32 @@
+{ lib
```
akokoshn, line 1: seems `bridge-test` derivation can't be called via root `flake.nix`, and will not be executed by command `nix check`.
How are we going to use it?

oclaw: Oops, commited it by mistake, will remove

nil/cmd/relayer/main.go
oclaw: I wanted to highlight that it is only infra thing (that it does not contain any specific payload). But your suggestion to add some diagnostic endpoints is pretty interesting, will change it after adding diagnostics endpoint

flake.nix
```diff
@@ -109,6 +109,7 @@
           });
           rollup-bridge-contracts =
             (pkgs.callPackage ./nix/rollup-bridge-contracts.nix {
+              nil = packages.nil;
```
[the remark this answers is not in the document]
oclaw: Not anymore actually. And I guess it works because it is required only for checkPhase

nil/services/relayer/service.go
oclaw: Implemented debug RPC with stats endpoint, should be enough for the beginning

nil/services/relayer/internal/debug/debug.go
```diff
@@ -0,0 +1,211 @@
...
+		l1Storage:              l1Storage,
+		l2Storage:              l2Storage,
+		l1FinalizedBlockGetter: finalizedBlockGetter,
+		clock:                  clock,
+		logger:                 logger,
+		stats:                  atomic.Pointer[RelayerStats]{},
+	}
+}
+
+func (p *RelayerStatsAPI) runRefreshLoop(ctx context.Context, started chan struct{}) error {
```
akokoshn, line 143: Seems we may update statistic only when called `GetStats`, not in the permanent loop

akokoshn (approved):

nil/services/relayer/internal/debug/debug.go
oclaw: It is a whole db query (badger is not giving convenient table stats as far as there is actually no tables under the hood), so it is potentially possible to ddos service with frequent queries. I have decided to reduce this possibility by background polling

PR #855, opened 2025-04-24
[clijs] add receipt command

Gezort:
add receipt command to clijs. also during this I noticed that the CLI hangs for ~5-10 seconds after finishing the command logic
main reason seems to be in many promises hanging inside the node event loop. those are created during package imports or some open-rpc connections and closed only by timeout. so this PR also adds logic to exit from the process on success

### DIFF
--- a/clijs/src/common/config.test.ts
+++ b/clijs/src/common/config.test.ts
@@ -5,18 +5,18 @@ import ConfigManager, { ConfigKeys } from "./config.js";
 
 describe("ConfigManager", () => {
   CliTest("should create a default config file if it does not exist", async ({ cfgPath }) => {
-    const configManager = new ConfigManager(cfgPath);
+    fs.rmSync(cfgPath, { recursive: true, force: true });
+    expect(fs.existsSync(cfgPath)).toBe(false);
+    new ConfigManager(cfgPath);
     expect(fs.existsSync(cfgPath)).toBe(true);
   });
 
-  CliTest("should load the default config", async ({ cfgPath }) => {
-    const configManager = new ConfigManager(cfgPath);
+  CliTest("should load the default config", async ({ configManager }) => {
     const config = configManager.loadConfig();
     expect(config).toHaveProperty("nil");
   });
 
-  CliTest("should get a config value", async ({ cfgPath }) => {
-    const configManager = new ConfigManager(cfgPath);
+  CliTest("should get a config value", async ({ configManager }) => {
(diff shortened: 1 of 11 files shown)


clijs/src/common/config.test.ts
```diff
@@ -1,22 +1,19 @@
 import * as fs from "node:fs";
 import { describe, expect } from "vitest";
 import { CliTest } from "../../test/setup.js";
-import ConfigManager, { ConfigKeys } from "./config.js";
+import { ConfigKeys } from "./config.js";
 
 describe("ConfigManager", () => {
   CliTest("should create a default config file if it does not exist", async ({ cfgPath }) => {
```
Un1oR, line 7: This is a strange test. It seems to check the same test code (`CliTest`).

clijs/test/commands/receipt.test.ts
```diff
@@ -0,0 +1,25 @@
...
+        "--amount",
+        "100",
+        "--feeCredit",
+        10_000_000_000_000 as unknown as string,
+      ])
+    ).result as Hex;
+    expect(txHash).toBeTruthy();
+
+    const r = (await runCommand(["receipt", txHash])).result as ProcessedReceipt;
+    expect(r.success).toBeTruthy();
```
Un1oR, line 23: Let's also check that `output` contains the string `Receipt data:`.
By the way, I guess in the case of `--quiet` we only need to print the json itself.

Gezort: fixed. also fixed for the `block` command

Un1oR: Doesn't [this](https://github.com/oclif/test/blob/0e9113800a3edb99779316a68dc91d6f004d4aa1/test/run-command.test.ts#L13) approach work?

Rolaman (approved):

clijs/test/commands/receipt.test.ts
Un1oR: I did some research, and found out that without `disableConsoleIntercept` in `vitest`, intercepting the output in `runCommand` doesn't work. They are in conflict. We need to figure out how to make them work together correctly.

Un1oR: Hmm. The [documentation](https://www.npmjs.com/package/@oclif/test) for `oclif/test` explicitly says to use `disableConsoleIntercept` for `vitest`.

Un1oR: #866

PR #795, opened 2025-04-11
sync committee: fix state root divergence

oclaw:
## Short Summary

In case of disabled L1 fetcher it is invalid to fetch new state root via RPC if some non-empty is already stored in local database

## What Changes Were Made

- DisableL1 flag is forwarded to the proposer config (false by default)
- ProposerStorage is extended with new method of getting current state root
- In case of L1 is disabled and storage contains some non-empty state root value - it is used instead of the "fetched" one

## Checklist

- [x] I have read and followed the [Contributing Guide](https://github.com/NilFoundation/nil/blob/main/CONTRIBUTION-GUIDE.md)
- [x] I have tested the changes locally
- [ ] I have added relevant tests (if applicable)
- [ ] I have updated documentation/comments (if applicable)

PR #860, opened 2025-04-25
Sync Committee: State Root Synchronization

zadykian:
## Sync Committee: State Root Synchronization

Follow-up #798 #795

Introduced a mechanism to synchronize finalized state root stored locally with the L1 and L2 networks. Synchronization is triggered:

* By the new `stateRootSyncRunner` worker during Sync Committee node startup. It will retry syncing the state until it succeeds or execution is canceled.
* During full state reset (`StateResetLauncher.LaunchResetToL1WithSuspension()`).

Synchronization logic is determined by the `DisableL1` flag and implemented in the `stateRootSyncer` type


nil/services/synccommittee/core/service.go
```diff
@@ -112,7 +121,7 @@ func New(ctx context.Context, cfg *Config, database db.DB) (*SyncCommittee, erro
 
 	syncCommittee.Service = srv.NewService(
 		logger,
-		proposer, agg, lagTracker, taskScheduler, taskListener,
+		syncRunner, proposer, agg, lagTracker, taskScheduler, taskListener,
```
akokoshn, line 124: Why we need to check local state root each N sec?

zadykian: `syncRunner` synchronizes state root only once during startup, before all other workers are launched :)

```go

// stateRootSyncRunner is responsible for synchronizing the local state root with the
// latest finalized state root from L1.
//
// The runner will retry syncing the state root during application startup
// until it succeeds or the context is canceled.
//
// Once the runner successfully syncs the state root, it sends a
// message on the started channel and exits, letting the following workers to start.
type stateRootSyncRunner struct {
    // ...
}
```

akokoshn (approved):

PR #867, opened 2025-04-28
[execution, contracts] Introduce Relayer and non-enshrined tokens

shermike:
Description: TBD


nil/client/direct_client.go
```diff
@@ -348,7 +348,7 @@ func (c *DirectClient) SetTokenName(
 		return common.EmptyHash, err
 	}
 
-	return c.SendExternalTransaction(ctx, data, contractAddr, pk, types.NewFeePackFromGas(100_000))
+	return c.SendExternalTransaction(ctx, data, contractAddr, pk, types.NewFeePackFromGas(500_000))
```
dmtrskv, line 351: let's introduce a shared const for all cases

nil/internal/contracts/contract.go
```diff
@@ -241,7 +262,7 @@ func DecodeCallData(method *abi.Method, calldata []byte) (string, error) {
 
 	args, err := method.Inputs.Unpack(calldata[4:])
 	if err != nil {
-		return "", fmt.Errorf("failed to unpack arguments: %w", err)
+		return fmt.Sprintf("%s: failed to unpack arguments: %s", method.Name, err), nil
```
dmtrskv, line 265: could you comment on why is that?

nil/internal/execution/account_state.go
```diff
@@ -12,25 +12,7 @@ import (
...
-		return res
-	}
-	res, err := asr.TokenTrieReader.Fetch(id)
-	if errors.Is(err, db.ErrKeyNotFound) {
-		return types.Value{}
-	}
-	check.PanicIfErr(err)
-	return *res
-}
+type AccountStateReader struct{}
```
dmtrskv, line 15: why is it left at all?

nil/internal/execution/state.go
```diff
@@ -312,6 +311,11 @@ func NewExecutionState(tx any, shardId types.ShardId, params StateParams) (*Exec
 	}
 	logger := l.Logger()
 
+	// FIXME: remove
```
dmtrskv, line 314: remove at what moment?

nil/tests/cli/cli_test.go
```diff
@@ -599,10 +600,11 @@ func (s *SuiteCliExec) TestCliCometa() {
...
 		s.Require().Contains(result[0]["Transaction"], txnHash)
-		s.Require().Equal("Counter", result[1]["Contract"])
-		s.Require().Equal("get()", result[1]["CallData"])
-		s.Require().Equal("Counter", result[1]["Contract"])
-		s.Contains(out, "└ eventValue: [0]")
+		// TODO: we should skip Relayer from the output of `nil debug`
+		// s.Require().Equal("Counter", result[1]["Contract"])
+		// s.Require().Equal("get()", result[1]["CallData"])
+		// s.Require().Equal("Counter", result[1]["Contract"])
+		// s.Contains(out, "└ eventValue: [0]")
```
dmtrskv, line 607: what's the future of this code?

dmtrskv (commented):
There is too much to comprehend it completely.
Is it possible to make separate pull requests for the refactoring and deletion, and for the new features.

nil/client/direct_client.go
shermike: Done in #907

nil/internal/contracts/contract.go
shermike: We found method but failed to unpack arguments. The user should be warned about it.

nil/internal/execution/account_state.go
shermike: It can be removed, I guess. But I don't want to overload this already not small PR.

nil/internal/execution/state.go
shermike: Removed.

PR #868, opened 2025-04-29
start nild binary inside clijs tests

Gezort:
this PR makes clijs tests self-sufficient, now you need to provide `nild` binary to correctly run tests

### DIFF
--- a/clijs/test/globalSetup.ts
+++ b/clijs/test/globalSetup.ts
@@ -1,0 +1,46 @@
+import { type ChildProcess, spawn } from "node:child_process";
+import { closeSync, openSync } from "node:fs";
+import { testEnv } from "./testEnv.js";
+
+interface Nild {
+  pid: number;
+  process: ChildProcess;
+  logFd: number;
+}
+
+let nildInstance: Nild;
+
+export async function setup() {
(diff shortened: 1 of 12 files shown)


clijs/test/globalSetup.ts
```diff
@@ -0,0 +1,46 @@
...
+  pid: number;
+  process: ChildProcess;
+  logFd: number;
+}
+
+let nildInstance: Nild;
+
+export async function setup() {
+  console.log("launching nild:", testEnv.nild);
+  const logFd = openSync("nild.log", "w");
```
olegrok, line 15: I don't see any logs in CI console. Will be it possible to see a content of that file in case of test failure?

Gezort: are you sure you need the whole `nild.log` in CI console? I suggest keeping only test logs there. the rest can be inspected during local run (btw currently it works in the same way)

olegrok: It would we useful in case of failure. Sometimes some issues happen only in CI.

Gezort: ok, changed this. if it turns out to be inconvenient, we can always turn off

olegrok (approved):

PR #870, opened 2025-04-29
sync_committee: L1 fee params updater

oclaw:
## Short Summary

This PR introduces L1 fee params updater component into the sync committee service

## What Changes Were Made

Added new `feeupdater.Updater` component which is responsible for: 
- fetching latest blocks from all cluster shards 
- calculating new `maxFeePerGas` value according to the config
- checking if it is time to publish updated value to L1
- publishing data to the `NilGasPriceOracle` contract

## Checklist

- [ ] I have tested the changes locally


nil/services/synccommittee/core/feeupdater/contract_wrapper.go
oclaw, line 1: ABI integration will be done in the following PR after @defistar updates are merged

nil/services/synccommittee/core/feeupdater/updater.go
```diff
@@ -0,0 +1,216 @@
...
+		}
+	}
+	u.logger.Debug().
+		Uint32("shard_id", uint32(shardWithMaxBaseFee)).
+		Uint256("max_current_base_fee", maxBaseFee.String()).
+		Msg("fetched new value of max base fee")
+
+	maxBaseFee = maxBaseFee.Mul64(uint64(100 + u.config.MarkupPercent)).Div64(100)
+
+	const minFeePerGas = 20_000_000 // 0.02 gwei transformed into wei
```
akokoshn, line 129: make sense to configure via config like `MaxPriorityFeePerGasFixed`

nil/services/synccommittee/core/feeupdater/updater.go
```diff
@@ -0,0 +1,216 @@
...
+	if err := u.contractBinding.SetOracleFee(ctx, update); err != nil {
+		u.logger.Error().Err(err).Msg("failed to update L1 state")
+		return err
+	}
+
+	// Do we need to preseve these values between service restarts?
+	u.state.feeParams = &update
+	u.state.lastUpdatedTimestamp = now
+
+	u.logger.Info().Msg("L1 fee data updated")
```
akokoshn, line 172: make sense add metric for check how frequently SC update this data

oclaw: good point, added

nil/services/synccommittee/core/feeupdater/updater.go
oclaw: done

nil/services/synccommittee/core/feeupdater/updater.go
```diff
@@ -0,0 +1,226 @@
...
+
+func (u *Updater) Name() string {
+	return "l1_fee_updater"
+}
+
+func (u *Updater) Run(ctx context.Context, started chan<- struct{}) error {
+	return u.runUpdater(ctx, started)
+}
+
+func (u *Updater) runUpdater(ctx context.Context, started chan<- struct{}) error {
```
zadykian, line 92: Let's use `srv.WorkerLoop` here for consistency across all workers

nil/services/synccommittee/core/feeupdater/updater.go
```diff
@@ -0,0 +1,226 @@
...
+			return err
+		}
+
+		if block.BaseFee.Cmp(maxBaseFee) > 0 {
+			maxBaseFee = block.BaseFee
+			shardWithMaxBaseFee = shard
+		}
+	}
+	u.logger.Debug().
+		Uint32("shard_id", uint32(shardWithMaxBaseFee)).
```
zadykian, line 132: Not critical, but usually log attribute keys are specified in `camelCase`

nil/services/synccommittee/core/feeupdater/updater_test.go
```diff
@@ -0,0 +1,260 @@
...
+	entered := false
+	s.l1ContractMock.SetOracleFeeFunc = func(
+		_ context.Context,
+		params feeParams,
+	) error {
+		s.Require().False(entered, "expected only one update")
+
+		s.EqualValues(250_000_000, params.maxFeePerGas.Uint64())
+		s.EqualValues(10_000_00, params.maxPriorityFeePerGas.Uint64())
+		close(updateCalled)
```
zadykian, line 162: Does the channel really should be closed here?

nil/services/synccommittee/core/feeupdater/updater.go
oclaw: Current implementation of WorkerLoop is relying on system timer so it was not convenient to test. 

I have made an adjustment (switched to `clockwork.Clock` here) and moved `close(started)` below the ticker initialization in order to avoid races inside tests. Default behavior of other workers is not affected, they keep using default clock

nil/services/synccommittee/core/feeupdater/updater.go
oclaw: fixed

nil/services/synccommittee/core/feeupdater/updater.go
zadykian: I don't think anything prevents us from using an external `clockwork.Clock` across all workers by default
But that can be done later, not in this PR :)

akokoshn (approved):

nil/services/synccommittee/internal/srv/worker_loop.go
```diff
@@ -5,21 +5,51 @@ import (
...
+type (
+	WorkerLoopAction = func(ctx context.Context) error
+	StarterAction    = func(ctx context.Context) error
+)
 
 type WorkerLoopConfig struct {
 	Name     string
 	Interval time.Duration
 	Action   WorkerLoopAction
+	Starter  StarterAction
```
zadykian, line 20: It seems that starter isn't used anywhere

zadykian (approved):

nil/services/synccommittee/internal/srv/worker_loop.go
oclaw: It was my thought initially to do what I want via Starter but in the end I have come up with explicit timer solution
But Starter still could be useful as a concept so I decided to keep it

PR #874, opened 2025-05-01
Proof Provider: Persistent TaskExecutorId

zadykian:
### Proof Provider: Persistent TaskExecutorId

Refactor `TaskExecutor` and related components to use `executor.IdSource` for generating `TaskExecutorId`

Previously, both `Prover` and `ProofProvider` used random "nonce" IDs that were regenerated on each service startup. While it's fine for `Prover` to use a volatile identifier, `ProofProvider` requires a persistent ID to support top-level task result submission after a restart

Introduced `IdSource` implementations:
* `InMemoryIdSource` – used by `Prover`
* `PersistentIdSource` – backed by a database and used by `ProofProvider`

### DIFF
--- a/nil/services/synccommittee/internal/executor/id_source.go
+++ b/nil/services/synccommittee/internal/executor/id_source.go
@@ -1,0 +1,47 @@
+package executor
+
+import (
+	"context"
+	"sync"
+
+	"github.com/NilFoundation/nil/nil/services/synccommittee/internal/types"
+)
+
+type inMemoryIdSource struct {
+	cachedId types.TaskExecutorId
+	once     sync.Once
+}
+
+// NewInMemoryIdSource creates a new `IdSource` which generates a random `TaskExecutorId` on
+// the first call of `GetCurrentId` and caches it in memory.
+func NewInMemoryIdSource() IdSource {
+	return &inMemoryIdSource{}
+}
+
+func (n *inMemoryIdSource) GetCurrentId(context.Context) (*types.TaskExecutorId, error) {
+	n.once.Do(func() {
+		n.cachedId = types.NewRandomExecutorId()
+	})
+
+	return &n.cachedId, nil
+}
+
+type IdStorage interface {
+	GetOrAdd(ctx context.Context, idGenerator func() types.TaskExecutorId) (*types.TaskExecutorId, error)
+}
+
+type persistentIdSource struct {
+	storage IdStorage
+}
+
+// NewPersistentIdSource creates a new `IdSource` backed by persistent storage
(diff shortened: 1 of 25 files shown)


nil/services/synccommittee/internal/executor/id_source.go
```diff
@@ -0,0 +1,75 @@
...
+
+type idGenerator struct{}
+
+func (idGenerator) generateRandomId() (*types.TaskExecutorId, error) {
+	bigInt, err := rand.Int(rand.Reader, big.NewInt(math.MaxInt32))
+	if err != nil {
+		return nil, fmt.Errorf("failed to generate random id: %w", err)
+	}
+
+	randomId := types.TaskExecutorId(uint32(bigInt.Uint64()))
```
akokoshn, line 22: result converted to `uint32`, so may be generate random `uint32`?

nil/services/synccommittee/internal/scheduler/task_cancel_checker.go
```diff
@@ -60,23 +61,28 @@ func (c *TaskCancelChecker) runIteration(ctx context.Context) {
 }
 
 func (c *TaskCancelChecker) processRunningTasks(ctx context.Context) error {
-	canceledCounter, err := c.taskSource.CancelTasksByParentId(ctx, c.isActive)
+	executorId, err := c.executorIdSource.GetCurrentId(ctx)
+	if err != nil {
+		return fmt.Errorf("failed to get current executor id: %w", err)
+	}
+
+	isActiveFunc := func(ctx context.Context, id types.TaskId) (bool, error) {
```
oclaw, line 69: What do we mean here by `isActive`?

oclaw (approved):

nil/services/synccommittee/internal/executor/id_source.go
zadykian: Replaced with a simple `rand.Uint32()` call

nil/services/synccommittee/internal/scheduler/task_cancel_checker.go
zadykian: This function checks if a task with the given `Id` is still active
Renamed to `isTaskActiveFunc`

nil/services/synccommittee/internal/types/prover_tasks.go
```diff
@@ -1,9 +1,11 @@
 package types
 
 import (
+	"encoding/binary"
 	"errors"
 	"fmt"
 	"iter"
+	mathRand "math/rand"
```
akokoshn, line 8: why not just `rand`?

akokoshn (approved):

PR #875, opened 2025-05-01
Sync Committee: L1 Communication Fix

zadykian:
### Sync Committee: L1 Communication Fix

This commit introduces several fixes to the communication between the Sync Committee and the Rollup contract:

* Added support for lazy state root initialization in the `NilRollup` contract
* The contract is now initialized with an empty `genesisStateRoot` during deployment
* Refactored public data input validation (behavior remains unchanged)
* The `wrapper.simulateTx` method now correctly uses all sidecar-related data during contract calls


nil/common/hash.go
```diff
@@ -65,6 +65,10 @@ func IntToHash(i int) Hash {
 	return BigToHash(b)
 }
 
+func (h Hash) Equals(other Hash) bool {
```
oclaw, line 68: Hash is an alias for fixed-size array, why not just `hash1 == hash2`?

nil/services/synccommittee/core/rollupcontract/commit_batch.go
```diff
@@ -18,26 +18,19 @@ import (
 
 // CommitBatch creates blob transaction for `CommitBatch` contract method and sends it on chain.
 // If such `batchIndex` is already submitted, returns `nil, ErrBatchAlreadyCommitted`.
-func (r *wrapperImpl) CommitBatch(
-	ctx context.Context,
-	sidecar *ethtypes.BlobTxSidecar,
-	batchIndex string,
-) error {
+func (r *wrapperImpl) CommitBatch(ctx context.Context, batchId types.BatchId, sidecar *ethtypes.BlobTxSidecar) error {
```
akokoshn, line 21: seems logically keep `batchId string` for avoid dependency from SC types

rollup-bridge-contracts/contracts/NilRollup.sol
```diff
@@ -169,7 +169,7 @@ contract NilRollup is OwnableUpgradeable, PausableUpgradeable, NilAccessControlU
             isFinalized: true,
             versionedHashes: new bytes32[](0),
             oldStateRoot: ZERO_STATE_ROOT,
-            newStateRoot: _genesisStateRoot,
+            newStateRoot: ZERO_STATE_ROOT,
```
akokoshn, line 172: idea was initialize contract with real `genesisStateRoot` why we have to set zero here?

nil/services/synccommittee/core/rollupcontract/commit_batch.go
zadykian: `rollupcontract.Wrapper` already uses SC types :)
For instance, `PrepareBlobs(...)` methods returns `types.DataProofs`. Personally, I don't see anything wrong in it

rollup-bridge-contracts/contracts/NilRollup.sol
zadykian: To initialize the contract with the actual `genesisStateRoot` from L2, we would need to access the Nil RPC during deployment. This would make the setup process even more complicated than it already is

So, the alternative approach is the following:

1. `NilRollup` contract is initialized with an empty state root
2. `SC` fetches the genesis block if `NilRollup.finalizedStateRoots` returns an empty value
3. `NilRollup.updateState` accepts the genesis hash along with the first batch submitted by the `SC`

zadykian: Agreed to decouple deployment and initialization. State root on L1 will be initialized by `SC` during startup

Moving this PR to draft for now

nil/services/synccommittee/core/rollupcontract/wrapper.go
```diff
@@ -348,47 +385,90 @@ func (r *wrapperImpl) errorByName(err error) error {
...
+
 	return nil
 }
 
-func (w *noopWrapper) LatestFinalizedStateRoot(context.Context) (common.Hash, error) {
+func (w *noopWrapper) GetLatestFinalizedStateRoot(context.Context) (common.Hash, error) {
 	w.logger.Debug().Msg("FinalizedStateRoot noop wrapper method called")
-	return common.Hash{}, nil
+
+	w.mutex.Lock()
```
akokoshn, line 455: `RLock`?

nil/services/synccommittee/core/rollupcontract/wrapper.go
```diff
@@ -348,47 +385,90 @@ func (r *wrapperImpl) errorByName(err error) error {
...
+
+	return nil
+}
+
+func (w *noopWrapper) UpdateState(_ context.Context, data *types.UpdateStateData) error {
 	w.logger.Debug().Msg("UpdateState noop wrapper method called")
+
+	w.mutex.Lock()
+	defer w.mutex.Unlock()
+	w.stateRoot = data.NewProvedStateRoot
```
akokoshn, line 447: `SetGenesisStateRoot` make additional check - empty `w.stateRoot`, so maby make sense add check here: `w.stateRoot = data.OldProvedStateRoot`

akokoshn (commented):
Looks ok for me.
Let's wait for approve from @defistar

nil/services/synccommittee/core/rollupcontract/wrapper.go
zadykian: done

nil/services/synccommittee/core/rollupcontract/wrapper.go
```diff
@@ -348,47 +385,101 @@ func (r *wrapperImpl) errorByName(err error) error {
...
+		return "latest"
+	}
+	return hexutil.EncodeBig(number)
+}
+
 type noopWrapper struct {
-	logger logging.Logger
+	logger    logging.Logger
+	stateRoot common.Hash
+	mutex     sync.RWMutex
```
x-mass, line 424: Is it expected you use it only for `noopWrapper`, but not for `wrapperImpl`?

zadykian: Yep, `wrapperImpl` itself doesn't have any mutable state, but `noopWrapper` does

defistar (approved):

Issue #852, opened 2025-04-23
Improved mock generation


Un1oR:
We are currently [using](https://github.com/NilFoundation/nil/blob/f4b969591263b4ecb07ca7926628516d90b6cce2/nil/client/client.go#L20) `matryer/moq` in conjunction with `//go:generate` and [Makefile](https://github.com/NilFoundation/nil/blob/f4b969591263b4ecb07ca7926628516d90b6cce2/nil/Makefile.inc#L18-L18). The generated files don't have the `//go:build test` annotation, so they end up in the application executable, which is not necessary. As far as I can tell, `matryer/moq` doesn't have a setting to fix this. We might consider [mockery](https://vektra.github.io/mockery/v3.0/), which looks more flexible and is also centrally configurable via yaml-config.

BenFaruna:
Is it okay if I take this?

PR #877, opened 2025-05-02
Switch mock generation from **moq** to **mockery** (build-tag `test`) markdown

reject-i:
### What’s inside
* **.mockery.yaml** – central config (`mock-build-tags: "test"`, `inpackage`, `with-expecter`).
* Replaced every `//go:generate moq …` with `//go:generate mockery --name=…`.
* Deleted legacy `*_generated_mock.go` / `*_moq_test.go`; added new `mock_*.go` (guarded by `//go:build test`).
* Makefile targets **mocks / generate_mocks**  
  * regenerate SSZ first  
  * iterate over packages that contain `//go:generate mockery` and run `go generate`.
* CI / local build: mocks stay out of production binary, are reproducible via `make mocks`.

Closes #[852](https://github.com/NilFoundation/nil/issues/852)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -87,6 +87,6 @@ openrpc.json
 .qodo
 nil/**/mock_*.go
 
-# mockery output (исключаем helper-тесты)
+# mockery output (excluding helper-tests)
 nil/**/mock_*.go
 !nil/**/mock_*_test.go


nil/client/mock_Client.go
```diff
@@ -0,0 +1,2436 @@
+// Code generated by mockery v2.53.3. DO NOT EDIT.
```
olegrok, line 1: We don't commit autogenerated files into repository. Please add them to gitignore.

Makefile
```diff
@@ -107,3 +107,9 @@ clean:
 	go clean -cache
 	rm -fr build/*
 	rm -fr contracts/compiled/*
+
+.PHONY: mocks
+mocks: ssz_types
+	@for p in $(shell grep -rl '//go:generate mockery' --exclude-dir=vendor | xargs -n1 dirname | sort -u); do \
+		go generate -run "mockery" $$p; \
+	done
```
olegrok, line 115: You can remove `go:generate` usage and use explicit make targets to avoid such strange constructions.

Un1oR: Ideally, all the customization for what to generate mocks for should now be in `.mockery.yaml`. Via `include-interface-regex` or whatever.

nil/services/relayer/Makefile.inc
```diff
@@ -20,9 +20,9 @@ $(root_relayer)/generate_l1_abi: $(root_relayer)/embed_l1_abi
...
 $(root_relayer)/gen_l1_mocks: $(root_relayer)/generate_l1_abi
-	cd $(root_relayer)/internal/l1 && go run github.com/matryer/moq -out eth_client_generated_mock.go -rm -stub -with-resets . EthClient
-	cd $(root_relayer)/internal/l1 && go run github.com/matryer/moq -out l1_contract_generated_mock.go -rm -stub -with-resets . L1Contract
+	cd $(root_relayer)/internal/l1 && mockery --name=EthClient
+	cd $(root_relayer)/internal/l1 && mockery --name=L1Contract
 
 .PHONY: $(root_relayer)/gen_l2_mocks
 $(root_relayer)/gen_l2_mocks: $(root_relayer)/embed_l2_abi
-	cd $(root_relayer)/internal/l2 && go run github.com/matryer/moq -out l2_contract_generated_mock.go -rm -stub -with-resets . L2Contract
+	cd $(root_relayer)/internal/l2 && mockery --name=L2Contract
```
olegrok, line 28: Did you try to build a project?

.gitignore
```diff
@@ -85,3 +85,8 @@ openrpc.json
 .pnpm/
 .pnpm-store/
 .qodo
+nil/**/mock_*.go
+
+# mockery output (исключаем helper-тесты)
```
olegrok, line 90: Usually we write comments in English even we use AI tools

zadykian:
Do we really need to migrate to `mockery`?  
What benefits are we getting by adding a new build dependency and updating all existing tests?

Regarding the `//go:build test` annotation, it can be easily handled by a [script](https://github.com/NilFoundation/nil/blob/main/nil/services/synccommittee/internal/scripts/generate_mock.sh#L61).

Un1oR:
@zadykian I'll answer as the author of the original proposal)

As far as I see, `mockery` is more actively maintained and developed than `moq`. Maybe it's not so important for a "just working" library, but still.
OK, we could add workaround for build tags, but why when it's already there out of the box?
I like the idea of being able to remove unnecessary stuff from Makefiles. The boilerplate for various code generation in them makes me sad.
Also `mockery` positions itself not only as a mock generator, but also as a framework for generating general purpose code using customizable templates. This could be interesting for replacing reflection with codegeneration in RawAPI, for example.
Well, and adding one dependency _instead_ of another doesn't look like a big problem.
And test adaptation is part of the task. If it is accomplished, we don't have to worry about additional effort.

Un1oR:
@reject-i do you plan to correct the comments?

PR #880, opened 2025-05-06
[GH-#] Added playground for experiments with libp2p

Un1oR:
Added a playground for debugging libp2p. See `nil/libp2p-playground/README.md` for details.

Initially I planned to debug hole-punching, but
a) I couldn't get it to work quickly
b) I realized that in our current production topology such a task is not worth it now.

Therefore, after consulting with @Zerg1996, I decided to postpone further research on the hole-punching issue, and to commit the current developments as they are, so that they can be used as a starting point for future research.

For the same reason, the current solution does not have deep integration with the rest of the codebase (Nix, etc.). If this stand is developed and this is required (e.g. on CI), it could be addressed.


libp2p-playground/README.md
```diff
@@ -0,0 +1,44 @@
+Test environment for debugging the interaction of multiple nodes via libp2p with a specified network topology.
```
olegrok, line 1: Couldn't we put it into a separate repository?

Looks a bit unrelated to nil project.

Un1oR: I've been thinking about that option, too. I suppose we could do that. After all, it's never too late to move it back if it becomes necessary.

olegrok: My main concern is using of Docker and Redis. This is not typical for our repository where we use nix and focus only on nil development.

libp2p-playground/relay/Dockerfile
```diff
@@ -0,0 +1,17 @@
+# ---------- builder ---------
```
olegrok, line 1: Maybe we can use only one Dockerfile. The first one is a builder, the second one is a "base" with dependencies (iproute2-tc netcat-openbsd, ...) and then we can put separate binaries into "base" image.

PR #881, opened 2025-05-06
Sync Committee: Block Storage Fix

zadykian:
### Sync Committee: Block Storage Fix

This PR addresses the following error encountered on `nil-dev` after a full state reset was triggered:

```
error preparing batch: error loading latest batch from the storage: batch with the specified id is not found, id=23d4376b-f92e-488a-abb6-8a3dc8cf6c20
```

Changes:
* Fixed `BlockStorage.ResetAllBatches(...)` by setting `latestBatchId` to `nil`
* Added `BlockStorage.HasFreeSpace(...)` method to avoid unnecessary batch creation in `Aggregator`

### DIFF
--- a/nil/services/synccommittee/core/fetching/aggregator.go
+++ b/nil/services/synccommittee/core/fetching/aggregator.go
@@ -42,6 +42,7 @@ type AggregatorBlockStorage interface {
 	GetLatestFetched(ctx context.Context) (types.BlockRefs, error)
 	TryGetProvedStateRoot(ctx context.Context) (*common.Hash, error)
 	TryGetLatestBatch(ctx context.Context) (*types.BlockBatch, error)
+	HasFreeSpace(ctx context.Context) (bool, error)
 	PutBlockBatch(ctx context.Context, batch *types.BlockBatch) error
 }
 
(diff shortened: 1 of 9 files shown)


nil/services/synccommittee/core/fetching/aggregator.go
```diff
@@ -210,11 +213,27 @@ func (agg *aggregator) tryPrepareBatch(ctx context.Context) (*types.BlockBatch,
...
+	}
+	if canBeExtended {
+		return latestBatch, nil
+	}
+
+	return nil, nil
+}
+
+// checkLatestBatch checks the latest batch to determine whether it can be further extended or not
+func (agg *aggregator) checkLatestBatch(
```
x-mass, line 228: Not only this, but sealing, resetting the state. Mb move `constraints.CheckResultTypeCanBeExtended` case to caller (actually needs only this to determine either to attach blocks to latest batch or to the new one) and handle other cases in separate func?

x-mass (approved):

nil/services/synccommittee/core/fetching/aggregator.go
zadykian: Good suggestion, rearranged code a bit:

```go
func (agg *aggregator) tryPrepareBatch(ctx context.Context) (*types.BlockBatch, error) {
	// ...

	checkResult, err := agg.batchChecker.CheckConstraints(ctx, latestBatch)
	if err != nil {
		return nil, fmt.Errorf("error checking batch constraints, batchId=%s: %w", latestBatch.Id, err)
	}

	if checkResult.CanBeExtended() {
		return latestBatch, nil
	}

	return nil, agg.handleUnextendableBatch(ctx, latestBatch, checkResult)
}

```

Issue #769, opened 2025-04-07
Restart services on chain version change


dmtrskv:
Chain version is defined by the hash of the main shard's genesis block.
The following services should wipe their databases (based on an argument) on version change detected at the start:
- archive node
- cometa
- indexer
They should also restart if the version change is detected during their lifetime.

olegrok:
Just drop of DB is not enough (at least for archive node). You need to have a trusted genesis block source or zerostate config.
Cometa and indexer probably also need to verify their state against nil config

dmtrskv:
I plan to hardcode some chain data here: https://github.com/NilFoundation/nil/issues/786

Issue #659, opened 2025-03-25
Handle protocol version change


Un1oR:
Now the installation of an archive node does not know how to recognize that the version of the cluster it works with has been incompatibly updated. It is necessary to implement a mechanism for tracking such changes.

PR #884, opened 2025-05-08
[version] Restart services on chain or protocol change

dmtrskv:
Syncing nodes stop if the chain or protocol version changed.
Indexer stops if the chain version changed.

Changes:
- Moved existing version-related stuff to a separate file and refactored a bit for my needs.
- Added a version check loop to the syncers (called on the main shard syncer only).
- Added a version check loop to the indexer.
- Refactored indexer a bit.
- Implemented version check and db reset in badger indexer.

Fixes https://github.com/NilFoundation/nil/issues/769 and https://github.com/NilFoundation/nil/issues/659

Tested locally:
1. Start cluster -> Start archive node -> Stop cluster; drop validator keys and db (keep network keys); start the cluster anew -> Archive node stops after discovering the change in the version.
2. Same for indexer instead of archive.

### DIFF
--- a/nil/internal/collate/version.go
+++ b/nil/internal/collate/version.go
@@ -1,0 +1,127 @@
+package collate
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/NilFoundation/nil/nil/common"
+	"github.com/NilFoundation/nil/nil/common/check"
+	"github.com/NilFoundation/nil/nil/common/logging"
+	"github.com/NilFoundation/nil/nil/internal/db"
+	"github.com/NilFoundation/nil/nil/internal/network"
+	"github.com/NilFoundation/nil/nil/internal/types"
+)
+
+const topicVersion = "/nil/version"
+
+type NodeVersion struct {
+	ProtocolVersion  string
+	GenesisBlockHash common.Hash
+}
+
+type VersionChecker struct {
+	nm     network.Manager
+	fabric db.DB
+	logger logging.Logger
+}
+
+func NewVersionChecker(nm network.Manager, fabric db.DB, logger logging.Logger) *VersionChecker {
+	return &VersionChecker{
+		nm:     nm,
+		fabric: fabric,
+		logger: logger,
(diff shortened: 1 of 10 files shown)


nil/tests/archive_node/complex_test.go
```diff
@@ -84,7 +84,9 @@ func (s *SuiteArchiveNodeComplex) TestProtocolUpdate() {
 		s.stopArchiveNode()
 		s.Cancel()
 		s.startCluster(newProtocolVersion)
+	})
 
+	s.Run("RunArchiveNode", func() {
```
dmtrskv, line 89: added while debugging failing tests

nil/internal/collate/version.go
```diff
@@ -0,0 +1,103 @@
...
+	var err error
+	for _, peer := range peers {
+		var peerId network.PeerID
+		peerId, err = nm.Connect(ctx, network.AddrInfo(peer))
+		if err != nil {
+			continue
+		}
+
+		var protocolVersion string
+		protocolVersion, err = nm.GetPeerProtocolVersion(peerId)
```
olegrok, line 77: Let's add some logging here. I guess it can be useful for debugging.

dmtrskv: done. introduced a VersionChecker for the sake of OOP

nil/services/indexer/indexer.go
```diff
@@ -87,25 +90,41 @@ func (e *Indexer) waitForRpc(ctx context.Context, timeout time.Duration) error {
...
-		AllowDbDrop: e.allowDbDrop,
-		Version:     version,
-	}); err != nil {
+
+	localVersion, err := e.driver.FetchVersion(ctx)
+	if err != nil {
 		return nil, err
 	}
 
+	if !bytes.Equal(localVersion[:], remoteVersion[:]) {
```
olegrok, line 111: Isn't hashes comparable via `==`/`!=`?

olegrok (approved):

PR #899, opened 2025-05-14
[release] Release 2025.05.14 (0.2.0)

dmtrskv:
Release changes:

- Switching from poseidon to Keccak256 as hash function
- For more compatibility with Etherium all data now serialised using RLP mechanism instead of SSZ

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -19,7 +19,7 @@
       let
         revCount = self.revCount or self.dirtyRevCount or 1;
         rev = self.shortRev or self.dirtyShortRev or "unknown";
-        version = "0.1.5-${toString revCount}";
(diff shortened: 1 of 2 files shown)


ukorvl:
@dmtrskv niljs `0.26.0` is already published https://www.npmjs.com/package/@nilfoundation/niljs
You need to bump the version of niljs to include the latest changes in the new release and publish it.
If there were changes in `smart-contracts` package, its version should be bumped as well and published.
hardhat package and create-nil-hardhat-project also need to be bumped (because they depend on particular version of `smart-contracts`).

dmtrskv:
Replaced with https://github.com/NilFoundation/nil/pull/903

PR #907, opened 2025-05-16
Support all transaction types in Relayer

shermike:
- New Relayer functional:
	- Support gas forwarding
	- Support request/response transactions
	- Support deploy transactions
	- Support bounce transactions
- Introduce `async` modifier for functions that initiate cross-shard transactions. It is required to enable gas forwarding.
- Since deployment is currently performed by the Relayer (i.e., from Solidity code), we should leverage the CREATE2 opcode. Therefore, `Nil.CreateAddress` internally uses `Nil.CreateAddressForCreate2`, with the Relayer address as the sender.
- Add `NilBase.nilReceive()` method to allow receiving funds even if the contract lacks a `receive()` function.


smart-contracts/contracts/Relayer.sol
```diff
@@ -138,36 +357,100 @@ contract Relayer {
...
     }
 
+    /**
+     * @dev Calculates the gas required for a target call.
+     * @param request Indicates whether the call is a request.
+     * @return The amount of gas required for the call of the target contract.
+     */
+    function calculateGasForTargetCall(bool request) internal view returns(uint) {
+        uint gasForResponse = 50_000;
+        uint requiredGasForFinish = 50_000;
```
isergeyam, line 401: I think we should add a comment here explainig that requiredGasForFinish is actually for finishing receiveTx, otherwise it's not very clear

PR #931, opened 2025-05-28
[jsonrpc] add more fields into RPCBlock structure

olegrok:
Previously SmartContracts and OutTransaction roots were missed. Also removed redundant function in validator module (no functional changes).


olegrok:
> Have you checked if we need to add support for new fields explicitly in `clijs` or if the fields will be added transparently in JSON format?

Mainly I exported them to be able to read them via curl. We can return it via CLI as well but it's a separate task

PR #933, opened 2025-05-28
bridge state getter: fix block ref type

oclaw:
## Short Summary

By mistake used invalid type for referencing a block while getting bridge state from the L2 contract

### DIFF
--- a/nil/services/synccommittee/core/bridgecontract/bridge_state_getter.go
+++ b/nil/services/synccommittee/core/bridgecontract/bridge_state_getter.go
@@ -6,6 +6,7 @@ import (
 	"math/big"
 
 	"github.com/NilFoundation/nil/nil/client"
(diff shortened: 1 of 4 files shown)


nil/services/synccommittee/core/bridgecontract/bridge_state_getter.go
```diff
@@ -22,7 +23,7 @@ type BridgeState struct {
 }
 
 type BridgeStateGetter interface {
-	GetBridgeState(ctx context.Context, blockID any) (*BridgeState, error)
+	GetBridgeState(ctx context.Context, blockID common.Hash) (*BridgeState, error)
```
zadykian, line 26: Rename `blockID` to `blockHash` for clarity?

zadykian (approved):

Issue #240, opened 2025-02-10
Finish migrating CI for macOS to Nix


Un1oR:
We [have added](https://github.com/NilFoundation/nil/pull/158) macOS builds via Nix for `clijs`, but they take a very long time because we rebuild `nodejs` from scratch for each run. This is because the build results are not cached as they are in Linux runners thanks to `post-build-hook`.

We need to implement a similar mechanism in macOS builds where Nix is installed directly by the build step.

We need to consider the security issue. We need to fill in the `secret-key` parameter to send to the cache. If it will be available in the build steps, it is important that no third party can run CI with the steps modified in such a way that this value is logged in them.

PR #936, opened 2025-05-28
[GH-240] Nix cache for macOS CI

Un1oR:
The `nix_check_macos` matrix job has been added.
First we try to authorize our workflow via OIDC for write access in AWS to the Nix cache. This is possible for main and PR's according to the role settings. For runs via `workflow_dispatch` and forks, the access will be read-only. The cache bucket is made public read-only for this purpose.

Restored `clijs` builds for macOS `aarch64` & `x86_64`. Also added `nil` builds for the same platforms, but disabled for `x86_64` for now due to test issues. I plan to deal with this later and leave only Nix builds for macOS.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -5,6 +5,10 @@ on:
   workflow_dispatch:
   merge_group:
 
+permissions:
+  id-token: write
+  contents: read
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -26,54 +30,137 @@ jobs:
...
+            ${{ steps.calculate_write_flag.outputs.has_write && 'post-build-hook = /etc/nix/upload-to-cache.sh' }}
+
+      - name: Show /etc/nix/nix.conf
+        run: cat /etc/nix/nix.conf
+
+      # The following "build" command can be used for debugging without having to wait a long time:
+      # nix-build -E 'let pkgs = import (builtins.fetchTarball "https://github.com/NixOS/nixpkgs/archive/nixos-24.05.tar.gz") {}; in pkgs.writeText "example" "001"'
+      # Counter 001 can be modified to check for cache hits or misses.
+      - name: Run check
+        run: nix build .#checks.${{ matrix.platform.nixArch }}.${{ matrix.check == 'nil' && 'others' || matrix.check }} -L
```
olegrok, line 155: Not sure we need to run nil tests of MacOS at all. Especially because we are on the way of deprecation of go CLI.

olegrok (approved):

.github/workflows/main.yml
Un1oR: What's wrong with running tests on a platform used by a fraction of the developers?

olegrok: It's not officially supported platform for nild as result it can slow down development process

Un1oR: But after all, broken tests by some developers that CI missed can also slow down the development process.
In any case, we may change the decision to run tests on macOS later if there are real problems with it.

PR #943, opened 2025-06-02
rollup contracts: use faucet instead of nild

oclaw:
## Short Summary

In order to keep consistency between tests and nildev deployment now we are using faucet as a separate binary / RPC URL referred by L2 deployment scripts 

## What Changes Were Made

- Faucet included into rollup-bridge-contracts subPackages
- Separate env vars for binary name and rpc addr are introduced and used

## Checklist


rollup-bridge-contracts/test_integration.sh
```diff
@@ -107,7 +112,21 @@ echo "L1BridgeMessenger deployed to: $l1_contract_addr"
...
-wait_for_http_service "http://127.0.0.1:8529"
+
+nild_endpoint="http://127.0.0.1:8529"
+wait_for_http_service $nild_endpoint
+
+faucet_port=8527
+faucet_endpoint="http://127.0.0.1:$faucet_port"
+
+echo "Starting faucet service"
+$FAUCET_BIN run \
```
olegrok, line 123: Why do you need standalone faucet for tests?

oclaw: In order to be aligned with how we do it on the nildev. Right now we are facing some difficulties with our ansible part from infra repo and this patch brings more consistency between tests and dev deployment

olegrok: You can add FAUCET_RPC_ENDPOINT env handling for rollup-bridge-contract part. But other changes are excessive.

BTW, this won't work since you try to run `faucet` binary that is not built (https://github.com/NilFoundation/nil/blob/53677c9128e8323c04b6f5ace886a882a32788c9/flake.nix#L65)

PR #958, opened 2025-06-11
[execution] Improve caching in accessors + simplify code

dmtrskv:
State accessor appears in profiles as an active creator of objects in memory.
Here I simplify the code of the accessor tailoring it to the current needs. The caching is reworked and improved (it wasn't happening at some places at all).

There are also some simplifications in the other code, especially improvements in the usage of accessors.


nil/services/rpc/rawapi/internal/local_block.go
```diff
@@ -99,65 +99,48 @@ func (api *localShardApiRo) getBlockByHash(
...
-		for _, transaction := range transactions {
-			txnHash := transaction.Hash()
-			errMsg, err := db.ReadError(tx, txnHash)
-			if err != nil && !errors.Is(err, db.ErrKeyNotFound) {
-				return nil, err
-			}
-			if len(errMsg) > 0 {
-				result.Errors[txnHash] = errMsg
+		if len(errMsg) > 0 {
+			if result.Errors == nil {
```
Un1oR, line 138: How about building a temporary dictionary, and assigning it to `result.Errors` at the end if it's not empty? Or conversely, assigning `nil` to `result.Errors` if it's empty. By the way, why do we even want to maintain such an invariant?

nil/internal/collate/proposer_test.go
```diff
@@ -183,36 +182,17 @@ func (s *ProposerTestSuite) TestCollator() {
...
 	s.Run("Execute", func() {
 		m := execution.NewExecutionTransaction(to, to, 1, contracts.NewCounterAddCallData(s.T(), 3))
 		pool.Reset()
 		pool.Add(m)
 
-		generateBlock()
-		s.checkReceipt(shardId, m)
-	})
-
-	s.Run("CheckRefundsSeqno", func() {
```
Un1oR, line 199: Why did we delete this test?

nil/internal/execution/state_accessor.go
```diff
@@ -14,814 +11,241 @@ import (
...
 type StateAccessor struct {
-	cache    *accessorCache
-	rawCache *rawAccessorCache
+	txnCache *txnCache
+
+	blockByHash       *BlockByHashAccessor
+	blockHashByNumber *BlockHashByNumberAccessor
 }
 
 func NewStateAccessor(blockLRUSize, txnLRUSize int) *StateAccessor {
```
Gezort, line 34: btw if you wanna improve cache it may make sense to implement slightly different algo than simple LRU

https://en.wikipedia.org/wiki/Adaptive_replacement_cache - this one seems to be quite efficient
https://github.com/alexanderGugel/arc - and here's some implementation in go

PR #975, opened 2025-06-23
[hash] Fix empty code hash and don't write empty code to db

dmtrskv:
I suppose they were omitted from the original PRs by mistake.

Incorrect empty cache potentially could cause problems.
Writing empty code to db (at the same address btw) doesn't make sense.


nil/internal/types/code.go
```diff
@@ -22,7 +25,7 @@ func (c Code) Clone() Code {
 
 func (c Code) Hash() common.Hash {
 	if len(c) == 0 {
-		return common.EmptyHash
+		return EmptyCodeHash
 	}
```
olegrok, line 29: Good catch!

Maybe we can drop this condition at all because `common.KeccakHash` from empty array should be exactly EmptyCodeHash

olegrok (approved):

PR #980, opened 2025-06-25
[cache] BlockAccessor with caching separate from StateAccessor

dmtrskv:
A lot of usages of state accessor require only blocks, so we can introduce a simpler block accessor.
The state accessor is modified in another PR: https://github.com/NilFoundation/nil/pull/958


nil/services/rpc/rawapi/internal/local_call.go
```diff
@@ -228,7 +228,7 @@ func (api *localShardApiRo) Call(
 	es, err := execution.NewExecutionState(tx, shardId, execution.StateParams{
 		Block:          block,
 		ConfigAccessor: configAccessor,
-		StateAccessor:  api.accessor,
+		BlockAccessor:  execution.NewBlockAccessor(32), // todo: use shared accessor
```
dmtrskv, line 231: will be fixed in state accessor PR

nil/internal/execution/block_accessors.go
```diff
@@ -0,0 +1,304 @@
...
+
+	return &BlockHashByNumberAccessor{
+		cache: cache,
+	}
+}
+
+func (b *BlockHashByNumberAccessor) Get(tx db.RoTx, shardId types.ShardId, num types.BlockNumber) (common.Hash, error) {
+	key := shardAndBlockId{shardId, num}
+	// todo: use cache
+	// currently, the block can be somehow updated
```
dmtrskv, line 96: This problem is reproduced in https://github.com/NilFoundation/nil/pull/981
Not caching here is no problem for now

PR #985, opened 2025-06-30
drop external deployment

olegrok:
This patch removes external deployment from public interfaces (CLI, niljs) and replaces it with deployment via faucet.
We need it to support EOA because each address with positive balance and without contract code should be considered as EOA. But right now this address can be used to deploy a contract.


nil/services/relayer/internal/l2/init.go
```diff
@@ -103,13 +103,9 @@ func initDebugL2SmartAccount(
...
-	addr, err := debugDeployer.CreateSmartAccount(
-		types.BaseShardId,
-		salt,
-		amount,
-		fee,
-		&key.PublicKey,
-	)
+	pubkey := crypto.FromECDSAPub(&key.PublicKey)
+	smartAccountCode := contracts.PrepareDefaultSmartAccountForOwnerCode(pubkey)
+	addr, err := debugDeployer.Deploy(types.BaseShardId, smartAccountCode, *salt, amount)
```
dmtrskv, line 108: the original code was a single call, and it'd be nice to preserve it and extract this to a function, especially since it's also used elsewhere

niljs/src/smart-accounts/SmartAccountV1/SmartAccountV1.ts
```diff
@@ -231,45 +234,12 @@ export class SmartAccountV1 implements SmartAccountInterface {
...
-    const { hash } = await this.requestToSmartAccount({
-      data: data,
-      deploy: true,
-      seqno: 0,
-      feeCredit: refinedCredit,
-      maxPriorityFeePerGas: maxPriorityFeePerGas,
-      maxFeePerGas: maxFeePerGas,
+    return faucetClient.deploy({
+      shardId: this.shardId,
+      code: data.slice(0, data.length - 32), // remove salt from code
```
dmtrskv, line 239: is it a todo?

nil/services/faucet/jsonrpc.go
```diff
@@ -128,6 +134,89 @@ func (c *APIImpl) TopUpViaFaucet(
 	return hash, nil
 }
 
+func (c *APIImpl) Deploy(
+	ctx context.Context, shardId types.ShardId, code hexutil.Bytes, salt types.Uint256, amount types.Value,
+) (types.Address, error) {
```
dmtrskv, line 139: Let's extract the common code with the top-up.
Also, top-up can benefit from receipt waiting too. It's always used along with it.

olegrok: Done

nil/services/relayer/internal/l2/init.go
olegrok: Sounds reasonable. Fixed.

joedy27:
> ## Pull Request Overview
> This PR removes the legacy “external deployment” flow and switches all public interfaces to use a faucet-based deployment.
> 
> * Changed `selfDeploy` signature to accept a `FaucetClient` instead of a boolean flag.
> * Updated smart account and faucet contracts/tests to use the new `deploy` RPC.
> * Cleared out deprecated external‐deploy code paths from CLI, examples, docs, and services.
> 
> ### Reviewed Changes
> Copilot reviewed 33 out of 33 changed files in this pull request and generated 1 comment.
> 
> Show a summary per file
> Comments suppressed due to low confidence (2)

PR #986, opened 2025-06-30
Update badger db

olegrok:
This patch updates dependencies in general but the main change is updating badgerdb to this branch: https://github.com/NilFoundation/badger/tree/nil


olegrok:
> Why did biome files change?

New manor version of biome is delivered with new nixpkgs version

